PR #20, opened 2020-10-29
Translation : de-Central-Servers.md

SmuSmu:
Sorry ther is no good Translation for Headless. So I kept it. Because kopflos is very strange :-). Normally I would use unbeaufsichtigt (unattended) but headless is a very common it word in german.

Can you add again label hacktoberfest-accepted :-)

Or simply a hacktoberfest label to the complete repository to get even more translators

### DIFF
--- a/wiki/de/de-Central-Servers.md
+++ b/wiki/de/de-Central-Servers.md
@@ -1,0 +1,24 @@
+---
+layout: wiki
+title: "Zentrale Server"
+lang: "de"
+permalink: "/wiki/Central-Servers"
+---
+
+# Registrierung : Öffentliche Server
+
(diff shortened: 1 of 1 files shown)


SmuSmu:
Shame on me you already add hacktoberfest label. 👍 

Maybe create some issues to get attention for more translators?

ann0see:
I will be looking into it soon b

ann0see:
Same here: Sie should be changed to du. But otherwise everything seems ok. 

> Maybe create some issues to get attention for more translators?

We will be doing a freeze and probably a period of restructuring @gilgongo ? I don't think we should open new issues at the moment.

PR #36, opened 2020-11-04
Add button linking to getting started page on overview

ann0see:
Since the site https://jamulus.io/wiki/ ranks high on Google I highly suggest to add a prominent button to the Getting Started page there. This is a quite important fix which I think must be addressed before the freeze is over

### DIFF
--- a/wiki/de/index.md
+++ b/wiki/de/index.md
@@ -5,3 +5,4 @@ lang: "de"
 permalink: "/wiki/"
 ---
 # Übersicht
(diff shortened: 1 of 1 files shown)


ann0see:
@gilgongo this is also something to discuss for the next release

gilgongo:
What Google search(es) does it rank for. BTW?

ann0see:
It used to rank high for jamulus wiki but it has actually changed a bit. I must recheck.

ann0see:
Ok. At least for German now the getting started page ranks better. Nevertheless the overview page must be overthought (but not removed). We could add a search function there or something similar

PR #49, opened 2020-11-19
Changes 3.6.1

gilgongo:
Containing a couple of small changes as a test of the proposed translation process.

### DIFF
--- a/wiki/en/en-Installation-for-Linux.md
+++ b/wiki/en/en-Installation-for-Linux.md
@@ -29,16 +29,10 @@ tar -xvf latest.tar.gz
 
 First, update your package list (e.g. on Debian based distributions with `sudo apt-get update`).
 
-On **Ubuntu-based** distributions 18.04+, and on Debian 9 or 10:
+On **Ubuntu-based** distributions 18.04+, Debian 9 or later,  Raspberry Pi Raspbian Buster release or later:
 
 ```shell
 sudo apt-get install build-essential qt5-qmake qtdeclarative5-dev qt5-default qttools5-dev-tools libqt5concurrent5 libjack-jackd2-dev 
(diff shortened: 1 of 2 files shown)


wiki/en/en-Installation-for-Linux.md
```diff
@@ -29,18 +29,12 @@ tar -xvf latest.tar.gz
 
 First, update your package list (e.g. on Debian based distributions with `sudo apt-get update`).
 
-On **Ubuntu-based** distributions 18.04+, and on Debian 9 or 10:
+On **Ubuntu-based** distributions 18.04+, **Debian** 9 or later,  Raspberry Pi Raspbian Buster release or later:
 
 ```shell
 sudo apt-get install build-essential qt5-qmake qtdeclarative5-dev qt5-default qttools5-dev-tools libqt5concurrent5 libjack-jackd2-dev
```
ann0see, line 35: I think the libqt5concurrent5 is not needed since it's already in one of the packages mentioned before.

What was the difference between debian 11 and 10 (especially why is qt5-qmake now required)?

ann0see (changes requested):

wiki/en/en-Installation-for-Linux.md
gilgongo: OK yes remove libqt5concurrent5  then.  qt5-qmake is required for Ubuntu, so I assume it's required for Debian.

ann0see: Ok. I don't know about that. But I trust you.

ann0see (approved):

gilgongo:
Looks like I can't squash this to the translation branch. I'm closing and starting again.

PR #52, opened 2020-11-20
Some minor changes I propose for the next version

ann0see:
Despite that we should rethink:
* Do we need the feel free to add your own demos?
* we need to rethink the getting started page especially the audio interface section to reduce stressing the USB microphone


ann0see:
Wrong branch

PR #47, opened 2020-11-18
translation fr 1

Saimyx:
I translated :
wiki/fr :
- index
- getting started
- Demos

_data/fr :
- general.yml
- navigation.yml

- _includes/wiki/fr/footertext.md
- _config.yml
- fr-index.html

PR #41, opened 2020-11-15
Starting french translation (index and get started pages)

gegeweb:
French translation of Jamulus website.
Index and get starded page are translated.

Issue #38, opened 2020-11-13
[Translate] French


Saimyx:
Hello, I discovered Jamulus a few days ago and I started to translate website in french (I am french).
My fork is here : https://github.com/Saimyx/jamulussoftware.github.io

:)

ann0see:
That looks great. I can't merge it yet since you need to add fr to the array in config.yml and translate the footertext.md file into french (create a folder called fr in main/_includes/wiki/ and copy the translated file footertext.md there). After doing that you can open a pr

trebmuh:
Hi.
I'm the Jamulus French translator. I've had "translate Jamulus website" on my TODO list but have lately been running out of time to do so. Hence, I'm very glad someone started it.
Please, make a PR against master and I'll be happy to make a review.

ann0see:
Ah ok. So @trebmuh will do the check and review for french then? 

It's always good to have two people who can cross check.

Saimyx:
@ann0see I just started the translation, so it's not finished yet. I corrected the -blank, thank you.

@trebmuh Yes, I take care with pleasure !

Today I looked at the size of the files, and there's a lot of translation to do :P
Do I make a PR once I have everything, or do we do it bit by bit? I spoke to French people about this software, and they seemed to be disturbed not to be able to read in French.

I copied all the English files from the wiki, but I guess as long as they're not translated they shouldn't be there? When they are translated I rename them with 'fr-'.

I just made a second commit where I started to translate the the wiki

ann0see:
> Do I make a PR once I have everything, or do we do it bit by bit?

I would create two PRs the first for the important pages (Homepage, Getting Started, Installation on Windows/Mac/Linux) the second one for the less important pages.

You should only add translated files since then everything which is not translated will stay in sync with the English page.

Saimyx:
Here is the diagram of the fr-getting-starded page :
![fr-79309764-bd387280-7ef2-11ea-9d81-1e81302525e6](https://user-images.githubusercontent.com/44293594/99189852-b2351180-2763-11eb-96fc-660f5038a83c.png)

trebmuh:
Nice!
A few things:
- "audio interface" in EN would be better translated to "interface audio" since, it can be an mother-board chip too, and not necessarily a sound card
- "Publics" would be better without a capital "P"
- "membre" is missing a "s"

Overall, good job!

gegeweb:
[...]

PR #55, opened 2020-11-21
W: [fr] index, getting-started, navigation

gegeweb:
/ping @trebmuh @Saimyx for review.

This is based on the current merge request on the 'main' branch and the remarks.
I used 'Exemples d'utilisation' because of the title of the linked page.
Same for 'Manuel du logiciel'.

### DIFF
--- a/1-fr-index.html
+++ b/1-fr-index.html
@@ -1,0 +1,69 @@
+---
+title: "Jamulus – Jouez de la musique en ligne. Avec vos amis. Librement et gratuitement."
+lang: "fr"
+permalink: /
+layout: mainhomepage
+ldjAppCategory: "Musique"
+ldjSlogan: "Jouez de la musique en ligne. Avec vos amis. Librement et gratuitement."
+ldjCreator: "Volker Fischer et contributeurs"
+metadescription: "Jamulus est un logiciel libre open qui permet aux musiciens et musiciennes de jouer de la musique, répéter ou simplement « taper le bœuf », en temps réel via Internet."
+mAltProgIcon: "Icône de Jamulus"
+mTSlogan: "Jouez de la musique en ligne. Avec vos amis. Librement et gratuitement."
+mTGetStartedNow: "Premier pas…"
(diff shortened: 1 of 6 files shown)


trebmuh:
Please, don't rush! This PR includes work duplication and add mistakes to [the French initial PR](https://github.com/jamulussoftware/jamuluswebsite/pull/47) that we already have been reviewing.

Please, don't rush! Specifically you [wrote yourself](https://github.com/jamulussoftware/jamuluswebsite/issues/38#issuecomment-730958272): "_I'm waiting to the updated merge request from @Saimyx to start._" and we both agreed with the idea of working on a "one file, one PR" after the first initial French translation (which isn't the case here).

Note that it is quite possible that @Saimyx needs the weekend to merge our suggestions because that's a bunch of work to do. I suggest we can leave him the week end. Monday afternoon, if no communication from him has hit the thread, I'll act onto moving forward in order for us not to be blocked too long, then we can start working on other documentation pages.

Can you wait until then?

gegeweb:
> I'll act onto moving forward in order for us not to be blocked too long, then we can start working on other documentation pages.
> 
> Can you wait until then?

Of course, since this is the documentation branch for 3.6.1, I didn't think I would create a conflict by doing this.
I've already worked on many pages (see #41), if the first pages are not merged, it complicates a bit the work for the rest if we want 1 page = 1 merge request.
I had some time, I still have some this weekend, I'll have much less from next week (less monday, I also have a new version to push in production, last bug hunt, with the documentation that goes with it for my work... and also technical recommendations to write quickly).
And the French version is cruelly missing...

Even if appearances suggest otherwise, I had started working on it after this [discussion](https://github.com/corrados/jamulus/issues/719).
As a result, I'm a little frustrated to see all this work backwards and blocked.
Yes, I didn't look at the issues before I pushed my work. But nobody told me, "Wait! Someone has already started too" and [here](https://jamulus.io/wiki/Contribution) it's not clear either. ;):
> You can contribute to this documentation by opening a pull request with with the following title […]

Well... let's wait.

trebmuh:
This PR should be closed without merging.

PR #81, opened 2020-11-25
W: [fr] Server Win Mac

gegeweb:
/ping @Saimyx @trebmuh


gegeweb:
I'll fix #88 here.
Wait for the feedback and conclusion about this issue before merging.
Reviewing can be start for the other points (wait for the new commit before starting)

gegeweb:
@trebmuh Do I going back to keep close to the EN, or are we waiting for the discussion and the fix on the EN about the issues above?

trebmuh:
Wait for the discussion and the fix on the EN version first.

gegeweb:
> Wait for the discussion and the fix on the EN version first.

Last commit to be in sync with en actual EN version.

PR #65, opened 2020-11-24
W: [all] fix #64

gegeweb:
Fix the structure of navigation menu : don not close and reopen 'li' after the 'h3'.

PR #86, opened 2020-11-25
W: [all] same as #65, fix #64

### LINKED ISSUE #64: [Bug] Incorrect structure for navigation
As you can see in the screenshot below (using w3m) a new list item is created after the `h3`title.
The 'li' should not be closed after the 'h3' and not reopeened before the unordered list for submenu.

Capture d’écran 2020-11-24 à 13 14 34

Here is the output of the good structure (I'll open a MR) with the fix below:
Capture d’écran 2020-11-24 à 13 57 13

Capture d’écran 2020-11-24 à 14 15 43

```html
<aside id="gnav" class="fx-col-100-xs fx-col-25-m">
	{% include langselect.html %}
	<nav id="mnv">
		<ul>
			{% if site.data.navigation.nav[0] %} {% for item in site.data.navigation.nav %} {% if item.url %}
			<li>
				<h3><a href="{{ item.url | relative_url }}" class="{% if subentry.url == page.url %}selected{% endif %}">{{ item.title }}</a></h3>
				{% else %}
				<li>
					<h3>{{ item.title }}</h3>
					{% endif %} {% if item.subfolderitems[0] %}
					<ul>
						{% for entry in item.subfolderitems %}
						<li><a href="{{ entry.url }}" class="{% if entry.url == page.url %}selected{% endif %}">{{ entry.page }}</a> {% if entry.subsubfolderitems[0] %}
							<ul>
								{% for subentry in entry.subsubfolderitems %}
								<li><a href="{{ subentry.url }}" class="{% if subentry.url == page.url %}selected{% endif %}">{{ subentry.page }}</a></li>
								{% endfor %}
							</ul>
							{% endif %}
						</li>
						{% endfor %}
					</ul>
				</li>
				{% endif %} {% endfor %} {% endif %}
		</ul>
	</nav>
</aside>
```

gegeweb:
Same as #65 merged in `changes`branch.

/ping @ann0see

### DIFF
--- a/_includes/general/nav.html
+++ b/_includes/general/nav.html
@@ -7,14 +7,11 @@
       {% if item.url %}
       <li>
         <h3><a href="{{ item.url | relative_url }}" class="{% if subentry.url == page.url %}selected{% endif %}">{{ item.title }}</a></h3>
-      </li>
       {% else %}
       <li>
         <h3>{{ item.title }}</h3>
-      </li>
       {% endif %}
       {% if item.subfolderitems[0] %}
-      <li>
         <ul>
           {% for entry in item.subfolderitems %}
           <li><a href="{{ entry.url }}" class="{% if entry.url == page.url %}selected{% endif %}">{{ entry.page }}</a>


ann0see:
I think we should use a cherry pick rather than a new pr)

gegeweb:
> I think we should use a cherry pick rather than a new pr)

I don't know what is a cherry pick. If it is better, I can close this PR.

ann0see:
It's basically a "please merge only this commit into this branch" command. But I think it's not even needed since the main branch already has this change. So if we merge translation into main there won't be a conflict also since it's also in changes it will be part of the next translation.

gegeweb:
OK, I close.
Not yet used to working with multiple branches like here. ;)

PR #90, opened 2020-11-27
JACK/Jack : consistency around the English pages

trebmuh:
Hi, I've saw a bit of a mix between the way it's written in the different pages.

JACK is an acronym (a recursive one) meaning: JACK Audio Connection Kit.
Hence here is a first argument to keep it written in capital letter.
Another argument is that it helps with disambiguating "JACK" the software, and "jack" an audio cable (guitar,...).

Here is a PR for the few occurrences fixes I found around the English pages.

HTH


trebmuh:
Pinging @gilgongo since it's changes against the English pages.

gegeweb:
A good practice for acronym is ton enclose it inside a `<abbr>` tag with a title like this for JACK:

```html
<abbr title="JACK Audio Connection Kit">JACK</abbr>
```

trebmuh:
I'm not against the idea.
Can you please open another issue for that? This one is about a fix for "JACK/Jack : consistency around the English pages" because there are a few inconsistency so far, while you're proposing something new.
Thanks.

gilgongo:
Good idea - I'll merge this and then we can put the tags in.

trebmuh:
Thanks for the quick merge. Very appreciated.

PR #170, opened 2020-12-10
Add Software Synth by @niebert

ann0see:
In this PR we should discuss how to restructure the Software Synth page since it’s now at a completely different spot.

@niebert should comment on this since he's the author.


ignotus666 (commented):
There are a number of things I'd change/add to this guide:

- Screenshots. Describing the connections in QjackCtl without screenshots is painfully confusing.

- I really can't see the need to launch QjackCtl and a2jmidid from the command line. It's much less confusing for a new user (that's who I gather this guide is intended for) to launch it from the menu or a toolbar shortcut and have QjackCtl automatically launch a2jmidid by adding a line in "Execute script after Startup" in the 'Options' tab: `a2jmidid -e &`. That way you never have to worry about launching it again.

ann0see:
> Screenshots

Agree

> I really can't see the need to launch QjackCtl and a2jmidid from the command line

In fact, the official documentation sometimes also does this. Not sure how many linux users don't use the command line. But of course we should add both. @ignotus666, you can add a few of these changes. I haven't tested the guide myself, but I will try it (should work since it seems to be ok).

ann0see:
Verified it. It does work (but you need to start jack before doing anything else). But why can't we use ALSA MIDI?

ignotus666:
> But why can't we use ALSA MIDI?

Because Jamulus uses JACK MIDI

ignotus666:
> Not sure how many linux users don't use the command line

I'm all for using the command line when it makes sense or an application doesn't provide a launcher, but cluttering up your desktop screen with terminal windows when there's no need isn't practical IMHO.

I can add the info and the screenshots; I don't have a MIDI keyboard but I can change the aliases in QjackCtl to rename the MIDI gear I do have so it matches the gear described in the guide. Not sure when I'll have time though.

ann0see:
> Because Jamulus uses JACK MIDI

Ok. Not sure how all this works (I don't usually connect my digital piano to my PC), but I thought that we only need to connect the Audio? Which kind of MIDI signal does Jamulus support (except for the mixing desk commands). 

>  cluttering up your desktop screen with terminal windows when there's no need isn't practical IMHO

Agree. We could say something like "you can also open it in your terminal with [command]".

> I don't have a MIDI keyboard

But I do. Let me take a few screenshots.

ignotus666:
> but I thought that we only need to connect the Audio?

Duh, you're right... I haven't tried it but I assume it's the synth that uses JACK MIDI then. 
I also don't understand the reason to use 'MIDI through'. If you're launching a2jmidid you can just connect your MIDI keyboard straight to the synth. But again, I haven't set it up so maybe I'm missing something.

> But I do. Let me take a few screenshots.

Ok cool.

ann0see:
> If you're launching a2jmidid you can just connect your MIDI keyboard straight to the synth.

Yes. I think so and it worked for me.

These are my screenshots: 

@niebert could you please explain why we need the midi through etc?
We might be able to simplify the workflow

Edit @ignotus666 they say that although ALSA midi is possible, they recommend JACK midi: https://github.com/pantherb/setBfree/issues/28

Jamulus Synth connect qjackctl
QJackctl ALSA-MIDI MIDI through
QJackctl JACK-MIDI a2j
QJackctl synth connect
setBfree synth

JACK-MIDI connect synth MIDI
QJackctl connect synth to Jamulus

ignotus666:
What MIDI tab does the synth appear in? I'm obviously missing something so maybe the MIDI through bit does have an explantation.

ann0see:
Not sure, but I assume JACK midi. Can test tomorrow evening

ignotus666:
I just installed setBfree and I see what happens: setBfree auto-connects to 'Midi Through' in the JACK MIDI tab, so @niebert's instructions do make sense. You can connect your MIDI keyboard directly to setBfree in the JACK MIDI tab though, without going through 'Midi Through'.
setBfree_midi

ann0see:
Ok. So should we merge it then? Any thoughts/changes needed?

Especially from @niebert

ignotus666:
I think I agree with what @gilgongo suggested elsewhere. I'm reluctant to spend any time on this since it's not going to be part of the official documentation. If @niebert wants to follow any of the suggestions, fine - and if not, that's fine too.

ann0see:
Added pictures now and I think we can merge it after the translation progress if finished

ignotus666:
Sorry to be a pest, but - in the last step, when connecting setBfree to Jamulus, it would be advisable to disconnect setBfree's outputs from the 'system' inputs in order to comply with Rule Number 1: do not listen to your local audio. Otherwise it would also lead to a flanger/echo effect. Other than that small detail, it's much better with the screenshots.

ann0see:
Can you change it please? I have to work on the german translation and fix a few things.

ignotus666:
Ok - I'll add that advice.

ignotus666:
Done. I also corrected a couple of things.

ann0see:
synth jack connect

Fixed the screenshot and edited your addition.

ann0see:
I think we can merge it, after @ignotus666 you've rechecked these additions?

ignotus666:
Looks good.

ann0see:
Merged

PR #204, opened 2020-12-18
[en] consistency: Internet speed not clear: MBit/s, Mbps,...

ann0see:
Currently in the wiki there are multiple ways expressing internet bandwidth. This PR (and the mbit branch) should try to discuss and fix a few of these inconsistencies. 

In German we usually use MBit/s but I'm not sure if this is also the case in English. @gilgongo should comment on that.

### DIFF
--- a/wiki/en/en-Getting-Started.md
+++ b/wiki/en/en-Getting-Started.md
@@ -62,5 +62,5 @@ Further information about how to avoid listening to yourself (and help with othe
 _Jamulus works on the client server principle. Everybody’s audio is sent to a server, mixed and processed there. Afterwards the audio is sent to every client. If a server is made public and registered on a central server, its information will be broadcasted to all clients._
 
 ## Footnotes
-[^1]: [Full details here](Network-Requirements){: target="_blank" rel="noopener noreferrer"}. If you have 10Mbits down and 1Mbps up, you're unlikely to run into bandwidth-related issues.
+[^1]: [Full details here](Network-Requirements){: target="_blank" rel="noopener noreferrer"}. If you have 10MBit/s down and 1MBit/s up, you're unlikely to run into bandwidth-related issues.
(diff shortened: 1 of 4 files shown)


gilgongo:
Oh I always get confused with this. 

"megabit per second (symbol Mbit/s or Mb/s, often abbreviated "Mbps")" as per [Wikipedia](https://en.wikipedia.org/wiki/Data-rate_units#Megabit_per_second)

So how about we standardise on Mbit/s? I think MBit/s is wrong as the capital "B" denotes bytes not bits?

ann0see:
Mb/s at least for me means megabyte per second. 
I think Mbit/s or MBit/s would be good. MacOS corrected MBit/s to Mbit/s as I typed this. So I think the capital b might be wrong (although in German I usually read it with a capital B).

gene96817:
I understand the confusion between bits and bytes. In the science and engineering world "b" is bits and "B" is bytes. Often, even junior engineers get confused. "Mbps" means megabits per second. When we talk about how much data is moved, it could be Mbps when thinking in terms of bits or MBps when thinking about bytes. This is unusual. But when we talk about packet size, this distinction is very important.

gilgongo:
As long as we don't look like junior engineers getting b missed up with B, I'm ok. So, Mbit/s?

ann0see:
Agree. Just found the German Wikipedia article and Mbit/s should preferred in German too (although we have "Bit", which refers to the unit of measurement).

ann0see:
@gilgongo ready for review

gilgongo:
Looks fine - is there a reason why sometimes we have a space between the number and the unit, and other times we don't? So "10Mbit/s" and "5 Mbit/s"

ann0see:
No. There's no reason on that. I also thought about removing/adding it. We should agree on what to use.

ignotus666:
There should always be a space between numbers and units of measurement (there are some exceptions but I don't think any are applicable here). That also goes for ms or any other units - I've been sloppy about that myself and am pretty sure I've written e.g. both 10ms and 10 ms.

ann0see:
Ok. I can't do it a the moment, feel free to fix these issues on the mbit branch and then merge this pr

ann0see:
Done. Should be fixed now. Will merge it in roughly 1 hour probably.

PR #210, opened 2020-12-22
A few changes 

ann0see:


wiki/en/en-Installation-for-Macintosh.md
```diff
@@ -10,14 +10,15 @@ permalink: "/wiki/Installation-for-Macintosh"
 Make sure you've already read the [Getting Started](Getting-Started) page.
 
 1. [Download Jamulus](https://sourceforge.net/projects/llcon/files/latest/download){: target="_blank" rel="noopener noreferrer" .button}
-1. **Open the downloaded `.dmg` file**. Agree to the license. You will now see two icons (Jamulus client and server).
-1. **Drag and drop both icons into your Applications folder** to install Jamulus.
+1. **Install Jamulus**. Open the downloaded `.dmg` file, agree to the license, *drag and drop* the two icons you see in the window (Jamulus client and server) into your *Applications folder*. After that, you can close this window.
```
ann0see: Do we really need so many steps?

wiki/en/en-Installation-for-Macintosh.md
```diff
@@ -10,14 +10,15 @@ permalink: "/wiki/Installation-for-Macintosh"
...
 1. **Run Jamulus**. Now you should be able to use Jamulus just like any other application. If you see a message which tells you that Jamulus can't be opened because the developer can't be verified, read on and have a look at the following section.
 
 _You can remove the folder in the Downloads directory containing the `.dmg` and eject the "Jamulus" drive on your desktop. They are no longer needed._
 
 You can also [compile your own binary](Compiling#macos) if you feel so inclined.
 
+***
```
ann0see, line 18: Add hr like on windows

wiki/en/en-Installation-for-Windows.md
```diff
@@ -8,10 +8,10 @@ permalink: "/wiki/Installation-for-Windows"
 # Installation for Windows
 Make sure you read the [Getting Started](Getting-Started) page.
 1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have one (especially internal sound cards), install [this free ASIO driver (ASIO4ALL)](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"} before installing Jamulus.
-1. [Download Jamulus](https://sourceforge.net/projects/llcon/files/latest/download){: .button target="_blank" rel="noopener noreferrer"}.
-1. **Install Jamulus** If you get a warning notice from SmartScreen, click on "More info" and "Run anyway" to install Jamulus. This is needed since we don't pay for code signing yet.
+1. [Download Jamulus](https://sourceforge.net/projects/llcon/files/latest/download){: .button target="_blank" rel="noopener noreferrer"}
```
ann0see, line 11: Add button like on macOS.

Discussion:

Is this too prominent? Do we need to focus the ASIO driver too (but not everyone needs ASIO4ALL)

wiki/en/en-Installation-for-Macintosh.md
gilgongo: Perhaps not. I don't know whether having an "install" step inside a page called "install" is a bit unclear, since it implies the other steps aren't about installing, but I doubt most people will think about that.

wiki/en/en-Installation-for-Macintosh.md
```diff
@@ -10,14 +10,15 @@ permalink: "/wiki/Installation-for-Macintosh"
...
 
 1. [Download Jamulus](https://sourceforge.net/projects/llcon/files/latest/download){: target="_blank" rel="noopener noreferrer" .button}
-1. **Open the downloaded `.dmg` file**. Agree to the license. You will now see two icons (Jamulus client and server).
-1. **Drag and drop both icons into your Applications folder** to install Jamulus.
+1. **Install Jamulus**. Open the downloaded `.dmg` file, agree to the license, *drag and drop* the two icons you see in the window (Jamulus client and server) into your *Applications folder*. After that, you can close this window.
 1. **Run Jamulus**. Now you should be able to use Jamulus just like any other application. If you see a message which tells you that Jamulus can't be opened because the developer can't be verified, read on and have a look at the following section.
 
 _You can remove the folder in the Downloads directory containing the `.dmg` and eject the "Jamulus" drive on your desktop. They are no longer needed._
 
 You can also [compile your own binary](Compiling#macos) if you feel so inclined.
```
gilgongo: Since less than about 1% of anyone reading this page will know how to do this, and those that do will know that Jamulus is FOSS, I'm inclined to remove this line. It might even confuse some people.

wiki/en/en-Installation-for-Windows.md
gilgongo: Agree on the button. I don't think it's too prominent really, but the ASIO issue is tricky. I think we have it about right (although see my comment on possibly clarifying what "internal sound card" means).

wiki/en/en-Installation-for-Windows.md
```diff
@@ -8,10 +8,10 @@ permalink: "/wiki/Installation-for-Windows"
 # Installation for Windows
 Make sure you read the [Getting Started](Getting-Started) page.
 1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have one (especially internal sound cards), install [this free ASIO driver (ASIO4ALL)](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"} before installing Jamulus.
```
gilgongo: I wonder if we should say "If you don't have an external sound card", since it's more likely that those new to this idea will understand that if they don't have another box then it must be internal?

ann0see: Agree.

ann0see: There are some internal sound cards with ASIO drivers, I think (Realtek ASIO), but they are rare.

wiki/en/en-Installation-for-Macintosh.md
ann0see: Ok. So we will also have to remove it on windows.

ann0see: @gilgongo I only see the problem for more experienced users: they would have to search for it on https://jamulus.io/wiki/ we might have to add it to the navigation

wiki/en/en-Installation-for-Windows.md
```diff
@@ -9,30 +9,28 @@ permalink: "/wiki/Installation-for-Windows"
 Make sure you read the [Getting Started](Getting-Started) page.
 1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have an external sound card, you probably need to install [this free ASIO driver (ASIO4ALL)](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"} before installing Jamulus.
 1. [Download Jamulus](https://sourceforge.net/projects/llcon/files/latest/download){: .button target="_blank" rel="noopener noreferrer"}
-1. **Install Jamulus**. Execute the Jamulus installer. If you get a warning notice from SmartScreen, click on "More info" and "Run anyway" to install Jamulus. This is needed since we don't pay for code signing yet.
+1. **Install Jamulus**: Execute the Jamulus installer. If you get a warning notice from SmartScreen, click on "More info" and "Run anyway" to install Jamulus. This is needed since we don't pay for code signing yet.
```
gilgongo: "Execute" is a bit of a technical word :-) Stylistically, perhaps "run" or even "double click" would be better here. Also, what's SmartScreen? Isn't it just a warning dilogue?

ann0see: I think double click would be better. And yes, the dialog is from SmartScreen

ann0see: Now the dialog from SmartScreen is gone (probably since Jamulus was downloaded by a specific number of Windows users which now adds this version to a whitelist?)

wiki/en/en-Installation-for-Macintosh.md
ann0see: Let's use my proposal and since it might add benefit for inexperienced users.

wiki/en/en-Installation-for-Windows.md
ann0see: https://github.com/corrados/jamulus/issues/792#issuecomment-752946745

Maybe we could add a statement to the installer.

wiki/en/en-Installation-for-Macintosh.md
gilgongo: We can put it in a footnote if you really want, but those wanting to compile Jamulus on Windows/Mac will just go to Github I think as they are likely to be contributors to the project anyway. I highly doubt anyone would go to the trouble of compiling Jamulus on those platforms for any other reason.

gene96817: Do we want all users to install both the client and server software (from line 14: "* Drag and drop both icons"). It seems to me most users only need to install the client. I don't want to be helping naive users with server questions. The server issues are hard enough for me.

gilgongo: Heh - well, on Windows they get it whether they want it or not. You raise a good point, but it's hard to know what to say here without diverging into some kind of discussion about which to install and why (and I think we don't label the icon "Jamulus client" which makes referring to it harder).

ann0see: > I highly doubt anyone would go to the trouble of compiling Jamulus on those platforms for any other reason.

Ok. Valid point. We might need to add the overview page link somewhere more prominent. 

@gene96817 

> It seems to me most users only need to install the client.

Probably true. Also the server on macOS seems to be a own binary? I think windows and Linux only have one binary? So the user would even benefit from less space being used on his hard drive.

ann0see:
@gilgongo If you're ok with this, please merge it.

PR #217, opened 2020-12-27
Change some texts on the homepage

ann0see:
Jamulus is currently bad at SEO. I think we could improve the wording on the homepage

### DIFF
--- a/1-index.html
+++ b/1-index.html
@@ -1,5 +1,5 @@
 ---
-title: "Jamulus ‒ Play music online. With friends. For free."
+title: "Jamulus ‒ Jam online. With friends. For free."
 lang: "en"
 permalink: /
 layout: mainhomepage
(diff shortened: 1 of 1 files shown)


1-index.html
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Jamulus ‒ Play music online. With friends. For free."
+title: "Jamulus ‒ Jam online. With friends. For free."
```
gilgongo, line 2: From an SEO point of view, I agree that "jam" is probably going to be a more realistic term for us to target than "play". It's just a pity that perhaps only a small subset of people actually use Jamulus to jam with.

gilgongo (commented):
Looks good. As I've pointed out before, tinkering with the home page content won't make any difference to our ranking at this point though, but there's no harm in it :-)

ann0see:
> As I've pointed out before, tinkering with the home page content won't make any difference to our ranking at this point though

Not 100% sure about that, but we can try.

gilgongo:
> Not 100% sure about that, but we can try.

Hehe. Well, you can't be 100% sure about anything in SEO :-)

PR #232, opened 2021-01-04
Add Debian packages

ann0see:
They are not yet completely tested, but once they work correctly, this can be added to the wiki. Please review.

### DIFF
--- a/wiki/en/en-Installation-for-Linux.md
+++ b/wiki/en/en-Installation-for-Linux.md
@@ -23,6 +23,7 @@ If you're on **Debian** or **Ubuntu** (only amd64), you may try the compiled deb
 1. Download the deb file from the [Jamulus GitHub release](https://github.com/corrados/jamulus/releases/tag/latest) page
 1. Update apt: `sudo apt-get update`
 1. Install the package: `sudo dpkg -i <Jamulus-deb-file-name>.deb`
+1. Since Jamulus needs JACK, you have to install it too. We recommend to use `QjackCtl` to configure JACK. You can install it via `sudo apt-get install qjackctl`
 
 **Note:** The generated deb files will install Jamulus with a lower case "j". You can therefore start it in terminal by typing `jamulus`.
 
--- a/wiki/en/en-Installation-for-Linux.md
+++ b/wiki/en/en-Installation-for-Linux.md
@@ -10,11 +10,23 @@ Make sure you read the [Getting Started](Getting-Started) page.
 
 ## Jamulus packages
 
-Although we don't provide official packages, you might find Jamulus in the package manager of your distro. Have a look at this table:
+Although we don't provide official packages for every Linux distribution, you might find Jamulus in the package manager of your distro. Have a look at this table:
 
 [![Packaging status](https://repology.org/badge/vertical-allrepos/jamulus.svg)](https://repology.org/project/jamulus/versions)
 
-There is also list of unofficial [packages for different distributions here](https://github.com/corrados/jamulus/issues/223). If you don't find it in your package manager, you need to compile Jamulus from source. That's quite easy:
+There is also list of unofficial [packages for different distributions here](https://github.com/corrados/jamulus/issues/223).
+
+### Debian and Ubuntu amd64
+
+If you're on **Debian** or **Ubuntu** (only amd64), you may try the compiled deb packages from GitHub Actions.
+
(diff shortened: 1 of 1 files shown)


trebmuh:
There is no .deb package on the link pointed at.
Is that expected to be there for the next Jamulus version?

ann0see:
Yes. They should be there for the next release

trebmuh:
OK, seems all good for me then.

wiki/en/en-Installation-for-Linux.md
```diff
@@ -10,11 +10,24 @@ Make sure you read the [Getting Started](Getting-Started) page.
...
+Although we don't provide official packages for every Linux distribution, you might find Jamulus in the package manager of your distro. Have a look at this table:
 
 [![Packaging status](https://repology.org/badge/vertical-allrepos/jamulus.svg)](https://repology.org/project/jamulus/versions)
 
-There is also list of unofficial [packages for different distributions here](https://github.com/corrados/jamulus/issues/223). If you don't find it in your package manager, you need to compile Jamulus from source. That's quite easy:
+There is also list of unofficial [packages for different distributions here](https://github.com/corrados/jamulus/issues/223).
+
+### Debian and Ubuntu amd64
+
+If you're on **Debian** or **Ubuntu** (only amd64), you may try the compiled deb packages from GitHub Actions.
```
trebmuh: The (only amd64) is a bit confusing here. One can't know for sure if this apply for Ubuntu or Ubuntu&Debian. My guestimate (don't have time to research for it right now) is that it's only for Ubuntu. Hence I would suggest the following wording:
> If you're on **Debian** or **Ubuntu** (only amd64 works for Ubuntu)

ann0see: Currently the produced deb file is compiled for 64 Bit on Ubuntu. I'm not 100% sure if it doesn't work on 32 Bit Debian, but this is to be tested.

ann0see: I now made a few changes to this file. I'm not sure if we really need to support 32 Bit .deb builds. As far as I read, Ubuntu doesn't support 32 Bit anymore. The only supported architecture which I think might be missing is armhf or arm64 for Raspberry Pi. But that's another issue.

trebmuh: Ubuntu doesn't, but a lot of other distributions will keep 32 bits for a while (Debian and a bunch of its derivatives comes to mind).
As a figure, 32 bits intel users are still ~17% for LibraZiK from the numbers I have.

@corrados : can you tell us if you plan to drop or keep 32 bits support?

corrados: On Windows I want to keep compatibility to 32 bit. For Linux, @ann0see can decide this if he like.

trebmuh: @corrados thanks. Curiosity: why keeping 32 bits compatibility on windows and not on linux?

ann0see: I'm not even sure if it's possible to provide 32Bit support for Linux since we need to build it on Ubuntu or use a Docker image. Should we open an issue on the main repo on this topic? I don't really see much benefit of adding 32Bit support for Intel since I think it's outdated. If somebody wants to look into it and is interested in supporting it, feel free to do so.

corrados: @trebmuh since almost nobody is using Jamulus on Linux, unfortunately, see:
grafik

ann0see: That's just the sourceforge download. I assume most Linux users just compile Jamulus itself.

trebmuh: ...and/or get it from their distributions.

> I'm not even sure if it's possible to provide 32Bit support for Linux since we need to build it on Ubuntu or use a Docker image. Should we open an issue on the main repo on this topic? I don't really see much benefit of adding 32Bit support for Intel since I think it's outdated. If somebody wants to look into it and is interested in supporting it, feel free to do so.

For what it's worth, I dislike the "linux = ubuntu" approach. Last time I've had a discussion about this, we've looked and found some figures showing that ubuntu was roughly 35% of the linux users choice. It would be a wrong move IMHO to stop 32 bit Intel support just because Canonical did this choice with regards to their internal policy.

ann0see: Ok. Let me try if the resulting .deb works on a 32bit Debian 10 VM.

PR #234, opened 2021-01-06
Add more detail for the Windows build process

jdrage:
While setting this up for the first time I ran into a couple of snags, so I thought I would update the documentation to help others who tread this path.

Qt setup requires selection of components under a particular version, so I added details about how to find those.

The destination path for the ASIO SDK files was incorrect, I fixed that.

Added license info for Qt, since there is also a commercial version available and that was a bit confusing.

The first time you load the project in Qt you need to configure the kit for the project, I added that.

Also added links to a few of the relevant sites, including Visual Studio and a direct link for the ASIO SDK download.


wiki/en/en-Compiling.md
```diff
@@ -62,12 +62,22 @@ Will build the file and make it available in `./Release/Jamulus.app`
 ## Windows
 You don't need to compile Jamulus for Windows since we provide official binaries, but if you really want to, follow the instructions below.
 
-**Required software**: [Qt](https://www.qt.io/download), a compiler like Visual Studio, ASIO development files
+### Required Software
+- A compiler like [Visual Studio](https://visualstudio.microsoft.com)
+- [Qt](https://www.qt.io/download)
+    - Use the free GPLv2 license for Open Source development
+    - To determine the Qt version you need, check [qt-installer-windows.qs](https://github.com/corrados/jamulus/blob/master/windows/qt-installer-windows.qs): under INSTALL_COMPONENTS you will see `qt.qt5.[version]`, e.g., 5123 means version 5.12.3.
+    - Select Components during installation: Expand the Qt section, find the matching version, e.g., **Qt 5.12.3**, and add the compiler components for your compiler, e.g., `MSVC 2017 32-bit/64-bit` for Visual Studio 2019
```
ann0see: Is it really MSVC 2017 for Visual Studio 2019?

wiki/en/en-Compiling.md
```diff
@@ -62,12 +62,22 @@ Will build the file and make it available in `./Release/Jamulus.app`
 ## Windows
 You don't need to compile Jamulus for Windows since we provide official binaries, but if you really want to, follow the instructions below.
 
-**Required software**: [Qt](https://www.qt.io/download), a compiler like Visual Studio, ASIO development files
+### Required Software
+- A compiler like [Visual Studio](https://visualstudio.microsoft.com)
```
ann0see: Not 100% sure about the MD syntax here. I usually use `*` instead of `-`. But it seems to work this way too. @ignotus666 do you know what the correct way of generating a list is?

ann0see (commented):
I think we should merge my proposed changes into this file. Can you copy them to your repo?

wiki/en/en-Compiling.md
jdrage: Yes, it is working for me.

wiki/en/en-Compiling.md
jdrage: Either one seems to work but I will change to `*` if that's the convention here.

wiki/en/en-Compiling.md
ann0see: Ok. Not exactly sure what MSVC 2017 is then. I thought it was the compiler and there might be compatibility issues if you choose MSVC 2017 but run Visual Studio 2019.

wiki/en/en-Compiling.md
ignotus666: > @ignotus666 do you know what the correct way of generating a list is?

Sorry, I don't really know much about this, but AFAIK in other documents (e.g. software manual) hyphens are used for bullet points, ** for bold and ### for larger font bold.

ann0see: Ok, so probably both are correct.

PR #257, opened 2021-01-20
French l10n improvements

trebmuh:
- informs about external links in English when relevant
- fixes a non-functional link

@ann0see : please merge


ann0see (approved):
If you find a french equivalent to the links, you can use them of course.

PR #260, opened 2021-01-22
[en] update footertext.md

### LINKED ISSUE #255: [Bug] footer outdated
In the footer, we have "`Summer 2020! Tell us about Jamulus - take our anonymous survey!`".
It starts to be outdated (we're not in summer 2020 any more)  and we might want to change that to something more up-to-date.

trebmuh:
closes #255

### DIFF
--- a/_includes/wiki/en/footertext.md
+++ b/_includes/wiki/en/footertext.md
@@ -1,6 +1,6 @@
 
 
-**Summer 2020! Tell us about Jamulus - [take our anonymous survey!](https://forms.gle/hSSjsxjWj2Pnp5kr7)**
+**Tell us about Jamulus - [take our anonymous survey!](https://forms.gle/hSSjsxjWj2Pnp5kr7)**
 
 
 _This documentation is licensed under a [Creative Commons License](https://creativecommons.org/licenses/by-sa/4.0/deed.en){: target="_blank" rel="noreferrer" }. Want to [contribute](Contribution)?_


_includes/wiki/en/footertext.md
```diff
@@ -1,6 +1,6 @@
 
 
-**Summer 2020! Tell us about Jamulus - [take our anonymous survey!](https://forms.gle/hSSjsxjWj2Pnp5kr7)**
+**Tell us about Jamulus - [take our anonymous survey!](https://forms.gle/hSSjsxjWj2Pnp5kr7)**
```
ann0see, line 3: Thanks. Nevertheless, merging it to release instead of changes would violate the "typo fix only" policy. Translators would not be aware of this change. But I must admit, I've partly violated it with the new installer.

ann0see (changes requested):

_includes/wiki/en/footertext.md
trebmuh: I thought you were doing something like a hook which automatically updated changes with modifications made to release?

trebmuh: Any way, done: https://github.com/jamulussoftware/jamuluswebsite/commit/38511a38b4372e3843a2883dc2a0eb3f4e5e0cde

ann0see: Yes. But this doesn't include content which needs to be translated. So if you change something in french and merge it to release, it will also add the commit to changes. But obviously this isn't possible for multiple languages.

trebmuh: You're right. Thanks for the explanation and sorry for the mistake here.

PR #288, opened 2021-02-12
Clarify translation

ann0see:
Probably the contributing file needs to be shortened/clarified. Just a small addition

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,7 +26,7 @@ If not: This will be discussed this later (since you introduce a new language).
 If so look here:
 ### 1. Be clear if your change needs translation
 **No:** A fix which doesn't need to be translated (e.g. typo) --> Open a Pull Request to `release`. It will go live a few minutes after the PR is merged
-**Yes:** A fix which needs to be translated (e.g. if you want to make a sentence or paragraph clearer) --> Commit to `changes` since others will need to translate it. It will be part of the next release.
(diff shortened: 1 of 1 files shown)


gilgongo:
For reasons I can't remember, we now have two overlapping "contribution" pages (was that on purpose?). I'm going to have a go at consolidating them down to one, shrinking [CONTRIBUTING.md](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md) to basically just a link to one on [the wiki](https://jamulus.io/wiki/Contribution) to contain content from both (and try to put it into some kind of roughly similar style while I'm at it).

ann0see:
As far as I know, the file on the website was for a new contributor who probably doesn't know much about FLOSS while the CONTRIBUTING.md file was related to somebody who's more familiar with git etc.

The flow would be: 1. Contributing page on the website to get general information --> CONTRIBUTING.md file on the repo to get more detailed information.

pljones (approved):

gilgongo:
I've also just noticed that README.md has even **more** stuff about contributing! (and [other one here](https://github.com/jamulussoftware/jamulus/blob/master/CONTRIBUTING.md)) This is getting out of hand... :-) Having semi-duplicated information maintained is four separate places is not a good idea for the docs. 

> while the CONTRIBUTING.md file was related to somebody who's more familiar with git etc.

OK so we can make a distinction between a contributor of **code** rather than anything else (so engineers will be looking at CONTRIBUTING.md). If so, I'll put the code contribution details in there, and the non-code stuff on the wiki, and link between the two. README.md can just link to CONTRIBUTING.md. Also BTW I'll treat the general process of translation of the application as the same as that of the website for now. 

Anyway, I'll make a pull request for all this and we can discuss that.

PR #290, opened 2021-02-13
Link --serverpublicip desc. to docs

gilgongo:
As per https://github.com/jamulussoftware/jamulus/pull/954#issuecomment-775757311

### DIFF
--- a/wiki/en/en-Command-Line-Options.md
+++ b/wiki/en/en-Command-Line-Options.md
@@ -52,4 +52,4 @@ You can see all possible options your version supports by starting Jamulus with 
 |    `-z` |`--startminimized` | Start minimized | (server only) |
 |       |`--ctrlmidich`     | MIDI controller channel to listen on and control number offset, format: `channel[;offset]` | (client only) see [Tips & Tricks](Tips-Tricks-More#Using-ctrlmidich-for-MIDI-controllers) |
 |       |`--clientname`     | Window title and JACK client name | (client only) |
-|       |`--serverpublicip` | When running your own central server with own slave servers in a private network, specify your public IP with this option to enable clients from the Internet to connect. You will still need proper port forwarding in your router/firewall.| (server only) |
+|       |`--serverpublicip` | Allows servers to be seen on the Internet when listed by a Central Server on a LAN | (server only) [See note](Choosing-a-Server-Type) |
--- a/wiki/en/en-Command-Line-Options.md
+++ b/wiki/en/en-Command-Line-Options.md
@@ -52,4 +52,4 @@ You can see all possible options your version supports by starting Jamulus with 
 |    `-z` |`--startminimized` | Start minimized | (server only) |
 |       |`--ctrlmidich`     | MIDI controller channel to listen on and control number offset, format: `channel[;offset]` | (client only) see [Tips & Tricks](Tips-Tricks-More#Using-ctrlmidich-for-MIDI-controllers) |
 |       |`--clientname`     | Window title and JACK client name | (client only) |
-|       |`--serverpublicip` | Allows servers to be seen on the Internet when listed by a Central Server on a LAN | (server only) [See note](Choosing-a-Server-Type) |
+|       |`--serverpublicip` | Supply the server's public IP address  | (server only) [See note](Choosing-a-Server-Type) |


wiki/en/en-Command-Line-Options.md
```diff
@@ -52,4 +52,4 @@ You can see all possible options your version supports by starting Jamulus with
 |    `-z` |`--startminimized` | Start minimized | (server only) |
 |       |`--ctrlmidich`     | MIDI controller channel to listen on and control number offset, format: `channel[;offset]` | (client only) see [Tips & Tricks](Tips-Tricks-More#Using-ctrlmidich-for-MIDI-controllers) |
 |       |`--clientname`     | Window title and JACK client name | (client only) |
-|       |`--serverpublicip` | When running your own central server with own slave servers in a private network, specify your public IP with this option to enable clients from the Internet to connect. You will still need proper port forwarding in your router/firewall.| (server only) |
+|       |`--serverpublicip` | Allows servers to be seen on the Internet when listed by a Central Server on a LAN | (server only) [See note](Choosing-a-Server-Type) |
```
pljones: This is not a requirement when running a Central Server on a LAN.  Rock, Jazz, Classical/Folk, Choir/Barbershop all run on my LAN here (192.168.1.x internal address).  Nothing needed other than a public static IP and port forwarding on the router.  So I think the new text is somewhat misleading.

gilgongo: OK do we mean, "Allows servers to be seen on the Internet when listed by a Central Server behind a NAT firewall"? Or maybe, "Allows servers running on a LAN to be seen on the Internet when listed by a Central Server on their network"?

pljones: Ah, sorry, I've just tested the _explicit_ scenario.  This is when _both_ servers are inside the LAN, behind the NAT wall.

So I've just started up jamulus.drealm.info:21224, port mapped by the router to internal port 55850.  I registered that on jamuluscentral.drealm.info:22824.

This confuses Jamulus Central and completely stops my connecting internally....
image

image

I'll try again with the --serverpublicip -- on the registering server.

gilgongo: Yes, sorry, this is the whole point of the `--serverpublicip` option as I understand it. Of course, if you are running the directory on your LAN with servers  on that LAN and don't want clients on the WAN in those servers, then it's not needed. 

(Note how the use of the word "Directory" nicely disambiguates "Server"  in this convo - possibly a bit of an edge case example, but still)

pljones:
Ignore what I was saying, I was testing wrong...

wiki/en/en-Command-Line-Options.md
pljones: I'm trying to get a server on my LAN registered with a directory on the LAN _and_ exposed to Jamulus Central (which is a Jamulus client outside my LAN for purposes of testing).

pljones: Test scenarios:

* Directory, no "serverpublicip".  Unregistered server, no "serverpublicip", no "centralserver", port not remapped by the router
image
image
Above is as expected
* Directory, no "serverpublicip".  Registered server, no "serverpublicip", "centralserver" as 127.0.0.1:22824, port not remapped by the router
image
image
It looks like this has a problem - it's showing the address of the registered server as 127.0.0.1 in the directory, which isn't much use...  Worst, the registered server port has been confused with the directory server.  I've no idea how that happens - possibly because the registered server pings Jamulus Explorer, exposing its real public IP address, which is the same as the directory server -- and Jamulus Explorer has nothing else it can do but link by that IP address, despite the port not matching.
```
Feb 13 14:33:27 fs-peter Jamulus-Central8[1079682]: Requested to register entry for 127.0.0.1:56850 (192.168.1.19:56850): Test Server (21124)
```
* Directory, no "serverpublicip".  Registered server, "serverpublicip" set to the 81.174.155.168, "centralserver" as 127.0.0.1:22824, port not remapped by the router
image
image
So as ever, port not remapped by the router _appears_ to work here.
```
Feb 13 14:47:27 fs-peter Jamulus-Central8[1082559]: Requested to register entry for 127.0.0.1:56850 (81.174.155.168:56850): Test Server (21124)
```

What can't happen successfully is to have a second registered server inside my LAN running on port 56850.  Well, internally, it would be okay: the IP:port pair would be unique.  But externally, the serverpublicip IP and port wouldn't, unless the port got mapped.  But, unfortunately, port mapping and directory server ping back don't mix.  So I could turn off the ping back - losing some support for registering external servers - or ensure all registered servers on the LAN, regardless of IP, have a unique port.

pljones: OK, so the next tests are the same as above but here we'll use port forwarding on the router to see how that affects life.  Not directly related to this but it's still something I'm hoping we'll one day get working with registered servers.

* Directory, no "serverpublicip". Unregistered server, no "serverpublicip", no "centralserver", port remapped by the router
(without forwarding)
image
(with forwarding)
image
And, of course, nothing registered.  Unregistered server log:
```
Feb 13 15:13:47 fs-peter Jamulus-wip[1086790]: - selected port number: 56850
Feb 13 15:13:47 fs-peter Jamulus-wip[1086790]: - maximum number of channels: 4
Feb 13 15:13:47 fs-peter Jamulus-wip[1086790]: - server info: Test Server (21124);West Ealing;224
```
* Directory, no "serverpublicip". Registered server, "serverpublicip" set to the 81.174.155.168, "centralserver" as 127.0.0.1:22824, port remapped by the router
image
image
The issue we're having is clear from the directory server log
```
Feb 13 15:20:07 fs-peter Jamulus-Central8[1086736]: Requested to register entry for 127.0.0.1:56850 (81.174.155.168:56850): Test Server (21124)
```
* Directory, no "serverpublicip". Registered server, "serverpublicip" set to the 81.174.155.168:21124, "centralserver" as 127.0.0.1:22824, port remapped by the router
image
image
Registered server log:
```
Feb 13 15:28:52 fs-peter Jamulus-wip[1089209]: - selected port number: 56850
Feb 13 15:28:52 fs-peter Jamulus-wip[1089209]: - maximum number of channels: 4
Feb 13 15:28:52 fs-peter Jamulus-wip[1089209]: - central server: 127.0.0.1:22824
```
Directory server log:
```
Feb 13 15:24:45 fs-peter Jamulus-Central8[1088417]: Requested to register entry for 127.0.0.1:56850 (127.0.0.1:56850): Test Server (21124)
```

So sadly, this fails -- and silently, which is bad.

pljones: I also notice that _nothing_ is logged when using `--serverpublicip`, which goes against normal practice for server start up parameters.  I should have spotted that before, when reviewing the code, rather than now, when reviewing the docs....

pljones: I can't explain the Jamulus Explorer output for the registered server, here, though. @softins, any ideas?

softins: Will follow up soon...

ann0see:
Mentioning @hoffie here since he introduced this code and might want to comment. I'll unsubscribe from this issue.

hoffie:
> This is when both servers are inside the LAN, behind the NAT wall

Right

> Mentioning @hoffie here since he introduced this code and might want to comment. I'll unsubscribe from this issue.

Happy to help, but I'm a bit confused by the tests.

@pljones:
Do you think that something is broken what has been working before? If so, what exactly?
Do you think that something does not work which you expected to work after the PR?

wiki/en/en-Command-Line-Options.md
hoffie: > I also notice that nothing is logged when using --serverpublicip, which goes against normal practice for server start up parameters. I should have spotted that before, when reviewing the code, rather than now, when reviewing the docs....

I wasn't aware, I'm sorry. I'll fix that. See https://github.com/jamulussoftware/jamulus/pull/993

> Directory, no "serverpublicip". Registered server, "serverpublicip" set to the 81.174.155.168, "centralserver" as 127.0.0.1:22824, port remapped by the router
> Feb 13 15:24:45 fs-peter Jamulus-Central8[1088417]: Requested to register entry for 127.0.0.1:56850 (127.0.0.1:56850): Test Server (21124)

This sounds strange. Can you tell me what versions you are running?
For the new scenario (regular server + directory behind NAT + client from public internet) to work, the following is needed:
- The regular server needs to use Jamulus master code and `--serverpublicip` with the appropriate public IP.
- The directory needs to use Jamulus master (but no special command line options). I suspect that this requirement might not have been met in your test?

This is the scenario which @genesisproject2020 has validated (I also tested behind NAT + Jamulus Explorer as a client).

pljones: > The directory needs to use Jamulus master

What does this mean?  If you mean the code, then you can see the commit in use in the screen shots.
image

The problem only happens when trying to use port forwarding as well.  It's just the `--serverpublicip` syntax doesn't support supplying a public _port_ as well as a public IP.

hoffie: > What does this mean? If you mean the code, then you can see the commit in use in the screen shots.

Sorry, missed that. Everything looks good then.

> The problem only happens when trying to use port forwarding as well. It's just the --serverpublicip syntax doesn't support supplying a public port as well as a public IP.

Ah, right, you mean forwarding a specific port on your public router/firewall to a different internal port. I talked to @genesisproject2020 about adding support for this scenario as well, but this would require a change in the protocol (PROTMESSID_CLM_REGISTER_SERVER). For now, it wasn't deemed necessary.

So yes, the current code/protocol only supports supplying a different IP address. Both internal and external ports have to match.

pljones:
> Do you think that something is broken what has been working before? If so, what exactly?

No, this is new.  The breakage that exists existed before as I've shown.

> Do you think that something does not work which you expected to work after the PR?

I _thought_ this was going to allow both public IP and port to be supplied (no idea why - I clearly was in too much of a rush when reading the code).  It doesn't -- it works the way it's written to work.

What it still doesn't do (even with the logging that's been added) is syntax check the argument and complain if it's invalid.  So I can _try to_ supply public IP and port `--serverpublicip 81.174.155.168:21124` but this then gets ignored.

So all I'm really saying is "it would be nice if it could handle the port, too" -- either log a complaint that it's not supported or actually, better yet, have the directory server expose the port when a request come in from the outside world.

wiki/en/en-Command-Line-Options.md
gilgongo: OK so in terms of my suggested wording for the docs (which in this PR is supposed to be a summary, with the [full description here](https://github.com/jamulussoftware/jamuluswebsite/commit/8258ddb4ab912ef3ce12c2894d2807961301c4e6)) - does anything need to change?

pljones: It needs to say _how to use it_, not what it does.

gilgongo: The pattern we have adopted for other things on this page is to summarise the use here and then link to the corresponding docs. If we try to put all the explanation into the table it'll get very cluttered.

pljones: > `--clientname`     | Window title and JACK client name

So something like `--serverpublicip` | Supply the server public IP address

gilgongo: Sorry, how does this work? I've updated that wording but it;s in a separate PR. https://github.com/jamulussoftware/jamuluswebsite/pull/304

pljones (approved):
It's good enough.  The fuller docs need to say when it makes a difference.  May it could say "see notes for where needed" here, or something - but I'm approving as is.

hoffie:
> What it still doesn't do (even with the logging that's been added) is syntax check the argument and complain if it's invalid. So I can _try to_ supply public IP and port `--serverpublicip 81.174.155.168:21124` but this then gets ignored.
> 
> So all I'm really saying is "it would be nice if it could handle the port, too" -- either log a complaint that it's not supported or actually, better yet, have the directory server expose the port when a request come in from the outside world.

Good point. I have implemented a quick validation to make it less confusing (https://github.com/jamulussoftware/jamulus/pull/1004). If there's need to support the port as well, I can look into it. As said, it might need a protocol change (or at least a field re-definition). I suggest to do that only as use cases come up in order to keep the codebase as simple as possible.

pljones:
> If there's need to support the port as well, I can look into it. As said, it might need a protocol change (or at least a field re-definition).

Getting port mapping "working" for a registered server (whether the directory server has the same public IP or not) has been on my list of things to look at for a while -- I've had no time available to actually do any work on anything recently, though.

Issue #264, opened 2021-01-24
[Feature] Android installation explanation


ann0see:
**Is your feature request related to a problem? Please describe.**

GitHub actions is going to provide apk builds thankss to @nefarius2001 https://github.com/corrados/jamulus/pull/880#issuecomment-766328897

**Describe the solution you'd like**

A page (like for windows installation) which tells the user how to install the APK but also tells him that it's in early beta

**Describe alternatives you've considered**

We could also not mention it since it's not 100% stable yet (but you can listen at least).

gilgongo:
Have we had any reports of anyone actually using Android? If so, what hardware?  We also probably want to make sure they're using an Ethernet adaptor.

I've always thought the Android build was just a novelty thing and not actually for serious use, but it would be nice if I was wrong as it obviously broadens the usage base if it's feasible.

ann0see:
I did check android and at least listening is possible with a 5 year old Android device over 4G. Wifi causes too many dropouts.

Hk1020:
I‘d really like to know what your plans are in this regard. I can’t think of a way to use a mobile for Jamulus. Would it work with the built in mic and speakers? On the other hand having this on mobile would be huge. Though personally I‘d like to see this on iOS, I don’t have any Android stuff.

gilgongo:
I think it's really just at proof-of-concept as far as I know. At least, I can't see somebody whipping out their Huawei and joining a jam session on their ukelele without some extra hardware to handle both headphones and Ethernet. 

I suppose **possibly** with 5G? I agree it would be huge though.

But on the subject of this ticket - perhaps we should look at having an Android installation page, if only so that it encourages some alpha testing. I know hardly anything about it though. Would you have to compile and then side-load an APK? Seems pretty deeply geeky.

nefarius2001:
Pre-compiled APKs will be part of the autobuilds for releases.

I used Jamulus on mobile platform (tablet) with WiFi working well (in & output).

gilgongo:
@nefarius2001 Can you describe the installation process? If use on Android devices is feasible at this time, we should probably write an installation guide for it. Also, what is the reason why a tablet would work well using wifi but a desktop machine typically would not in most people's experience?

nefarius2001:
> Can you describe the installation process? 

There are many instructions on the web for APK installation, and Jamulus needs no extrordinary steps. Only a link to the downloadable APK-file.
[...]

PR #297, opened 2021-02-14
Add basic android guide

ann0see:
Solves #264 I added this guide but also made clear that it's in early beta. I think adding this (although it seems to be in early beta) does give a big value to users.

### DIFF
--- a/wiki/en/en-Installation-for-Android.md
+++ b/wiki/en/en-Installation-for-Android.md
@@ -17,7 +17,7 @@ Although you **can** install Jamulus on Android devices (and hear sound), we str
 
 If you do want to try Jamulus on Android:
 
-1. Allow the installation of apps from unknown sources (look in Settings>Security. Note: this can vary for device and OS-version)
+1. Allow the installation of apps from unknown sources (look in Settings>Security. Note: The exact way to do this depends on your device and OS-version)
 1. Download and install the APK from the [GitHub release page](https://github.com/corrados/jamulus/releases/tag/latest)
 1. You should now be able to run Jamulus on your Android device
 
--- a/wiki/en/en-Installation-for-Android.md
+++ b/wiki/en/en-Installation-for-Android.md
@@ -1,0 +1,33 @@
+---
+layout: wiki
+title: "Installation for Android"
+lang: "en"
+permalink: "/wiki/Installation-for-Android"
+---
+
+# Installation for Android
+
+Make sure you've already read the [Getting Started](Getting-Started) page.
+
+## Things to note about Android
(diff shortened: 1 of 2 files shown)


pljones:
I'm not sure I'd go as far as "early beta".  It's more "proof of concept", in my view.  "Beta" tends to mean there's effort being made in bringing about a release and then maintaining it.  I don't see that happening.

ann0see:
Fair point. I think once this is on the website, some people will jump in and maybe try to bring it further. Thanks to nefarius, we now have a working and install-able version now, so it will be much more accessible. iOS is more of a proof of concept (headless server only) while android is accessible and to some extent works.

Is there something between beta and proof of concept?

pljones:
Well, "alpha".  But even that implies it's got some development effort behind it.  I'd leave it at proof of concept until someone does step up to get involved.

ann0see:
Ok so let's change it to proof of concept or PoC for the button?

ann0see:
Ok. @pljones done.

pljones (approved):

gilgongo (approved):

wiki/en/en-Installation-for-Android.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Things to note about Android
+
+Although you **can** install Jamulus on Android devices (and hear sound), we strongly recommend **not** doing so. Sound quality - especially over WiFi - is usually bad and latency is high. If you have don't own a PC, we suggest you to buy a [Raspberry Pi](https://www.raspberrypi.org/){: target="_blank" rel="noopener noreferrer" } which is an inexpensive and small device that performs very well with Jamulus. Android support is just a proof of concept.
+
+## Install the Android PoC
+
+If you do want to try Jamulus on Android:
+
+1. Allow the installation of apps from unknown sources (look in Settings>Security).
```
nefarius2001: maybe add a hint, that this can vary for devices and OS-versions.

wiki/en/en-Installation-for-Android.md
```diff
@@ -0,0 +1,33 @@
...
+1. Download the APK from the [GitHub release page](https://github.com/corrados/jamulus/releases/tag/latest)
```
nefarius2001: after download, the apk needs to be installed. In my case: open list of downloads in chrome, klick on the filename of APK, confirm that I want to install.

nefarius2001 (changes requested):

ann0see:
@nefarius2001 could you please review again? I‘ve added your proposal.

PR #299, opened 2021-02-14
Revert "Update"

ann0see:
Reverts jamulussoftware/jamuluswebsite#296

Sorry @dzpex I think we need to wait a little bit before this can be done. @gilgongo should we open the translate branch in two days and then start a new translation process?

### DIFF
--- a/_data/it/general.yml
+++ b/_data/it/general.yml
@@ -11,15 +11,3 @@ nav:
   altJamulusIcon:    "Icona di Jamulus. Links alla pagina di Wiki"
 wiki:
   titleAdd: "– Jamulus Wiki"
-kb:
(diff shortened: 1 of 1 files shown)


gilgongo:
> should we open the translate branch in two days and then start a new translation process?

Is that enough time to get those pull requests dealt with? We just call time on them if not done in the next 72 hours I guess. There's nothing there that a new release actually demands as far as I know.

ann0see:
Hmm. Not sure I'd like to merge the Android and Linux guides since this is new for the next version. Also we could merge ignotus666 PR and continue working on it.

dzpex:
ok no problem...i wait your start...

PR #300, opened 2021-02-15
Test prop bot

ann0see:


ann0see:
@gilgongo i‘ve now removed the propose bot since it doesn’t seem to do what I thought it did. I wanted to have a spell checker, but it seemed to do too much

gilgongo:
Yes, it's odd because it doesn't seem to be able to ignore things like markup and stuff. Maybe there's one that knows about stuff you get in .md files a bit more.

PR #302, opened 2021-02-15
Better local monitoring test

gilgongo:
From [Chris Rimple](https://sourceforge.net/p/llcon/discussion/software/thread/166288d63e/#e047)

### DIFF
--- a/wiki/en/en-Client-Troubleshooting.md
+++ b/wiki/en/en-Client-Troubleshooting.md
@@ -18,13 +18,14 @@ Start with the simple stuff: make sure your instrument/microphone and headphones
 
 You can test whether you are hearing your signal correctly by doing the following:
 
-1. Get your audio into your computer.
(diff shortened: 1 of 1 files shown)


ann0see (approved):
Yes. This sounds logical. There's another todo: the recording feature might need to be updated (siguser2 command on a headless server etc. It's not well documented at the moment)

gilgongo:
[There's this](https://jamulus.io/wiki/Server-Linux#controlling-recordings) (also linked to from the Windows/Mac server page) - seems OK to me. Is there something missing?

ann0see:
Argh. Missed that. I thought it was on the tips page. Probably we need to re-think the structure of the Wiki to make it clear where you can find that. I didn't look at the Linux server page.

gilgongo:
Hard to know how we might make it clearer. The use of siguser only relates to server operators running Linux headless, who (presumably) would consult the Linux headless server docs, where the operation of recording is described. I suppose we could link to it from Tips & Tricks, but where do you draw the line?

Server operators running Linux with the GUI, and all other OSes, don't need to know about the siguser method.

Issue #9, opened 2020-10-25
[Feature] Make wiki structure more understandable by using a link which shows parent pages (breadcrumb)


ann0see:
**Is your feature request related to a problem? Please describe.**
Currently if you are on a wiki page for the first time, it might be confusing if you clicked through multiple pages and are lost in the structure. e.g. if you are on the Choosing-a-server-type page you might be confused where you are and don't know how to come back since the navigation doesn't link this page.

**Describe the solution you'd like**
It would be great to have a breadcrumb or something similar like in the current issue.

**Describe alternatives you've considered**
Add a breadcrumb to every page manually (this is not a clean solution)
**Additional context**
See this:
![Screenshot_2020-10-25 corrados jamulus](https://user-images.githubusercontent.com/20726856/97116999-5b5a9000-1701-11eb-8dc2-d57f57a14910.png)

gilgongo:
I think this may be hard to automate, and not having the breadcrumbs is a bit of a problem. Now that we have the changes branch, I'll see if I can just add them in manually until we can have a better solution.

ann0see:
I think we have two ways to add breadcrumbs:

1. loop through all pages and build a structure by front matter (not recommended and probably messy)
2. generate a folder structure and build the sites with this structure (e.g. https://jekyllcodex.org/without-plugin/breadcrumbs/)

ann0see:
For 2 see the breadcrumb branch. But it does mean a restructuring of the pages (but I don't think that's a big problem)

gilgongo:
I was thinking maybe making a special "breadcrumb" link style and putting it over the H1 if we can keep the H1 in the same place. Or maybe put it in the blue header bar next to the logo? 

Also BTW there is the question of whether you repeat the name of the current page in the leaf of the breadcrumb or not. People do it both ways. On the old wiki I did a sort of "summary title" for the leaf to re-enforce the meaning of the H1.  One thing about automating the breadcrumbs is whether we can control that, and also whether we get "redundant" breadcrumbs on pages that don't need them.

ann0see:
Ok. So I suggest you add them if my PR to main is merged.

ann0see:
Fixed by @gilgongo

PR #306, opened 2021-02-16
Breadcrumb

gilgongo:
Setting up a breadcrumb include file (see https://github.com/jamulussoftware/jamuluswebsite/issues/9).  Put this above the H1 when needed, for example:

`{% include breadcrumb.html root="Using Jamulus" branch1="Getting Started" branch1-url="Getting-Started" %}`

I think this will do the job. I can put them on the relevant pages (on this branch?) now if so.

### DIFF
--- a/_includes/breadcrumb.html
+++ b/_includes/breadcrumb.html
@@ -8,4 +8,7 @@
  {% if include.branch2 %}
    &gt; <a href="{{ include.branch2-url }}">{{ include.branch2 }}</a>
  {% endif %}
+  {% if include.branch3 %}
+   &gt; <a href="{{ include.branch3-url }}">{{ include.branch3 }}</a>
+ {% endif %}
 </div>
--- a/_includes/breadcrumb.html
+++ b/_includes/breadcrumb.html
@@ -1,0 +1,11 @@
+<div id="breadcrumb">
+ {% if include.root %}
+   <strong>{{ include.root }}</strong> |   
+ {% endif %}
+ {% if include.branch1 %}
+   <a href="{{ include.branch1-url }}">{{ include.branch1 }}</a> 
+ {% endif %}
+ {% if include.branch2 %}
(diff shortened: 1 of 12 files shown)


_includes/breadcrumb.html
```diff
@@ -0,0 +1,11 @@
+<div id="breadcrumb">
+ {% if include.root %}
+   <strong>{{ include.root }}</strong> |   
+ {% endif %}
+ {% if include.branch1 %}
```
ann0see, line 5: Great idea! Do you think a loop would be better here so more than two items can be created?

gilgongo: Not sure how to create a loop, but I guess if we ever wanted to get more than three levels deep we could just change the include to accept that (or just a "branch3"....)

ann0see: Yes. I think, we should include three levels now. It's better to have too many than not enough ;-).

gilgongo: OK I'll add a third (BTW just been reading about loops but I'm not sure how to get the alternating linktext/link thing working with that). 

I'll put them on the page then and add to this branch. Would be good to get that in before translation starts.

ann0see: > ll put them on the page then and add to this branch. Would be good to get that in before translation starts.

Agree. 

Same for the linux installation guide.

ann0see (commented):
Great!

_includes/breadcrumb.html
gilgongo: OK I've done that now I think. Note that it's quite a "conservative" breadcrumb because it's not on every page, just the one's that are sort of "children" of others.

ann0see: Great! I've also edited the Linux install pages (client/server). But I think they will soon need to be re-written because they look quite messy...

Issue #31, opened 2020-10-30
Version control in Jamulus documentation


ann0see:
Currently we suffer from unstructured editing, have different versions in languages,... We should discuss how we solve this problem (= ensure every language is up to date).

To make development easier and faster, @gilgongo suggested to make a new branch.
I think that's a good idea. We could split it into 3 branches:
main: the branch which gets built to github pages. We only submit changes here if everything was approved.
Translation: The step before approval. Here translation begins. No changes
Development: Active development, restructuring,...

So a normal workflow would look like this:

1. Add a new page (english only): commit to the development branch.
2. After a period of time it gets merged into translation for the translators
3. If everything was checked, we merge it into main.

> Of course we could also create a "development" and a "translation" branch while the "translation" can not be changed content wise.

_Originally posted by @ann0see in https://github.com/jamulussoftware/jamulussoftware.github.io/pull/28#issuecomment-719692886_

Another option would be the projects tab on github (we could document changes and todos there).

ignotus666:
Sounds good to me. The Spanish translation is currently up to date with the English version.

gilgongo:
Do we need to avoid peppering the development branch with small changes? Or is that not an issue if we do a regular merge into the translation branch - I would imagine that means translators then get one diff to work on? If so then that's good. I'd like to be able to comment changes too so that we can communicate why they are being made, and maybe get feedback on improvements to the changes.

BTW I'm assuming we do the development to translation merge at some predictable interval, if only to send out a reminder to the translators. Monthly? Or maybe perhaps a week or so before a release, which is what Volker does with the app?

ann0see:
I think we should follow the release period of the app (for me it makes sense). I don't know how much work all these branches would make (especially management, of pull requests,...)

gilgongo:
Would tags be easier/better for this? I don't actually have any experience with tags so maybe not, but could we tag changes on a "translation" branch with the name of the upcoming release, thereby only having two branches?

ann0see:
We could probably also use milestones. They're better suited for that.
--- 
We could assign translators to new PRs they would then be able to translate all files before a merge.

Nevertheless, I think a fixed period for translating following the app is better and easier for translators.

ann0see:
> @gilgongo said:

I'm not sure if we resolved exactly how we'd be doing documentation changes other than saying we'd do them all prior to each software release. I'm keeping a list in a Google doc for the moment (and will add this). But what are the options?

[...]

Issue #14, opened 2020-10-28
[Translate] English version littered with {: target="_blank" rel="noopener noreferrer"} and some other issues


ignotus666:
Hi, I'm working on getting the Spanish version ready and have noticed that the English version is littered with this: "{: target="_blank" rel="noopener noreferrer"}" all over the place. Do I leave them there? Delete them?

Also, I know it has been done with the best of intentions, but I'm seeing a fair few additions in English that contain many mistakes, with some entries being close to unintelligible. E.g. there's a section in Client-Troubleshooting (Try to connect - Message in Jamulus Client on MacOSX) that is virtually unreadable and looks more like a kind of bug report - leaving aside linguistic issues I don't understand what it's purpose is in the wiki. Then there are syntax errors scattered here and there which will require some proofreading.

I'm also seeing that changes keep being made constantly to the files in English—my 2-day-old fork is already obsolete and I'll have to update all the files again to reflect those changes. I think there should be a "feature freeze" before encouraging translations because otherwise we just keep playing catch-up with an ever-moving target. I know there's a lot going on but I think we should use a more structured approach than the current scattergun one—e.g. bundle edits and commit them on a weekly or fortnightly basis after reviewing them carefully.

What do you guys think?

ann0see:
I agree. I'd suggest to make a freeze in 2 days then we can start translating.

@gilgongo should have a look at that. 

I think we should also discuss the use of opening pages in a new tab. When do we need to do that? 

I put it mainly on links to external sites.

gilgongo:
> I'm seeing a fair few additions in English that contain many mistakes, with some entries being close to unintelligible

@ignotus666 That makes two of us! When moving the content from the old wiki to the new site as I found various drive-by edits had been made that I either removed or re-wrote. There's no way of detecting changes on the Github wiki automatically and I didn't realise how random it had become. I'd not looked at the troubleshooting page - good spot! I'll have a look.

 > I think we should also discuss the use of opening pages in a new tab. 

Yes, we could adopt an "external site, new tab" convention perhaps. Not sure about pros/cons though. I think there can be accessibility issues with new tabs but I'm not sure.

ann0see:
I have to revert a few commits. So before editing, please wait a few minutes.

ann0see:
Ok. I was too late. But I'll add your changes later.

ann0see:
Done. It's all back to normal now.

gilgongo:
[...]

PR #308, opened 2021-02-17
Add GitHub actions binary links to config.yml file

### LINKED ISSUE #307: [Feature] How should we handle binary files from Autobuild?
**Is your feature request related to a problem? Please describe.**

Once the new build process is finished it will be possible to have direct download links to the binaries. 
@nefarius2001 and me agreed that it would be great to include them in the website so that a new user doesn’t need to get in touch with another website (SourceForge). Problem: the link to the binaries will change once a new version is released and will have to be updated on the website. 

**Describe the solution you'd like**

An easy way to change the download links can be achieved by saving them in the _config.yml file and then use a variable on all pages where these URLs should be included. 

**Describe alternatives you've considered**

A permanent URL to the binary. 

The current approach (automatically compile for "latest" and label the binaries \*latest\* & upload to a release called "latest" on GitHub has been dismissed since it involves too much complexity on the Autobild side. 

Nefarius proposed some kind of redirect but this doesn’t seem to be possible cleanly with GH pages (GitHub always returns HTTP 200 and the current redirect implemented here is a hack).

Any thoughts here?

**Additional context**

https://github.com/jamulussoftware/jamulus/pull/978

ann0see:
Solves: #307 

Probably still worth a discussion. This is a quick and easy approach.

### DIFF
--- a/wiki/en/en-Installation-for-Windows.md
+++ b/wiki/en/en-Installation-for-Windows.md
@@ -10,7 +10,8 @@ permalink: "/wiki/Installation-for-Windows"
 # Installation for Windows
 Make sure you read the [Getting Started](Getting-Started) page.
 1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have an external sound card, you probably need to install [this free ASIO driver (ASIO4ALL)](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"} before installing Jamulus.
-1. [Download Jamulus](https://sourceforge.net/projects/llcon/files/latest/download){: .button target="_blank" rel="noopener noreferrer"}
+1. [Download Jamulus]({{ site.download_root_link }}{{ site.download_file_names.windows }}){: .button}\\
+**Mirror 2:** [SourceForge](https://sourceforge.net/projects/llcon/files/latest/download)
 1. **Install Jamulus**: Double click the installer to launch it. If you get a warning notice from SmartScreen, click on "More info" and "Run anyway" to install Jamulus. (If you grabbed a new version of Jamulus and are one of the first people who downloaded it, Jamulus won't be whitelisted by SmartScreen yet. We don't pay for code signing.)
 1. **Run Jamulus**. Now you should be able to use Jamulus just like any other application.
 1. **Setup your sound card**. When you're done, you need to setup your audio hardware. Have a look at how to setup ASIO4ALL if you you're using that and/or [Hardware Setup](Hardware-Setup).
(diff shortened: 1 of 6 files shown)


wiki/en/en-Installation-for-Windows.md
```diff
@@ -10,7 +10,8 @@ permalink: "/wiki/Installation-for-Windows"
 # Installation for Windows
 Make sure you read the [Getting Started](Getting-Started) page.
 1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have an external sound card, you probably need to install [this free ASIO driver (ASIO4ALL)](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"} before installing Jamulus.
```
pljones, line 12: Can this also note that "It's well worth installing the ASIO4ALL Off-line Settings tool when installing ASIO4ALL", as well?  It lets people set up before running Jamulus - i.e. not having Jamulus make wrong decisions about what devices are available.  Indeed, I'd even say this is the point to say "If you're installing ASIO4ALL and haven't used it before - go read **this**" and link to something that will explain what's going on...  Once we have it.

pljones (commented):
So this shouldn't be _merged_ until after the _latest_ tag gets moved?

ann0see:
> So this shouldn't be merged until after the latest tag gets moved?

Yes it should! If we merge it to changes, it will not go live instantly. jamulus.io is built on the files on the release branch. Once a Jamulus release is planned, release will be merged into translate and once the translations are finished and Jamulus gets released, translate will be merged into release which then "publishes" the site.

wiki/en/en-Installation-for-Windows.md
ann0see: > If you're installing ASIO4ALL and haven't used it before - go read this

Yes. That's a good suggestion. Can you open a PR to the changes branch with this proposal?

Concerning the off-line settings, I'm not exactly sure if that helps. I think there was a discussion somewhere that it doesn't do what we thought it would do: You can't setup the A4A in/outputs with off line settings since A4A saves configuration on per app basis.

pljones:
Ah, right...  Is there a document for maintaining the jamuluswebsite, i.e. with the process needed to make changes and release them? (Not related to this, of course.)

ann0see:
Not exactly, but see https://github.com/jamulussoftware/jamuluswebsite/blob/changes/CONTRIBUTING.md and probably https://github.com/jamulussoftware/jamuluswebsite/issues/31 and https://github.com/jamulussoftware/jamuluswebsite/issues/14

But information seems to be cluttered everywhere.

wiki/en/en-Installation-for-Windows.md
pljones: > You can't setup the A4A in/outputs with off line settings since A4A saves configuration on per app basis.

Oh - OK, no point then, agreed.  I wonder why it exists?  (Maybe that's why it's so hard to install....)

pljones (approved):

ann0see:
Merged. Sure, we might need to move to another solution soon.

PR #325, opened 2021-02-22
Clarify ASIO4ALL to new UI

ann0see:
A few smaller changes which I think should be included in the next release (we've also had a UI change in this area)

### DIFF
--- a/wiki/en/en-Installation-for-Windows.md
+++ b/wiki/en/en-Installation-for-Windows.md
@@ -9,7 +9,7 @@ permalink: "/wiki/Installation-for-Windows"
 
 # Installation for Windows
 Make sure you read the [Getting Started](Getting-Started) page.
-1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have an external sound card, you probably need to install [this free ASIO driver (ASIO4ALL)](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"} before installing Jamulus.
+1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have an external sound card, you probably need to install a generic driver like [ASIO4ALL)](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"}. For more information scroll down to the [ASIO](#asio) section.
 1. [Download Jamulus]({{ site.download_root_link }}{{ site.download_file_names.windows }}){: .button}\\
 **Mirror 2:** [SourceForge](https://sourceforge.net/projects/llcon/files/latest/download)
 1. **Install Jamulus**: Double click the installer to launch it. If you get a warning notice from SmartScreen, click on "More info" and "Run anyway" to install Jamulus. (If you grabbed a new version of Jamulus and are one of the first people who downloaded it, Jamulus won't be whitelisted by SmartScreen yet. We don't pay for code signing.)
@@ -29,7 +29,7 @@ Jamulus uses [ASIO](https://en.wikipedia.org/wiki/Audio_Stream_Input/Output) to 
 ### Setting up ASIO4ALL
 *ASIO4ALL is not recommended if your sound card/audio interface provides its own ASIO driver, since that might have better latency.*
 
-For ASIO4ALL ([download](http://www.asio4all.org){: target="_blank" rel="noopener noreferrer"}), you may need to experiment a bit to find the right configuration since every computer is a bit different.
+For ASIO4ALL, you may need to experiment a bit to find the right configuration since every computer is a bit different.
 
 **Tip:** Set up your sound card while you're [connected to a server](Onboarding#2-connecting-to-a-server) to hear your instrument or voice and check if everything is correctly set up; but first read on.
 
@@ -41,15 +41,16 @@ Before you start with Jamulus:
 ### How to set up ASIO4ALL inputs (Guide)
 
 1. Open Jamulus's settings
-1. Go to _"ASIO Settings"_ (column on the left)
+1. Go to _"ASIO Device Settings"_ (column on the left; directly under the selection of the driver)
 1. Enable _advanced view_ in ASIO4ALL (click the tool icon on the bottom right)
 1. Enable only the sound card you want to use by clicking on the button next to its name
 1. Open your sound card inputs/outputs by clicking the _plus icon_ next to this sound card
-1. Now enable the correct inputs/outputs in the list under your sound card and disable everything else
+1. Now enable the correct inputs/outputs in the list under your sound card and disable everything else. You can hover over the inputs/outputs to see which of both they are and if they support the required sample rate for Jamulus of 48kHz (DVD quality).
 
-**Hint:** Many internal sound cards are named Realtek High Definition Audio, Conexant or similar.
-
-Concerning the inputs/outputs: Headphones and speakers are often labelled as output and microphones as inputs or capture. Stereo Mix is usually not the input/output you're looking for and therefore you can switch it off if you see it.
+**Hints:**
+1. Many internal sound cards are named Realtek High Definition Audio, Conexant or similar.
+1. Headphones and speakers are often labelled as output and microphones as inputs or capture.
+1. Stereo Mix is usually not the input/output you're looking for and therefore you can switch it off if you see it.
 
 ### Having trouble setting up ASIO4ALL?
 
--- a/wiki/en/en-Installation-for-Windows.md
+++ b/wiki/en/en-Installation-for-Windows.md
@@ -9,7 +9,7 @@ permalink: "/wiki/Installation-for-Windows"
 
 # Installation for Windows
 Make sure you read the [Getting Started](Getting-Started) page.
-1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have an external sound card, you probably need to install a generic driver like [ASIO4ALL)](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"}. For more information scroll down to the [ASIO](#asio) section.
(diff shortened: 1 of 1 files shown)


wiki/en/en-Installation-for-Windows.md
```diff
@@ -41,15 +41,16 @@ Before you start with Jamulus:
...
-1. Now enable the correct inputs/outputs in the list under your sound card and disable everything else
+1. Now enable the correct inputs/outputs in the list under your sound card and disable everything else. You can hover over the inputs/outputs to see which of both they are and if they support the required sample rate for Jamulus of 48kHz (DVD quality).
 
-**Hint:** Many internal sound cards are named Realtek High Definition Audio, Conexant or similar.
-
-Concerning the inputs/outputs: Headphones and speakers are often labelled as output and microphones as inputs or capture. Stereo Mix is usually not the input/output you're looking for and therefore you can switch it off if you see it.
+**Hints:**
+1. Many internal sound cards are named Realtek High Definition Audio, Conexant or similar.
+1. Headphones and speakers are often labelled as output and microphones as inputs or capture.
+1. Stereo Mix is usually not the input/output you're looking for and therefore you can switch it off if you see it.
```
hoffie:
```suggestion
1. Stereo Mix is usually not the input/output you're looking for and therefore you should switch it off if you see it.
```

In my experience after onboarding my choir: Stereo mix has been the reason for getting no input (or worse: feedback!), so I would be more explicit suggest disabling it.

wiki/en/en-Installation-for-Windows.md
```diff
@@ -9,7 +9,7 @@ permalink: "/wiki/Installation-for-Windows"
 
 # Installation for Windows
 Make sure you read the [Getting Started](Getting-Started) page.
-1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have an external sound card, you probably need to install [this free ASIO driver (ASIO4ALL)](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"} before installing Jamulus.
+1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have an external sound card, you probably need to install a generic driver like [ASIO4ALL)](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"}. For more information scroll down to the [ASIO](#asio) section.
```
hoffie: I wonder if we could make it even clearer where to click on the ASIO4ALL website.
- It seems like there is no dedicated download page and one has to scroll down quite a bit as there are recent news about an Intel-SST-related issue.
- The latest version (2.14) is only available in English, so people go for the latest version available in their language (2.13). However, 2.13 lacks an important bugfix. So I wonder if we should document it as "Get at least version 2.14"?
- I was shocked when I saw someone of my choir opening the ASIO4ALL site without an ad blocker. There are lots of ads and some of them are **really** misleading, if not malicious ("Driver download" in-page popup). A friend of mine even somehow ended up on a site which wanted him to enter his credit card details. I wonder if we should warn about that, too. That's a problem in SourceForge, too, btw.

```suggestion
1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have an external sound card, you probably need to install a generic driver like [ASIO4ALL)](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"}. Get at least version 2.14, even if is not available in your language. The proper download is called "ASIO4ALL 2.14 – English" (or newer) and has a flag next to it. ASIO4ALL is free. If you are asked for your credit card details or any other personal data, you have likely clicked on the wrong link. For more information scroll down to the [ASIO](#asio) section.
```

I'm unsure how many details we should add (i.e. specific versions), but this is a common preparation step which is required by many (often unexperienced) users, I feel somehow responsible for protecting our usres.

hoffie (approved):
Looks good to me, so, approving. I agree we should definitely get the updated button label in before the next release.

I have some added some suggestions for further improvements, but I'm also fine if those are added in a later PR.

wiki/en/en-Installation-for-Windows.md
ann0see: Ok. Added your suggestion. Can still be re-thought.

wiki/en/en-Installation-for-Windows.md
ann0see: > The latest version (2.14) is only available in English, so people go for the latest version available in their language (2.13). However, 2.13 lacks an important bugfix. So I wonder if we should document it as "Get at least version 2.14"?

For me it was the exact opposite. V 2.14 was less stable than v 2.13. That's why I usually suggest v 2.13.

> I was shocked when I saw someone of my choir opening the ASIO4ALL site without an ad blocker.

Interesting fact. I didn't know that there were ads? Are they part of the new page too?

I agree that the download link is hidden.

hoffie: > > The latest version (2.14) is only available in English, so people go for the latest version available in their language (2.13). However, 2.13 lacks an important bugfix. So I wonder if we should document it as "Get at least version 2.14"?
> 
> For me it was the exact opposite. V 2.14 was less stable than v 2.13. That's why I usually suggest v 2.13.

Interesting, then we should probably keep it as-is and don't make a version suggestion.

> > I was shocked when I saw someone of my choir opening the ASIO4ALL site without an ad blocker.
> 
> Interesting fact. I didn't know that there were ads? Are they part of the new page too?

Didn't know about the ads either, guess, most technical people live in an ad-blocked filter bubble. :)
No idea if this is new or not. 
asio4all
Attaching a screenshot for reference.

gilgongo: Wow so many ads! Tempting to host our own download link and just update that from time to time ... :-)

ann0see: @gilgongo should we include this in the next release of the website too?

gilgongo: @ann0see I'm not sure what the implications of hosting the installer ourselves would be, if that's what you mean. It would certainly be a better experience though!

pljones: The author clearly asserts copyright and grants no redistribution licence.  So we can't without asking -- and if the author wants the ad traffic (and this last year has probably been a bit of a windfall), I can't see us getting permission.

ann0see: I think the license shown in the installer is a bit different. I think it allows redistribution

pljones: Here's the relevant quote (v2.14 - I don't have 2.13 to hand) - you could be right:

> YOU MAY FURTHER REDISTRIBUTE ONLY VERBATIM COPIES OF THIS SOFTWARE WITH THE FOLLOWING RESTRICTIONS:
>
> 1.) THE SOLE SOURCE OF PROFIT TO BE MADE (IF ANY) - INCLUDING PROFIT INTENTED TO BE MADE - FROM THIS SOFTWARE SHALL BE THE SERVICE OF ITS REDISTRIBUTION ONLY. THIS SPECIFICALLY PROHIBITS PLACING SOFTWARE ON ANY MEDIA SOLD TOGETHER WITH AUDIO HARD- AND SOFTWARE, OR SIMILAR PRODUCTS TO WHICH THIS SOFTWARE WOULD ADD SUBSTANTIAL MARKET VALUE.
> 2.) EXCEPTIONS TO 1. REQUIRE WRITTEN CONSENT FROM THE AUTHOR.

We don't intend to profit at all from Jamulus, not even from redistribution of ASIO4ALL.  The "specifically prohibts" would kick in if we did.

Anyone wanting to distribute Jamulus for profit would have to do it elsewhere.

gilgongo: OK so that does look like we can host the download ourselves then. It then only remains for us to keep an eye on version updates I suppose.

hoffie: Sounds good. One could even think about installing it from the Jamulus installer?
Any volunteers for that? Should probably be its own ticket by now :)

pljones (approved):

wiki/en/en-Installation-for-Windows.md
ann0see: I'd create a repo where we host the ASIO4ALL installer for now. I wouldn't include it on the website repo since it might make the repo too large.

pljones: 452KB for the 2.14EN.  It's not that big :).  What's the repo limit again?

ann0see: >  It's not that big :)

No. But I don't really like the idea of including binaries in this repo. I'd suggest to just host it on another repo (jamulussoftware/assets) since it would allow us to just push new releases easily.

pljones: image
There's a lot of binaries, some quite a bit bigger...  But yes - anything we expect will "never change" by the project itself shouldn't live in the main repos, I'd agree on that basis! :)

wiki/en/en-Installation-for-Windows.md
```diff
@@ -9,7 +9,7 @@ permalink: "/wiki/Installation-for-Windows"
 
 # Installation for Windows
 Make sure you read the [Getting Started](Getting-Started) page.
-1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have an external sound card, you probably need to install [this free ASIO driver (ASIO4ALL)](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"} before installing Jamulus.
+1. **Download and install an ASIO Driver**. It is recommended to use a sound card/interface with a native ASIO driver. If you don't have an external sound card, you probably need to install a generic driver like [ASIO4ALL](https://www.asio4all.org){: target="_blank" rel="noopener noreferrer"}. **Note:** Do not fall for  ads! Download the ASIO4ALL installer by scrolling down a bit! For more information scroll down to the [ASIO](#asio) section.
```
pljones: Two spaces between "fall for" and "ads!" - should only be one.

ann0see: Done.

wiki/en/en-Installation-for-Windows.md
ann0see: Ok. I'll set up another repo for this binary.

ann0see:
@hoffie @pljones I've now added a link to our website. Two versions of ASIO4ALL are now linked. While this isn't optimal, I think some people have reported more success with v2.13 while others have more with v2.14

PR #356, opened 2021-03-13
Reduce size of PNG images using OptiPNG 0.7.7

paulmenzel:


ann0see:
Hi Paul! Thank you very much for your first PR! Optimizing images is surely good. I haven't looked into them yet ;-).

From my first glance, this looks ok. Let's wait on others.

ann0see:
Meta: Proabably we need to merge this into release, not changes

gilgongo (approved):
Nice! main-screen-large.png is ten times smaller with no noticeable loss of quality. It's a pity we can't (I assume) automate this on a build script.

paulmenzel:
> Meta: Proabably we need to merge this into release, not changes

Do I need to do something here?

> It's a pity we can't (I assume) automate this on a build script.

There is the GitHub App [Imgbot](https://github.com/marketplace/imgbot), which does not cost anything for FLOSS projects.

gilgongo:
The issue with release vs changes is that the latter is meant for changes that affect the translations. Images are a bit of tricky one because we have (soon, it's not hooked up yet) images that can be localised per language.  But I think if we merge it all to release for now that should OK? BTW @ann0see do I take it that after we merge changes to release we then make a clone of release and make that into the new changes? I've not been putting anything into changes for that reason recently but I'm a bit unsure on that point...

I'll have a look at ImgBot, particularly as it's going to be up to the translators to do the localised screenshots over time.

ann0see:
The workflow is changes>translate*>release>changes

So you can modify changes as you like now since we'll merge translate into release.

gilgongo:
Not sure I understand. If today I commit to changes, and you do a commit to release a bit later, if tomorrow we do release>changes what happens to my change? Apologies, I'm not very good at understanding git.

ann0see:
@gilgongo I hope my message helped you understand the workflow?

ann0see (approved):
I think this should be ok. Waiting for others but you can merge it.

gilgongo:
Let's merge this and then maybe look at using [ImgBot](https://github.com/marketplace/imgbot) later if we want.

PR #360, opened 2021-03-15
ASIO4ALL update links

### LINKED ISSUE #358: Link new ASIO4ALL version
ASIO4ALL just got updated with important bug fixes: https://www.asio4all.org/

I think we should link the new version on the website as soon as possible (better include it in the upcoming release). I have the feeling that v2.15 fixes many annoying bugs so it should be linked as soon as possible.

ann0see:
Probably we'll need to merge this to release once 3.7.0 is out.

### DIFF
--- a/wiki/en/en-Installation-for-Windows.md
+++ b/wiki/en/en-Installation-for-Windows.md
@@ -25,9 +25,9 @@ Jamulus uses [ASIO](https://en.wikipedia.org/wiki/Audio_Stream_Input/Output) to 
 
 You can try two versions of ASIO4ALL. ASIO4ALL v2.14 includes a workaround for a bug which might break other functionality.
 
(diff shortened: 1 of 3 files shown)


wiki/en/en-Installation-for-Windows.md
```diff
@@ -25,9 +25,9 @@ Jamulus uses [ASIO](https://en.wikipedia.org/wiki/Audio_Stream_Input/Output) to
 
 You can try two versions of ASIO4ALL. ASIO4ALL v2.14 includes a workaround for a bug which might break other functionality.
```
hoffie, line 26: This should probably be updated then?

ann0see: This is still true, so I'd remove it once 2.15 is out of beta.

ann0see:
Tagging @dzpex @trebmuh and @gegeweb here since this will have to be part of the translation for 3.7.0 but isn't part of the initial PR.

Issue #359, opened 2021-03-13
Add Download links for 3.7.0


hoffie:
`Download now for [Windows](https://github.com/jamulussoftware/jamulus/releases/download/r3_7_0/jamulus_3.7.0_win.exe), [Mac](https://github.com/jamulussoftware/jamulus/releases/download/r3_7_0/jamulus_3.7.0_mac.dmg), [Debian/Ubuntu](https://github.com/jamulussoftware/jamulus/releases/download/r3_7_0/jamulus_3.7.0_ubuntu_amd64.deb) or [other platforms](https://github.com/jamulussoftware/jamulus/releases/tag/r3_7_0)`

- [ ] Update front page `mHTMLDownloadNow` in `*index*.html`
- [x] Decide if/how we can do that without going to a translation process again (Keep "other platforms" untranslated? ~Use a dictionary?~) -> Use English
- [ ] Any other places to update?

(Just opening this for tracking, feel free to assign to you)

gilgongo:
I assume we can do the change to the release branch, and maybe for each lang just paste in the English - don't want to delay the release for that really.

Issue #345, opened 2021-03-03
Remove reference to SourceForge on footer


ann0see:
Currently the [footer](https://github.com/jamulussoftware/jamuluswebsite/blob/5271c43ee108818634e7071b48098d04180a633f/_includes/footer.html#L14) links to SourceForge. Shouldn't we link the GitHub discussions forum there?

josecollado:
Which GitHub discussions?

ann0see:
Hi @josecollado 

it‘s this link: https://github.com/jamulussoftware/jamulus/discussions

I‘d say that also another icon (maybe a speech bubble or question mark should  be displayed. An icon from FontAwesome would be great.

ann0see:
Do you want to do it or should I? This will probably have to wait until the next release is out.

josecollado:
Hi @ann0see You can assign it to me, i could send a PR by tomorrow if that works. 

Also, would one of these work? 
** only the first two i can choose without paying ** 
<img width="1059" alt="Screen Shot 2021-03-16 at 7 22 23 PM" src="https://user-images.githubusercontent.com/37092658/111392464-035efe80-868d-11eb-993f-24d4fe7b95e6.png">

ann0see:
If you have to pay, I wouldn't choose any of them. Can you find some here: https://icongr.am/fontawesome ?

josecollado:
Sent a PR for the changes

ann0see:
Great! Thank you.

Issue #322, opened 2021-02-19
Create download page


gilgongo:
This to be the default URL for all platform downloads at https://jamulus.io/downloads/ (or similar URL).

Some examples of different approaches:

https://www.torproject.org/download/
https://www.plex.tv/downloads/
https://ubuntu.com/download/desktop
https://www.videolan.org/vlc/#download

See also https://github.com/jamulussoftware/jamulus/discussions/1023 and https://github.com/jamulussoftware/jamulus/issues/473

pljones:
I'll just say I prefer the Ubuntu one over the VLC one... :)

mulyaj:
Would this use the wiki layout?

gilgongo:
Only if we can't work out anther way :-)

mulyaj:
Made a mockup. Thoughts?

![Jamulus Download Mockup 1](https://user-images.githubusercontent.com/63881221/109870939-aac54580-7c38-11eb-9236-7d1f3c63a963.png)

pljones:
My _only_ reaction: there's a bit too much whitespace vertically - it's more 4x3 than 16x9...  (It may even be taller than wide...)  Looks great other than that!  Should be the jamulus.io front page.

gilgongo:
That works! The problem of course is actually working out the CSS to do it :-) My knowledge of the [framework we're using](http://fox-css.com/documents/) is almost nil I'm afraid.

mulyaj:
@pljones The page includes scroll, so that explains the height, I'll try to reduce the whitespace.

@gilgongo I'll try to figure something out.

gilgongo:
@mulyaj as part of https://github.com/jamulussoftware/jamuluswebsite/issues/192 (which I have in a branch but was waiting until things settled down a bit) I'm putting optional floats on images, but apart from that, I have no real idea how to go about any structural stuff on the site. It would be great if we we could get that top nav you have in that mockup, for example!

mulyaj:
@ann0see Would you have any thoughts on how we would integrate a new page? 

What I was thinking is that it could just be its own standalone HTML and CSS. Would it need a layout and a corresponding .md?

gilgongo:
Well, we're using Jekyll to do the translations (and other bits and pieces). And we'd want the download page to be multi-lingual like the rest of the site. 

[...]

PR #364, opened 2021-03-17
Change SourceForge links to Github

hoffie:
This updates generic download or discussion links.
It keeps direct links to discussions on SourceForge.

Some installation documents have not been updated in all translations.
Those will link to Github only. The secondary location at Sourceforge will be linked once they are updated as part of the translation process.

This does not update the direct download links yet as this will be done as part of #359.
This does not update the page footer yet as more involved changes are planned as part of #345.

### DIFF
--- a/_layouts/mainhomepage.html
+++ b/_layouts/mainhomepage.html
@@ -13,7 +13,7 @@
       "name": "Jamulus",
       "applicationCategory": "{{ page.ldjAppCategory }}",
       "operatingSystem": "Windows 10, macOS 10.11, Linux",
-      "discussionUrl": "https://sourceforge.net/p/llcon/discussion/",
+      "discussionUrl": "https://github.com/jamulussoftware/jamulus/discussions/",
       "license": "GPLv2",
       "creator": {
         "@type": "Person",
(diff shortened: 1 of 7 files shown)


_layouts/mainhomepage.html
```diff
@@ -22,7 +22,7 @@
       "isAccessibleForFree": true,
       "alternateName": "llcon",
       "thumbnailUrl": "https://jamulus.io/assets/img/jamulus-icon-2020.svg",
-      "downloadUrl": "https://sourceforge.net/projects/llcon/files/latest/download",
+      "downloadUrl": "https://github.com/jamulussoftware/jamulus/releases/latest",
```
ann0see, line 25: Are you sure about that? Shouldn't we wait until a download page is created: https://github.com/jamulussoftware/jamuluswebsite/issues/322

hoffie: Hrm, I thought we'd move all download links to Github? I agree that a real download page would be best, but as it isn't ready yet, I'd suggest replacing the already technical, confusing and ad-loaded SourceForge link by Github which is still technical but hopefully less confusing and without ads.

ann0see: Yes. Maybe most users are probably not clicking on this button anyway.

ann0see (approved):

pljones (approved):

PR #369, opened 2021-03-18
Updated Spanish index page and corrections

ignotus666:
Making this PR against "release" as it's for the last-minute change to the index page and a couple of small corrections elsewhere.


hoffie (approved):
Thanks!
Yes, I think merging directly to release is OK because it will be propataged to changes automatically and no translation process is active at the moment.

PR #381, opened 2021-03-21
Updated Diagram Logos from optimized png

mulyaj:
Redid the diagram logos from the current changes branch to avoid having to use git :-)


gilgongo:
OK so merge to changes - think that works...!

PR #387, opened 2021-03-24
Add the info to try to avoid using an USB-hub

### LINKED ISSUE #384: USB-Hub causing some trouble with ASIO-driver only in jamulus
**URL**
https://jamulus.io/de/wiki/Client-Troubleshooting

**Describe the bug**
Not a bug but a highly recommended info should be added. I run in to an unspecified error telling me during the "connect to server" that I have "a problem with my soundcard".
I use a ZOOM H4pro connected via USB over a USB-Hub to my PC Win8 (Generic ASIO-driver) and it works great so far. But Jamulus refused work in clientmode. 
After hours of search on my computer I found out, the USB-Hub I used caused the problem for jamulus. 
Connecting my ZOOM H4 to an USB3-Port directly on my PC solved my problem.
So I would recommend to give the "Troubleshoot-advice" to connect any USB-Mic/Dvice directly to the computer and NOT use any USB-hub in between at first-aid try.

Feel free to delete or move this input/issue if needed.
And sorry, but I couldn't find a better place to get this info to you.

tackin:
Have had that exact problem with a HAMA USB2.0-HUB. The stupid thing is, that connection-error is only seen with jamulus. Other software didn't complain (even Cubase worked for me), so I did not expect that to be the cause for that jamulus errormessage.

### DIFF
--- a/wiki/en/en-Client-Troubleshooting.md
+++ b/wiki/en/en-Client-Troubleshooting.md
@@ -7,8 +7,8 @@ permalink: "/wiki/Client-Troubleshooting"
 
 # Troubleshooting
 
-### Don’t hear any sound/Others don’t hear you?
-Start with the simple stuff: make sure your instrument/microphone and headphones are connected to the correct sockets. Make sure no other applications like your browser, video conferencing tool etc. is also using your sound card. You should shut those down when using Jamulus. If that all looks OK and the problem persists, it's likely there's an issue with your sound device settings. This will depend on your particular setup (platform, hardware, software and drivers), so it's best to ask on [the forums](https://github.com/jamulussoftware/jamulus/discussions) for guidance.
+### Don’t hear any sound/Others don’t hear you or you get an error-message while connecting to a server?
+Start with the simple stuff: if you are using an USB-microphone, connect it directly to an USB-port in your computer and avoid using an USB-hub. Even newer USB-hubs have caused the error message "Your sound card is not working correctly. Please open the settings dialog and check the device selection and the driver settings.". Make sure your instrument/microphone and headphones are connected to the correct sockets. Make sure no other applications like your browser, video conferencing tool etc. is also using your sound card. You should shut those down when using Jamulus. If that all looks OK and the problem persists, it's likely there's an issue with your sound device settings. This will depend on your particular setup (platform, hardware, software and drivers), so it's best to ask on [the forums](https://github.com/jamulussoftware/jamulus/discussions) for guidance.
 
 **Windows users (ASIO4ALL)**: If you’re using the ASIO4ALL driver have a look at the [ASIO4LL setup section](Installation-for-Windows#setting-up-asio4all)
 
--- a/wiki/en/en-Client-Troubleshooting.md
+++ b/wiki/en/en-Client-Troubleshooting.md
@@ -15,7 +15,7 @@ Start with the simple stuff: make sure your instrument/microphone and headphones
 
 ### You permanently get an error-message while trying to connect to any server?
 
-If you are using an USB-microphone, try to connect it directly to an USB-port in your computer, bypassing any USB-hub. Some USB-hubs have caused the error message "Your sound card is not working correctly. Please open the settings dialog and check the device selection and the driver settings.".
+If you are using a USB-microphone, try to connect it directly to an USB-port in your computer, bypassing any USB-hub. Some USB-hubs have caused the error message "Your sound card is not working correctly. Please open the settings dialog and check the device selection and the driver settings.".
 
 ### You all sound OK, but it's difficult to keep together
 
--- a/wiki/en/en-Client-Troubleshooting.md
+++ b/wiki/en/en-Client-Troubleshooting.md
@@ -15,7 +15,8 @@ Start with the simple stuff: make sure your instrument/microphone and headphones
 
 ### You permanently get an error message while trying to connect to any server?
 
-If you are using a USB-microphone, try to connect it directly to an USB-port in your computer, bypassing any USB-hub. Some USB-hubs have caused the error message "Your sound card is not working correctly. Please open the settings dialog and check the device selection and the driver settings.".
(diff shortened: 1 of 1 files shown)


ann0see:
Fixes #384

ann0see:
Ok. Maybe you can even remove the error message here, since probably many people would get confused. I'd say that avoid using USB-Hubs is enough. 

Background info: At least that's what I've learnt from my work with the docs: try to keep as brief as possible while bringing the maximum of information to the average user.

gene96817:
Some people may have to use a USB hub. The trend is for notebooks to have fewer USB ports. Tablets are the logical endpoint with only one port (which also provides charging). If these devices are used with Jamulus, a USB hub cannot be avoided.

ann0see (changes requested):
Could you please add a new headline with the USB hub info? I don’t think it‘s related to the "don’t hear sound" issue.

tackin:
@gene96817  
> Some people may have to use a USB hub. The trend is for notebooks to have fewer USB ports. Tablets are the logical endpoint with only one port (which also provides charging). If these devices are used with Jamulus, a USB hub cannot be avoided.

Yes you are right, but this webpage is only about troubleshooting. If a quick test by bypassing the hub (for your troubleshooting) is a success, you're done with finding your problem and only need to buy a newer/better USB-hub.

@ann0see 
It was exactly this. I got the error message and could not hear somthing. But according to the config page I was connected to the server and got ping and latency-times. So yes I guess it's the right place for that hint in my eyes.

ann0see:
Yes. But since you got an error message, it should be included somewhere else. Most people I onboarded didn't have issues because of using USB Hubs. In most cases it was a wrongly set input or output device (especially with ASIO4ALL) and they didn't even get an error message since in principle their setup was correct.

tackin:
Ok, this makes sense.

wiki/en/en-Client-Troubleshooting.md
```diff
@@ -7,8 +7,8 @@ permalink: "/wiki/Client-Troubleshooting"
...
+### Don’t hear any sound/Others don’t hear you or you get an error-message while connecting to a server?
+Start with the simple stuff: if you are using an USB-microphone, connect it directly to an USB-port in your computer and avoid using an USB-hub. Even newer USB-hubs have caused the error message "Your sound card is not working correctly. Please open the settings dialog and check the device selection and the driver settings.". Make sure your instrument/microphone and headphones are connected to the correct sockets. Make sure no other applications like your browser, video conferencing tool etc. is also using your sound card. You should shut those down when using Jamulus. If that all looks OK and the problem persists, it's likely there's an issue with your sound device settings. This will depend on your particular setup (platform, hardware, software and drivers), so it's best to ask on [the forums](https://github.com/jamulussoftware/jamulus/discussions) for guidance.
```
gilgongo: When saying "if you are using an USB-microphone, connect it directly to an USB-port in your computer and avoid using an USB-hub." then the second sentence explaining why is unnecessary because the user can't act on that information. We don't explain why, for example, a video conferencing tool should be shut down.

In fact isn't "avoid using an USB-hub" also redundant if the user obeys "connect it directly to an USB-port in your computer"?

tackin: Is updated.

ignotus666: It should be **a** USB, not **an**. Although it is a vowel, the 'U' is pronounced like a 'Y', which is a consonant.

Also, no hyphen in 'error message' or 'USB port/hub'. They're not compound nouns.

tackin: Please refer to my latest PR-update.

gilgongo: I'm still not sure what the user is supposed to do with the information about the error message. Forgive me, but it just seems like noise.

tackin: That "noice" would have saved me hours of troubleshooting with an somehow unhelpfull error message. The point is, ONLY jamulus seems to complain. Having a working setup in Cubase, Zoom, BigBlueButton and many others but getting error messages in jamulus complaining about "Soundcard not working correctly" does obviously not point to that kind of problem. Isn't a "troubleshoot page" just made to give hints what to try out to solve a problem?
What I would have done with this input on the page? I would have tried it and in seconds have the result: my USB-hub just sucks.

ann0see (approved):
For me this looks better.
Maybe we could even add some information here: 

1. Check if your interface or microphone is properly connected to your PC (try another USB Port)
2. If you're using a USB hub (...)

wiki/en/en-Client-Troubleshooting.md
ann0see: Agree. That's why I also think it should be on this page ;-). Nevertheless this message is about a generic error. See my other comment

ann0see:
BTW: @gilgongo I'm not sure, but probably this will not go live yet because of the translation process?

ann0see (approved):

wiki/en/en-Client-Troubleshooting.md
gilgongo: OK but if we have to have this, it feels to me like it's a separate section in the doc. It's not _"Don’t hear any sound/Others don’t hear you?"_ it's _"Getting the message 'Your sound card is not working correctly' and don't know why?"_ isn't it?

We need to fix the bug that gives rise to this too, so having it as a separate item makes it easier to track (and we can add other things that relate to seeing that message as well I suppose).

ann0see: Yes. See the current page. I think this was changed: https://github.com/jamulussoftware/jamuluswebsite/pull/387/commits/220e012c45778e7c5330cc5a6e221f792f3a0365

gilgongo:
We have agreed not to translate the troubleshooting page for now. But if we do want to translate it, then we need to ask the translators if they can ensure it stays up to to date. I suppose it's not a high-traffic page, but it's yet to exist in Italian (and I'm not sure how up to date it is in the other languages).

PR #396, opened 2021-03-26
New download page and new navbar 

### LINKED ISSUE #322: Create download page

### LINKED ISSUE #349: New Header (Top navbar)
This is semi-related to #322. I think we should have a new consistent header throughout with links to important pages like download, setup, and docs.

I have this design at the moment: 
image

If we were to keep these links - setup (or installation, as it is now) would either need to link to the OS-specific page, or link to a general setup page that has links to the OS-specific setup. 

For the documentation link, I'm not entirely sure where this would link - possibly jamulus.io/wiki? However, that page is fairly unorganized at the moment.

Would like to hear your thoughts.

### LINKED ISSUE #416: Merge "Getting Started" with "Onboarding" for new "Setup" page
As per discussions on: https://github.com/jamulussoftware/jamuluswebsite/pull/396

The current "Getting Started" and "Onboarding" pages overlap conceptually, have overlapping/duplicated information between them and the various platform-specific installation pages (which may give rise to documentation bugs) and have other problems. We need to link to just one "Setup" page from the nav, so creating this new page out of the two will help the overall IA of the site.

mulyaj:
# Does this need translation

- [x] Yes 
- [ ] No 

# Changes

Created a new download page and navbar.

Main homepage header is now replaced with a consistent header throughout. 

Related to #322 and #349


_includes/header.html
```diff
@@ -1,8 +1,35 @@
 <header class="fx-row fx-row-between-xs fx-text-center">
-  <div class="fx-col fx-text-center"><a href="/{% if site.active_lang != site.default_lang %}{{ site.active_lang }}/{% endif %}">
-      <img alt="{{ site.data.general.nav.altJamulusIcon }}" width="40" height="40" src="{{ 'assets/img/jamulus-icon-2020.svg'| relative_url }}" /></a>
+  <div class="fx-col fx-text-center icon">
+    <a href="/{% if site.active_lang != site.default_lang %}{{ site.active_lang }}/{% endif %}">
```
ann0see, line 3: Not 100% sure here. We'll need to test if this is ok.

_includes/header.html
```diff
@@ -1,8 +1,35 @@
...
+        <img alt="{{ site.data.general.nav.altJamulusIcon }}" width="40" height="40" src="{{ 'assets/img/jamulus-icon-2020.svg'| relative_url }}" />
+        <span>
+          <div>Jamulus</div>
```
ann0see: Why don't we use a heading or similar here?

_includes/header.html
```diff
@@ -1,8 +1,35 @@
...
+    <a href="/wiki/Software-Manual">{{ site.data.general.header.documentationLink }}</a>
+    <a class="download-header" href="/Download">{{ site.data.general.header.downloadLink }}</a>
   </div>
   <div class="fx-col-center-xs fx-txt-right hide-on-dt">
     <a href="#mnv" onclick="showNav();return false;">{% include general/navbtn.html %}</a>
   </div>
 </header>
+
+<!-- highlights current page in navbar -->
+<script>
```
ann0see: Can this be done with CSS only?

assets/css/downloads.css
```diff
@@ -0,0 +1,185 @@
+main {
+    margin: auto;
+    flex-wrap: nowrap;
+}
+
+.container {
```
ann0see, line 6: Can't we do that with the framework somehow?

assets/css/downloads.css
```diff
@@ -0,0 +1,185 @@
...
+    position: relative;
+    top: 1px;
+}
+
+
+.caption-text {
+    font-size: 14px;
+}
+
+
```
ann0see: Spacing.

ann0see (commented):
Just quickly went over it. Will try it locally ;-)

_includes/header.html
mulyaj: Not entirely sure, you might be able to, but it'd probably be much more cumbersome. Is there a better place to add scripts?

_includes/header.html
mulyaj: Will fix

gilgongo: BTW we need to modify the language selector so that we can show it selectively on pages. Right now, it should only be on homepage, getting started and Windows/Mac/Linux install pages for EN, ES, DE, and only show on all other pages for ES.  As per this thread: https://github.com/jamulussoftware/jamulus/discussions/1302#discussioncomment-526668

_includes/header.html
ann0see: Ok. Seems to work.

_data/en/general.yml
```diff
@@ -1,3 +1,9 @@
+header:
+  tagline:             "Play music online. With friends. For free."
+  setupLink:           "Setup"
+  documentationLink:   "Documentation"
```
ann0see: Didn't follow the thread. But do we really want to link the manual here?

Also how can we ensure that this also gets shown on mobile devices?

If we design mobile first, we can't hide important links on mobile.

_includes/header.html
ann0see: Not sure if that's practical. I'd rather add a message saying that the page you're at might be outdated and you should have a look at the english version instead.

_includes/header.html
```diff
@@ -1,8 +1,35 @@
...
-      <img alt="{{ site.data.general.nav.altJamulusIcon }}" width="40" height="40" src="{{ 'assets/img/jamulus-icon-2020.svg'| relative_url }}" /></a>
+  <div class="fx-col fx-text-center icon">
+    <a href="/{% if site.active_lang != site.default_lang %}{{ site.active_lang }}/{% endif %}">
+        <img alt="{{ site.data.general.nav.altJamulusIcon }}" width="40" height="40" src="{{ 'assets/img/jamulus-icon-2020.svg'| relative_url }}" />
+        <span>
+          <div>Jamulus</div>
+          <div class="tagline">{{ site.data.general.header.tagline }}</div>
+        </span>
+    </a> 
+      </div>
```
ann0see: Could you please check if the HTML is valid? It looks a bit suspicious to have this div indented here.

_includes/header.html
```diff
@@ -1,8 +1,35 @@
...
+        
+  </div>
+  <div class="fx-col-center-xs fx-txt-right headerlinks">
+    <a href="/wiki/Getting-Started">{{ site.data.general.header.setupLink }}</a>
+    <a href="/wiki/Software-Manual">{{ site.data.general.header.documentationLink }}</a>
+    <a class="download-header" href="/Download">{{ site.data.general.header.downloadLink }}</a>
```
ann0see: This link seems to be broken because it should not be capitalized

_includes/header.html
ann0see: Yes. in assets/js/

Nevertheless: why don't you add a class via jekyll logic? See the navigation bar which includes an active class for pages you're on.

download.html
```diff
@@ -0,0 +1,71 @@
+---
+layout: download
+title: "Download - Jamulus"
+lang: "en"
+permalink: /download
+
+---
+
+
```
ann0see: Any reasons for this spacing here?

Also: would it be possible to use markdown here?

ann0see (changes requested):
In general, I'd say it looks great! Have a look at my comments.

_includes/header.html
```diff
@@ -1,8 +1,35 @@
 <header class="fx-row fx-row-between-xs fx-text-center">
-  <div class="fx-col fx-text-center"><a href="/{% if site.active_lang != site.default_lang %}{{ site.active_lang }}/{% endif %}">
-      <img alt="{{ site.data.general.nav.altJamulusIcon }}" width="40" height="40" src="{{ 'assets/img/jamulus-icon-2020.svg'| relative_url }}" /></a>
+  <div class="fx-col fx-text-center icon">
+    <a href="/{% if site.active_lang != site.default_lang %}{{ site.active_lang }}/{% endif %}">
+        <img alt="{{ site.data.general.nav.altJamulusIcon }}" width="40" height="40" src="{{ 'assets/img/jamulus-icon-2020.svg'| relative_url }}" />
+        <span>
```
ann0see: I'd not use span here. I think having a box element in an inline element is not valid html code. Please use <div>

ann0see (commented):
Also the language navigator on the downloads page seems to be broken.

_layouts/download.html
```diff
@@ -0,0 +1,25 @@
...
+  <title>{{ page.title }}</title>
+  {% include headtags.html %}
+  <link rel="stylesheet" href="/assets/css/downloads.css">
+</head>
+
+<body>
+  {% include header.html %}
+  <div>
+    <div>
+    {% include general/nav.html %}
```
ann0see, line 14: From my comment above: you should include {% include langselect.html %} instead.

download.html
mulyaj: You mean for the entire page instead of HTML?

_includes/header.html
mulyaj: div instead?

_includes/header.html
mulyaj: I don't see why it wouldn't be valid - I think for markdown indents will break the code.

Anyways, I'll fix the indents to make it easier to read

_includes/header.html
mulyaj: I'll give this a go, I think that should work

_includes/header.html
gilgongo: If readers can't know what's in an what's out of date in their translation, and if we can't remove the links, it would seem better to show a page saying "Sorry, well have paused translation of this page" or something. Even then that's fairly annoying to click on a link only to be told that. 

Perhaps better therefore to simply hard-code the links to pages that we're providing translations for?

download.html
ann0see: > You mean for the entire page instead of HTML?

Use as much Markdown as possible since it will make reading much easier for e.g. translators.

You can add html to .md files. See: https://stackoverflow.com/a/35614879

_includes/header.html
ann0see: Yes.

_includes/header.html
ann0see: Great!

_includes/header.html
ann0see: See: https://github.com/jamulussoftware/jamuluswebsite/blob/3431f8315324ee0edc094447ce8b8cd38f26c4a8/_includes/general/nav.html#L9

_includes/header.html
ann0see: Hard coding would probably result in a lot more development work and it would break the whole translation. So adding a message or removing the pages which are outdated would be much cleaner.

_data/en/general.yml
mulyaj: The manual link is sort-of a temporary link. @gilgongo and I discussed that we should probably make a documentation hub page which has the full list of wiki pages and kb articles etc. 

About the header, that's a good question - I have to think about that. We'd probably have to change the nav dropdown. If we do go ahead with the docs hub, then we could just have the three links on the dropdown. But then we'd also have to ask how the list of docs will be shown in the hub.

ann0see: Ok. Sort of unrelated, but https://youtu.be/bQ-JwnGM2-c?t=163 might give some insight on our website.

download.html
mulyaj: For a mostly text-based page I think .md would be fine, but I found the HTML formatting confusing when there's a lot on the page.

ann0see: Fair point.

_includes/header.html
gilgongo: I don't want to bother the translators to add messages all over the place. I'll remove the affected stuff in that case.

_includes/header.html
mulyaj: Great, I got it working, thanks!

_layouts/download.html
mulyaj: Sorry not sure I understand. Isn't langselect already included in nav.html?

ann0see: Yes it is. Unfortunately the layout of the download page seems to be broken (the language selector is somewhere in the middle of the page, not e.g. on the top right)

mulyaj: It seems to work locally for me.

It looks like this on my screen:

image

ann0see:
Just had another thought:

Did we remove the get started now button with this PR? If that's the case, I'd suggest to re-add it since quite a lot of videos on various pages hint at it. Also, I think we shouldn't break the current download path if possible

mulyaj:
@ann0see Yes, the Get Started button is removed. 

If people are just using the "Get Started Now" button as a path to download, I think having a straight-forward download link is more efficient. This could probably be verified with some usability tests.

ann0see:
> I think having a straight-forward download link is more efficient.

It could also lead to frustration for people who download and don't read. They might not have an ASIO driver (--> they'd be greater with an error message), use WiFi, etc.

download.html
```diff
@@ -0,0 +1,71 @@
...
+    <div class="container">
+        <div class="fx-col-100-xs fx-col-40-m fx-center infoarea">
+            <h3>Got a pair of headphones and an instrument/mic?<br>If so, you're basically all set!</h3>
+            <div class="requirements">
+                <p>Minimum System Requirements</p>
+                <ul class="list">
+                    <li class="listitem">Stable Internet Connection</li>
```
ann0see: Maybe add (WiFi isn't stable. Use Ethernet)

download.html
```diff
@@ -0,0 +1,71 @@
...
+                    <li class="listitem">1.5 Ghz Processor or better</li>
+                    <li class="listitem">90 mb of available storage</li>
+                    <li class="listitem">512 mb RAM or faster</li>
+                </ul>
```
ann0see, line 18: For windows you currently also need an ASIO Driver (like ASIO4ALL)

gilgongo:
> It could also lead to frustration for people who download and don't read.

If they don't read, then I don't think we can make them :-)

At first, the docs were constructed to heavily emphasises the initial requirements and setup before you got to the download. In the discussion on https://github.com/jamulussoftware/jamuluswebsite/issues/322 it was felt that there was a user expectation to have a straight download link, since that's the main convention on other sites.

ann0see:
Hmm. I hope that a direct download link is ok. Nevertheless, see my suggestions from above.

download.html
ann0see:
```suggestion
                    <li class="listitem">Stable Internet Connection (WiFi isn't stable; Use Ethernet)</li>
```

download.html
```diff
@@ -0,0 +1,71 @@
...
+    <div class="container">
+        <div class="fx-col-100-xs fx-col-40-m fx-center infoarea">
+            <h3>Got a pair of headphones and an instrument/mic?<br>If so, you're basically all set!</h3>
+            <div class="requirements">
+                <p>Minimum System Requirements</p>
+                <ul class="list">
+                    <li class="listitem">Stable Internet Connection</li>
+                    <li class="listitem">1.5 Ghz Processor or better</li>
+                    <li class="listitem">90 mb of available storage</li>
+                    <li class="listitem">512 mb RAM or faster</li>
```
ann0see:
```suggestion
                    <li class="listitem">512 mb RAM or faster</li>
                    <li class="listitem">Windows only: an ASIO Driver like [ASIO4ALL](Installation-for-Windows#asio)</li>
```

download.html
```diff
@@ -0,0 +1,71 @@
...
+        <div class="fx-col fx-center fx-txt-center helpbox">
+            <div>
+                <h4>Learn to use Jamulus!</h4>
+                <p>Check out our onboarding guide to learn how to get your first jam sounding great.</p>
+            </div>
+            <div>
+                <a href="wiki/Onboarding"><img width="48" height="48" src="{{ 'assets/img/blue-book.svg' | relative_url  }}" alt="Book icon, link to onboarding page">
+                    <div>Onboarding</div>
+                </a>
+
```
ann0see:
```suggestion
```

_data/en/general.yml
gilgongo: Wasn't he the guy who initially reviewed Jamulus (and some others?) and concluded that it didn't work?

download.html
mulyaj: How about: "Stable Internet Connection (Ethernet recommended)

mulyaj: I think we should leave min requirements as brief as possible

_data/en/general.yml
ann0see: I think so? It would be great to get into contact with him...

download.html
ann0see: If that's clear enough, sure.

_includes/header.html
pljones: `<span>` is an inline element.  https://www.w3schools.com/tags/tag_span.asp (and elsewhere if you don't like W3Schools).

mulyaj:
Ok, I've made the changes.

_layouts/download.html
mulyaj: I didn't add langselect, as I made some changes to the css instead.

download.html
gilgongo: Shouldn't it be "512 mb RAM or more"?

ann0see: Yep

ann0see:
Still worry a little bit about the change on the homepage. Couldn't we add a visible getting started and download button at a similar space the getting started button was?

gilgongo:
The fact that the main nav will have those is enough I think (there is an extremely strong convention of people looking to the nav bar). But if we want to have panels for these on the home page as well, then that's OK. It just means a re-design :-)

BTW can we mark some of the above comments as resolved? It's getting a bit hard to know what's outstanding and what isn't.

download.html
mulyaj: will fix

ann0see:
Ok. That's a picture in dark mode. I'm still wondering if the content (especially troubleshooting etc.) isn't a bit off? Also the color of the headline (minimum system requirements) seems strange to me: 

Download - Jamulus

mulyaj:
Yeah, I'll that headline color a change. The troubleshooting box should be aligned properly, I used a centering flexbox there. 

However, off-topic, if the text in the boxes has unequal lines the images become misaligned.

ann0see:
Can you use the FOX css grid for alignment not your own css? This will probably allow a more consistent layout. Also: what about the navigation? Do we really not want it?

Another thought: what about adding a button-secondary and button-primary class so that we can use your button style and the usual button style on other pages too?

mulyaj:
I'll see what I can do with fox-css. I'll also add the button styling as well.

For the nav, I agree with @gilgongo, I think the navbar is fine.

pljones:
This could do with a rebase and resolution of the conflicts, to keep it fresh.

_includes/header.html
```diff
@@ -1,8 +1,28 @@
...
+  </div>
+        
+  </div>
+  <div class="fx-col-center-xs fx-txt-right headerlinks">
+    {% for item in site.data.headerlinks.header-links %}
+      <a 
+        href="{{ item.url }}" 
+        class="{% if item.url == "/download" %}download-header{% endif %}
+              {% if item.url == page.url %}current{% endif %}"
+              >
```
ann0see, line 19: This way of styling and adding parameters looks a bit odd to me. Are you sure this is correct?

_includes/header.html
```diff
@@ -1,8 +1,28 @@
...
+        {{item.title}}</a>
```
ann0see:
```suggestion
        {{ item.title }}</a>
```

download.html
```diff
@@ -0,0 +1,65 @@
+---
+layout: download
+title: "Download - Jamulus"
+lang: "en"
+permalink: /download
+---
+    <div class="fx-row fx-row-between-xs container">
```
ann0see, line 7: I'm still not 100% convinced if we really want to maintain one page for every language with that much HTML. Yes, the index.html page looks quite similar.

ann0see (commented):
Just reviewed it again.
Before we merge this, maybe we should do some tests and have a thought on some questions:

Does this lead to better user experience? Does the user get enough information to set up Jamulus properly? Do we have duplicates?

Technical view:

Do we have valid HTML: https://validator.w3.org/

Do we use the framework when ever possible?

Any issues in dark mode? 

Does the page work well on mobile?

What doesn't work if JS is disabled (this should be close to zero)?

_includes/header.html
mulyaj: I'm open for any suggestions on styling.

gilgongo:
@ann0see 

> Does this lead to better user experience? 

As discussed, but I think the main points are:

1. The convention on other sites is to have a main nav bar, usually at or near the top of the page that "describes" the main things we offer: getting the software (download), setting it up, and getting to the details of how to use it (documentation). These may well be user "key terms" that they are scanning for, and we didn't have those things clearly laid out before.

1. There is also a convention to make downloading the main (or a major) call to action, and there is a view that people expect this. In fact until we moved away from SF, that's effectively what we had, so in that sense there is no change, but it's a better UX than SF for getting the software (which of course had completely different navigation and maybe a bunch of tracking links). We did at first take another view: that users should read something about setup before they downloaded it, but it was felt people just want to get their hands on Jamulus first and read about it later if they want.

1. It allows us to have a lot more navigation items on the "documentation" page, since that's expected from such a section. Putting lots of nav on the other pages make us look a bit too complicated.

> Does the user get enough information to set up Jamulus properly? 

This is a largely unanswerable question, but given the thinking above and the fact that at some point we will be handling first-time setup in the application anyway, we are keeping with convention and offer an easy download rather than making people read manuals prior to use.

> Do we have duplicates?

Not sure what you mean.

download.html
gilgongo: With luck it'll hardly change once it's settled down.

mulyaj:
> What doesn't work if JS is disabled (this should be close to zero)?

I'll need to style the no javascript message.

_includes/header.html
ann0see: I'd just put the whole a element in one line.

ann0see:
@gilgongo These are all valid arguments. We'll probably do a bunch of restructuring anyway (merge of getting started and onboarding,...), so probably it is ok anyway. 

By duplicates I mean having the "You will need" section on the Getting Started hub and on the new download page. If that's no problem, we can of course maintain it at multiple places.

gilgongo:
> Getting Started hub and on the new download page.

BTW the "setup"  link in the header here links to Getting Started temporarily while the new Setup page has yet to be merged. But there is a bit of "you will need" overlap between the two which we may need to be careful doesn't drift apart over time perhaps.

mulyaj:
Okay, I've styled the noscript message. Here's how it looks.

image

mulyaj:
Another thing to add: Should an audio device with a 48k Hz sample rate be also a minimum requirement?

gilgongo:
I think it depends on the chances of them not having a mic that doesn't support 48kHz. If it's most likely they will have one, then perhaps not bother people with that worry unless they encounter a problem (and I suppose that depends on whether the error they get is explanatory). So I'd be inclined not to say anything at this stage as it's not an easy thing to check and again sort of messages that Jamulus is complicated/hard work. BTW the same sort of argument could be applied to the rest of the minimum specs - eg somebody with less than 512Mb of RAM is going to be encountering issues with many other things well before they discover Jamulus I would think.

pljones:
I'd actually cut it down to three:
* Stable internet connection (Ethernet blah as now)
* Suitable audio input and output
* ASIO driver (Windows only)

The others are now what I'd say can safely be assumed in most cases without being stated.  (At least for Windows, MacOS and most Linux systems.)  I'd possibly add a link to "further details" where the 48KHz audio and CPU/Storage/RAM requirements are given.  Not sure where that should live, though.

gilgongo:
OK so how about what we have at the beginning of Getting Started? I have in fact preserved that on the new [Setup page](https://github.com/jamulussoftware/jamuluswebsite/pull/418) too:

To get the best from Jamulus, at a minimum you will need:

1. **A wired internet connection** (not wi-fi - see the [FAQ](/wiki/FAQ))
1. **Wired headphones** (not Bluetooth or speakers - see the [FAQ](/wiki/FAQ))
1. **A reasonable audio device, sound card and/or mic** ([see this list](/kb/2021/01/05/Jamulus-Sound-Devices.html) for examples)

(dunno if it's sensible to do it with an include so we can maintain that text in one place?)

jujudusud:
> 2. **Wired headphones** (not Bluetooth or speakers - see the [FAQ](/wiki/FAQ))

This is only valid for people who use a microphone. I play electric guitar, directly plugged into a usb audio interface, what's to stop me from using monitoring speakers?

gilgongo:
> what's to stop me from using monitoring speakers?

Nothing really, it's just that at his point all we want to do here is give the information that will work for **most** people (with the least risk of bad consequences). There is after all nothing to stop a violinist using a speaker if they are OK with playing ahead of the beat, or somebody using RF headphones (as I do) and not Bluetooth... But here is not the place to explain all that. 

BTW I have wondered if one day we might introduce an interactive version of this where you select what setup you have and we tell you what you need to do based on the potential combination of instrument type, platform/ASIO driver, internal/external audio interface and have mic or not. But that's for the future :-)

pljones:
> 
> 
> > 1. **Wired headphones** (not Bluetooth or speakers - see the [FAQ](/wiki/FAQ))
> 
> This is only valid for people who use a microphone. I play electric guitar, directly plugged into a usb audio interface, what's to stop me from using monitoring speakers?

No, it really helps with anyone playing with a sound they can hear directly.  It stops you hearing the direct sound rather than the returned sound.  That's the main point and applies to everyone.  Preventing feedback is important, too, of course, when using an open mic, whether for vocals or micing an amp or drum kit.  Even MIDI drums make a noise when you hit them, so hearing the returned sound rather than the dull thud without the expected MIDI triggered sound helps.

As @gilgongo says, the exception is that it might be okay not to, so the rule is that you should.

mulyaj:
So are we using minimum requirements or recommended requirements?

I prefer @pljones' recommendation.

gilgongo:
Yes, his suggestions are usually the best I think :-)

jujudusud:
> No, it really helps with anyone playing with a sound they can hear directly. It stops you hearing the direct sound rather than the returned sound. That's the main point and applies to everyone. Preventing feedback is important, too, of course, when using an open mic, whether for vocals or micing an amp or drum kit. Even MIDI drums make a noise when you hit them, so hearing the returned sound rather than the dull thud without the expected MIDI triggered sound helps.
> 
> As @gilgongo says, the exception is that it might be okay not to, so the rule is that you should.

Ok, that's exactly what I said. I stop trying to explain.

ann0see:
@mulyaj just had a look at the download page from a graphical POV again. 
grafik

Although I'm not a designer, shouldn't all the boxes (Got a pair of headphones and an instrument/mic?
If so, you're basically all set! with Download Jamulus) be in one line with the three boxes below? I think we can discuss this in our talk this week since I think I brought this up before. In the meantime, you could have a look at the grid layout of fox-css.You shouldn't need to use display:flex; manually.

mulyaj:
> If so, you're basically all set! with Download Jamulus) be in one line with the three boxes below? I think we can discuss this in our talk this week since I think I brought this up before. In the meantime, you could have a look at the grid layout of fox-css.You shouldn't need to use display:flex; manually.

Oh, the vertical alignment there is intentional. Yeah, we can discuss this in our chat this week.

ann0see:
Closing as discussed. 

Further to do here:

- [ ] Create download page based on wiki layout (and this draft)
- [ ] Open a separate PR with the navigation

PR #412, opened 2021-04-02
Add Presonus studio 24C audio interface to known good devices, add so…

### LINKED ISSUE #372: Presonus studio 24c
To add to the list of known good devices:

- Focusrite scarlett tested to works very well on several different macs (so 'should' can be removed)
- An older focusrite FireWire clarett works well on Mac

Presonus studio 24c works:
- on Windows with the provided ASIO driver, provides the absolute best USB audio interface round trip latency we have seen, tested to have 4.1ms at 48khz, 64 samples
- works very well on mac, we have not measured rtl

I have some more audio interface RTL measurements, measured with oblique RTL utility or with Jack delay, all at 48khz 64 samples:
-Behringer umc202hd: 7ms windows, 10.5ms Linux and mac
- focusrite Scarlett: around 7ms windows and Mac OS. 
- hifiberry dac+ adc pro on raspberry pi 4: 4.9-5.1ms stable, 3.1-3.3ms possible with realtime kernel, but unstable during UI interaction.

But since there are no RTL measurements on that page, perhaps not so interesting.

pieterbos:
…me RTL measurements and change one 'should work' to 'works

Fixes #372

# Does this need translation?

- [x] Yes 
- [ ] No 

# Changes

- Added Presonus Studio 24c as known good hardware for Mac and Windows, with RTL measurement in windows
- Changed Focusrite Scarlett on Mac OS X to works great, plus latency measurement
- Added latency measurement to Behringer UMC202HD on mac OS X
- Added latency measurement plus more working kernel versions on HifiBerry Dac+ ADC (Pro)

This would benefit from translation, but the original page is not translated, so not sure to what branch this should go.
Maybe the latency measurements should be in a separate table, to keep things more readable?

### DIFF
--- a/_posts/2021-01-05-Jamulus-Sound-Devices.md
+++ b/_posts/2021-01-05-Jamulus-Sound-Devices.md
@@ -105,7 +105,7 @@ _**Note:** The Ammoon AGM04 appears to be a re-branded [ART USBMix4](https://art
 
 **Windows**: Works great. [ASIO driver Windows 7 to 10](https://mediadl.musictribe.com/download/software/behringer/UMC/UMC-Driver_4-59-0.zip)
 
-**macOS**: Works great
+**macOS**: Works great (measured interface latency with Oblique Audio RTL Utility, 48000 64 samples and cable from output to input on Mac OS X Big Sur: `10.625ms`. Note this is **only the interface latency** not the latency in Jamulus)
 
 **Linux**: Works great (measured interface latency with `alsa_delay hw:1 hw:1 48000 64 2` and cable from output to microphone input on Debian 10: `8.451 ms`. Note this is **only the interface latency** not the latency in Jamulus)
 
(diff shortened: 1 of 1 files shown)


gilgongo:
KB articles don't require translation and are not part of the translation process for the rest of the site. But if anyone wants to do them that's fine. Ignore otherwise.

EDIT: Wait - this should be going out on the release branch for this reason, no?

gilgongo:
Maybe we should add a line to @hoffie release process stuff to clarify how the KB works? The original idea was to have a place on the website that would replace the old wiki, where anyone could just whack something up without having to concern the rest of the project work. Having these PRs handled as part of all the other stuff we track isn't the idea.

ann0see (changes requested):
Looks good. It should go to release. If you want to translate it, feel free to do so. But you'll need to be quick and translate every edit to not be out of sync of the English translation

gilgongo:
I think for KB stuff, we just publish it without the need to for any review. Part of the point of the KB is that it's for peripheral information maintained by people who aren't the main team. They don't have to wait for translation or anything.

PR #418, opened 2021-04-05
Merge "Getting Started" with "Onboarding"

### LINKED ISSUE #416: Merge "Getting Started" with "Onboarding" for new "Setup" page

gilgongo:
# Does this need translation?

- [x] Yes 
- [ ] No 

# Changes

Combines the current “Getting Started” and “Onboarding” pages into one “Setup” page. Also created new "FAQ" page for non-core common curiosity questions.


wiki/en/en-Setup.md
```diff
@@ -0,0 +1,69 @@
...
+permalink: "/wiki/Setup"
+---
+
+# Setup
+
+To get the best from Jamulus, at a minimum you will need:
+
+1. **A wired internet connection** (not wi-fi - see the [FAQ](/wiki/FAQ))
+1. **Wired headphones** (not Bluetooth or speakers - see the [FAQ](/wiki/FAQ))
+1. **A reasonable audio device, sound card and/or mic** ([see this list](/kb/2021/01/05/Jamulus-Sound-Devices.html) for examples)
```
ann0see, line 14: ... which is probably going to cause a lot of discussion:

Shouldn't we give a bit more information here or on the page we link. But this would then require it to be translated...

What I'd like to have somewhere:
1. Avoid internal sound cards (although you could start with it)
2. Only use devices which support 48kHz

wiki/en/en-Setup.md
```diff
@@ -0,0 +1,69 @@
...
+
+If you have any questions, please see the [Discussions](https://github.com/jamulussoftware/jamulus/discussions)
+## Getting started
+If you haven’t done so already, please follow the installation guide for your platform to get Jamulus working with your hardware:
+
+* [Windows](/wiki/Installation-for-Windows)
```
ann0see, line 21: ... which might hint at the need of another rewrite of the windows page and maybe a separate page for ASIO4ALL?

wiki/en/en.FAQ.md
```diff
@@ -0,0 +1,48 @@
...
+### Do I need to run a server?
+
+No. If you just intend to connect to other people’s servers, then all you need is a client. [Read this if you think you need to run your own server](/wiki/Running-a-Server). 
+
+### How does Jamulus work (in general)?
+
+[diagram]
+
+Jamulus works on the client-server principle. Everybody’s audio is sent to a server, mixed and processed there. Afterwards, the audio is sent back to each client. If a server is made public and registered on a central server, its information will be broadcast to all clients.
+
```
ann0see, line 47: ... does Jamulus provide Video support? 

No. That's for a good reason: Adding video support adds complexity! You can use other software like Jitsi or Zoom.

... I have trouble with my audio: I hear a loud squeal or nothing at all:

Your audio devices aren't correctly set up... see the troubleshooting page,

ann0see (commented):
That's a huge re-write ;-). Not sure if we want to think a little bit more about it soon. (keep the video I mentioned somewhere else in mind)

wiki/en/en-Setup.md
gilgongo: The primary mission of this page is to get the user connected and working. How **well** they are working is another matter and can be handled by troubleshooting, forums, software manual and all the rest of it. For a first time user, just to get the point of "MVP" is enough at this point. So - don't attempt to "optimise" them with advice on hardware purchases or will just seem like hard work. The link to the hardware list is enough, for example, if they are curious.

So what I was aiming for here was to just give instructions. Any _explanations_ for those instructions are elsewhere for those who want to read them if curious.

The 48kHz thing is arguable, but again - we're not expecting them to go dig out their manual on the off chance that it's not 48kHz. If it's not (and most are) then that's for later. Let's just make things easy.

wiki/en/en-Setup.md
gilgongo: I wrote this page so that it would fit in reasonably well with the installer pages (and removed all the overlapping/duplicated stuff). Was there anything in those pages that needs changing? If so can add that to this PR. (EDIT: I'll need to change links to point so Setup.md)

BTW do mean the ASIO4all stuff makes the Windows page seem too complicated? So if instead we said "You'll need an ASIO driver for your audio hardware. If your hardware doesn't come with it's own ASIO driver, look at this guide"? That might nice - although TBH it doesn't seem that bad. There is [this we can link to](https://jamulus.io/kb/2021/03/20/ASIO4ALL-Examples.html). In fact maybe we should.

wiki/en/en.FAQ.md
gilgongo: I'll add the video stuff, but "having trouble with X" is for Troubleshooting. This FAQ is just a page for general  "I'm curious/want to know more" stuff.

ann0see: Ok. Yes. However I think that's one of the main issues people face on their first install?

gilgongo: They won't be following a link that says something like "If you have problems using Jamulus". If they are that's a bug :-) The FAQ is a place where we can put all the bits and pieces that people ask about that aren't about using Jamulus itself. For usage problems they should instead be directed to Troubleshooting.  I'll add a link to that though at the end.

wiki/en/en-Setup.md
ann0see: > There is this we can link to. In fact maybe we should.

Yes, but that's only one part (since it gives specific examples but not a generic help). I think we might leave it as is at the moment but keep it in mind until others come up with an idea. Yes, the ASIO4ALL stuff does make the Windows page look difficult. But to be honest, a lot of windows users do benefit from a detailed ASIO4ALL set up guide.

mulyaj:
To add, I think the setup should address how one sets up their audio hardware. 

I think the overall process for setting someone up is Download -> Installation -> Setting up Hardware/Drivers -> First use. Right now the setup covers First Use, but doesn't really talk about Hardware/Drivers. I know there's _Hardware Setup_ but I think everything should stay on this one page.

ann0see:
Yep. Very valid point!

gilgongo:
@mulyaj 

> Right now the setup covers First Use, but doesn't really talk about Hardware/Drivers.

Hardware and drivers is an OS specific thing, so that's why this page links out to the installation for each platform where that is handled. Having Mac readers bother with ASIO is not what we need. And don't forget this page is where the main nav links to, it sits next to "download". The use cases here are distinct and we can't really combine them.

In any case, I don't think "Download -> Installation -> Setting up Hardware/Drivers -> First use" can work because you need some kind of "orientation" about wi-fi and headphones and things which affects installation and setting up hardware/drivers. At least, that's what we tried last time and why we ended up with this overlapping split between "getting started" and "onboarding".

mulyaj:
> In any case, I don't think "Download -> Installation -> Setting up Hardware/Drivers -> First use" can work because you need some kind of "orientation" about wi-fi and headphones and things which affects installation and setting up hardware/drivers. 

Hm. That's true, I don't think that content needs to be there then. However, in that case, could the links to installation be more emphasized? Maybe give them their own "Installation" header?

edit: For instance, making the text more imperative? 

## Installation
To get Jamulus working with your hardware, please follow the installation guide for your platform:

* [Windows](/wiki/Installation-for-Windows)
* [macOS](/wiki/Installation-for-Macintosh)
* [Linux](/wiki/Installation-for-Liknux):

## Getting Started

etc.

gilgongo:
> Maybe give them their own "Installation" header?

Yes that's better.

gilgongo:
Any more on this? Otherwise I'll be merging in preparation for 3.8.

wiki/en/en-Setup.md
```diff
@@ -0,0 +1,74 @@
...
+
+Once connected to a server, make sure you can hear yourself OK and fix any input volume or other problems. And of course, check that you are also able to obey [The Golden Rule](/wiki/Client-Troubleshooting).
+
+## Playing for the first time
+
+With your sound all set up, you are ready to go. When you connect to a server (you may want to select a genre from the list) and the sliders you see on the right are your own personal mix. Everything you change here will change what you hear, but won’t affect others. If you move a slider down, that user will be quieter, if you move it up, that user will be louder for you.
+
+[screenshot]
+
+If you don’t want others to hear your audio, click on the “Mute Myself” button which will stop your audio from being sent to other musicians. They won’t be able to tell you have done this though. And if you see a “mute” icon next to a player, that means they can’t hear you.
```
pljones: > And if you see a “mute” icon next to a player, that means they can’t hear you. 

This seems a bit out of context.  Maybe too advanced for now?

wiki/en/en.FAQ.md
```diff
@@ -0,0 +1,52 @@
...
+So it makes sense to minimise any sources of delay or other problems with the signal. Even fast wi-fi is usually too variable to use for long periods with Jamulus, and Bluetooth headphones and mics usually introduce latencies of about 50ms or more. 
+
+### Why should I not listen to my own signal?
+
+For the same reason as you need to minimise delay in your signal in order to play in time, you need to make sure you are playing to your own sound that other musicians are hearing. More information on this, and a way of testing your setup to make sure you are obeying this “Golden Rule” [can be found here](/wiki/Client-Troubleshooting).
+
+Of course, if you are playing an acoustic instrument, or are a singer, it will be hard to exclude your “local” sound. But you should at least try to do so by for example using closed back headphones turned up as loud as you can to mask your own sound so that you can concentrate on the mix coming back to you from the server.
```
pljones: > turned up as loud as you can

You'll have the Health And Safety Department on to you.  "as best you can" might be more acceptable :)

wiki/en/en.FAQ.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Do I need a fast Internet connection?
+
+For most people on standard broadband (e.g. 10 Mbit/s down and 1 Mbit/s up) you will have no problems. For those running servers at home, depending on how many people join, you may encounter issues if your upstream bandwidth is lower than about 5 Mbit/s. [More information on network requirements here](/wiki/Network-Requirements).
+
+### Do I need to run a server?
+
+No. If you just intend to connect to other people’s servers, then all you need is a client. [Read this if you think you need to run your own server](/wiki/Running-a-Server). 
+
+### How does Jamulus work (in general)?
```
pljones, line 43: Is this a frequently asked question - or just a resting place for the diagram?

wiki/en/en.FAQ.md
```diff
@@ -0,0 +1,52 @@
...
+
+[diagram]
+
+Jamulus works on the client-server principle. Everybody’s audio is sent to a server, mixed and processed there. Afterwards, the audio is sent back to each client. If a server is made public and registered on a central server, its information will be broadcast to all clients.
```
pljones: > central server

Hum....

wiki/en/en-Setup.md
gilgongo: This was out of an attempt to explain muting as it relates to "mute myself" in general. So maybe it should be "But if you see a “mute” icon next to a player..."

wiki/en/en.FAQ.md
gilgongo: "Turn it up until you hear nothing else"? "As much as you can"? "To eleven"? "Until unconsciousness or death intervene"?

wiki/en/en.FAQ.md
gilgongo: Oh, yes it's place for the diagram - gah I need to get the URL for that...

wiki/en/en-Setup.md
pljones: Ah, suddenly... Yes, "But..." makes it clear!  I'd add "because they've chosen to mute your mixer channel in their mix" (or something simpler but along those lines).  Otherwise I think it's still a bit disorienting.

wiki/en/en-Setup.md
```diff
@@ -0,0 +1,74 @@
...
+
+[screenshot]
+
+The most important thing about servers is their “ping time”. The bigger the number, the harder it will be to play in time with others. Usually, you would select a server with a ping of less than 15ms if you can. 
+
+Once connected to a server, make sure you can hear yourself OK and fix any input volume or other problems. And of course, check that you are also able to obey [The Golden Rule](/wiki/Client-Troubleshooting).
+
+## Playing for the first time
+
+With your sound all set up, you are ready to go. When you connect to a server (you may want to select a genre from the list) and the sliders you see on the right are your own personal mix. Everything you change here will change what you hear, but won’t affect others. If you move a slider down, that user will be quieter, if you move it up, that user will be louder for you.
```
mulyaj: You mentioned in the style guide you wanted to use "person", does it apply here?

pljones: > that user will be quieter, ... that user will be louder

That's wrong anyway... :)  It's the sound that's quieter or louder.  I still say it needs to be channel.  I've two channels labelled "pljones" when I'm connected.  Same person, same player, same users, whatever if you're thinking about "who it is".  It's just there are two channels - one for mic, one for instrument.  It allows the channels to be adjusted independently.  The idea of "channels" needs to be "natural" not made "something technical".

> ... for you.

Doesn't recover well enough.

gilgongo: > You mentioned in the style guide you wanted to use "person", does it apply here?

Hehe - yes, I thought I'd give that a definition and then see how or if the docs conformed to it. It may be too difficult to have a definition of "person" but in this case (and as given by my definition of "channel" being a system thing, and "fader" being a UI thing in the style guide) it would be:

"If you move a fader down, that channel will be quieter, if you move it up, the channel will be louder for you."

Thereby also addressing @pljones rage on this too.

BTW when/if mastermix comes in, this may need to be re-visited but right now AFAIKT it won't be affected.

gilgongo:
Does anyone know why, when I build this branch, the nav refuses to show the Setup page and instead replaces it with the Italian version of Getting Started? The only way I can get it to show up is to delete all the non-en wiki dirs.

wiki/en/en-Setup.md
```diff
@@ -58,7 +67,10 @@ Once connected to a server, make sure you can hear yourself OK and fix any input
 
 With your sound all set up, you are ready to go. When you connect to a server (you may want to select a genre from the list), the faders you see on the right are your own personal mix. Everything you change here will change what you hear, but won’t affect others. If you move a fader down, that channel will be quieter, if you move it up, the channel will be louder for you.
 
-[screenshot]
+<figure>
+  <img src="{{site.url}}/assets/img/en-screenshots/main-screen-medium.png" loading="lazy" alt="Screenshot" style="width:100%; float:none;">
+  <figcaption>The main window when you are connected to a sever</figcaption>
+</figure>
 
 If you don’t want others to hear your audio, click on the “Mute Myself” button which will stop your audio from being sent to other musicians. They won’t be able to tell you have done this though. But if you see a “mute” icon next to a player, that means they can’t hear you becuase they've muted you in their mix.
```
pljones: > if you see a “mute” icon next to a player, that means they can’t hear

Perhaps
```
if you see a “mute” icon next to a channel, that means that person can’t hear
```

pljones: > becuase 

Typo

gilgongo: Ah yes, "player" is not legit by our new style guide. Good spot.

pljones: > they've muted you

"your channel", I guess, too.

mulyaj:
Did you manage to fix the nav? I've tested it on my end and it seems to work fine.

gilgongo:
If I git checkout gilgongo-setup and hit the link the "Setting up" link in the nav, it comes back with Italian version of Getting-Started. Trying to go directly to Setup also redirects to Getting Started (which then 404s).  If I remove wiki/en/en-Getting-Started.md, I can go to /wiki/Setup but it 404s. Setup is also not listed in both cases at http://127.0.0.1:4000/wiki/

wiki/en/en-Setup.md
```diff
@@ -0,0 +1,86 @@
+---
+layout: wiki
+title: "Setup"
+lang: "en"
+permalink: "/wiki/Setup"
+---
+
+# Setup
```
gilgongo:
```suggestion
# Setup - getting started with Jamulus
```

wiki/en/en.FAQ.md
```diff
@@ -0,0 +1,53 @@
...
+
+### How do I know if I can join a server? Are there rules?
+
+In general, if somebody sets up a public server, then [by definition](/wiki/Choosing-a-Server-Type) they accept that anyone can play on it. Jamulus has no password protection or other authentication mechanisms. However, some servers may state their policies in the welcome message you will see in the chat window.
+
+You can also set up a private server on Jamulus and give others your address to connect to that. Have a look at [this page for more information](/wiki/Running-a-Server).
+
+### Why shouldn’t I use wireless equipment?
+
+Whether or not you will be able to play in time with other musicians depends mainly on how much latency you have in your sound signal. An overall delay much bigger than 50ms usually makes it too hard to play in time unless you train yourself to do so. Some musicians report they can play in time with delays of up to 70ms by playing ahead of the beat. But much beyond that and it becomes too difficult unless timing isn’t an important factor in the music.
```
gilgongo:
```suggestion
Whether or not you will be able to play in time with other musicians depends mainly on how much latency (delay) you have in your sound signal. An overall delay much bigger than 50ms usually makes it too hard to play in time unless you train yourself to do so. Some musicians report they can play in time with delays of up to 70ms by playing ahead of the beat. But much beyond that and it becomes too difficult unless timing isn’t an important factor in the music.
```

PR #421, opened 2021-04-09
DSCP Quality of Service Documentation

ghost:
# Does this need translation?

- [x] Yes 

- [ ] No 

# Changes

Added a new command line option (wiki/en/en-Command-Line-Options.md)
Added a new tip for Quality of Service (wiki/en/en-Tips-Tricks-More.md)

### DIFF
--- a/wiki/en/en-Command-Line-Options.md
+++ b/wiki/en/en-Command-Line-Options.md
@@ -42,6 +42,7 @@ You can see all possible options your version supports by starting Jamulus with 
 |    `-n` |`--nogui`          | Disable GUI | (client and server) |
 |    `-o` |`--serverinfo`     | Server location details in the format: <br/>`[name];[city];[locale value]` (see [values](https://doc.qt.io/qt-5/qlocale.html#Country-enum))| (public servers only) |
 |    `-p` |`--port`           | Local UDP port number. Default is 22124 | (server only) |
+|    `-Q` |`--qos`            | Request quality of service in the IP header octet.<br> (e.g. --qos 0)&nbsp;&nbsp;&nbsp;DF=0, AF31=104, CS4=128 default, EF=184 | see [Tips & Tricks](Tips-Tricks-More#quality-of-service) (global option)
 |    `-R` |`--recording`      | Include a writeable path where the files should be stored (in quotes if needed). | (server only) See description in [Server Setup](Server-Win-Mac#recording). |
 |       | `--norecord`      | Start recorder with recording disabled | (server only) |
 |    `-s` |`--server`         | Start in server mode | (server only) [See note](Choosing-a-Server-Type) |
--- a/wiki/en/en-Tips-Tricks-More.md
+++ b/wiki/en/en-Tips-Tricks-More.md
@@ -107,3 +107,42 @@ Fader strips in the mixer window are controlled in ascending order from left to 
 *Note*: Jamulus does not provide feedback on the state of the Solo and Mute buttons, meaning that your controller must keep track and toggle LEDs (if any) to 'on' or 'off' itself.
 
 Make sure you connect your MIDI device's output port to the Jamulus MIDI in port (QjackCtl (Linux), MIDI Studio (macOS) or whatever you use for managing connections). In Linux you will need to install and launch a2jmidid so your device shows up in the MIDI tab in Qjackctl.
+
+## Quality of Service
+
+Jamulus uses DSCP/CS4 to opportunistically deal with bufferbloat.<br>
+DSCP/CS4 is 128 (or 0x80) and is compatible with IPv4 and IPv6.<br>
+Other values can be set with: jamulus --qos [0..255]<br>
+( please see the informative RFC4594 and jamulus --help )<br>
+To disable QoS use: --qos 0 
+
+#### Windows Quality of Service instructions
+
+Prior to Windows Vista / Server 2008, Windows allowed software developers to set the
+Quality of Service field on IP traffic.
+With the release of Vista / Server 2008, Microsoft began to overwrite DSCP values
+that were set by applications with the Policy-based QoS layer.
+
+In order to enable Quality of Service you must follow these instructions:
+
+In Search box beside Start menu Type: Local Group Policy Editor (enter)<br>
+In new window, (click) on the menu icon to display the Actions third panel<br>
+Looking at the first panel of the Local Group Policy Editor<br>
+&nbsp;Local Computer Policy<br>
+&nbsp;&nbsp;Computer Configuration<br>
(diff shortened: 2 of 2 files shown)


gilgongo:
As a non-technical user, I'm trying to understand what I need to understand about  the -Q option. If I understand correctly, it implies that QoS is on by default at 128 ("To disable QoS use: --qos 0"), is that right? For other values, I should read an RFC, but what is the purpose of "Windows Quality of Service instructions"? Is it that I have to do this to get it to work on Windows? If so, do I need to do that on both client and server?

ghost:
> As a non-technical user, I'm trying to understand what I need to understand about the -Q option. If I understand correctly, it implies that QoS is on by default at 128

Yes

> ("To disable QoS use: --qos 0"), is that right?

Yes

> For other values, I should read an RFC,

Yes

> but what is the purpose of "Windows Quality of Service instructions"? Is it that I have to do this to get it to work on Windows?

Yes. The procedure shown will work with any version of Jamulus on Windows..

> If so, do I need to do that on both client and server?

No. The Windows QoS policy applies to Jamulus.exe (Which I believe is a client or server depending on cli options).
Note that Windows uses DSCP values from 0 to 63 which uses 6-bits. The values passed by --qos are 4 times larger because 8 bits are represented. That is why the Windows instructions have you set DSCP to 32, whereas, --qos is 128

Note that -Q and --qos apply to both client and server as a General option.

gilgongo:
@DavidSavinkoff  Thanks for the explanation. 

EDIT: just seen your https://github.com/jamulussoftware/jamulus/commit/377e6ec7e259332b856358520c5c4aff535d59a4#commitcomment-49361031 I'll look at that too.

So do I further understand that the Windows procedure is necessary in order to make the default value of 128 work? What happens to the default QoS in jamulus if I do not do the procedure?

Also, what is the effect on QoS if I set a different value on my client, and connect to a server that was running the default, and vice versa?

BTW I'm not asking for myself, but with a view to clarifying the documentation on this.

ghost:
The 'Windows procedure' makes a new DSCP QoS policy for Jamulus.exe
The 'Windows procedure' will set the DSCP value in the IP header for any version of Jamulus. In other words; The 'Windows procedure' works regardless of the new code that I made because it blindly over-writes what my code does. My code may work on older Windows versions, or if MSFT changes Policy based QoS.
Note that the 'Windows procedure' is for newer Windows versions that have Policy based QoS.

> So do I further understand that the Windows procedure is necessary in order to make the default value of 128 work?

No. Newer Windows versions overwrite this 128 with whatever value you use in the 'Windows procedure'.

> What happens to the default QoS in jamulus if I do not do the procedure?

Newer Windows versions overwrite this 128 with 0.

> Also, what is the effect on QoS if I set a different value on my client, and connect to a server that was running the default, and vice versa?

On non-Windows, each client and server are fully independent and can have different --qos values. One is free to experiment here.
On Windows, the Policy based QoS is set for Jamulus.exe. Thus, the clients and servers will have the same DSCP value set by the 'Windows procedure'. This is completely OK.
It may be possible to: copy Jamulus.exe Jamuluserver.exe and make a QoS Policy (Windows procedure) for each of Jamulus.exe and Jamuluserver.exe
Note that you may need to go through the 'Windows procedure' every time Jamulus is updated.
I believe that, over time, Windows users will come up with ideas on QoS policies which are better than what I'm suggesting.

gene96817:
What happens if some users are on Macs? Does this scheme require the server and all clients to be Windows-based?

gilgongo:
@gene96817 Macs use the Jamulus QoS OK as far as I understand - it's only Windows that will ignore it unless you reconfigure it not to do so. As to your second question, as stated above, "On non-Windows, each client and server are fully independent and can have different --qos values."

gilgongo:
@DavidSavinkoff  Now that I undertand more about QoS and the use of `-Q` on Jamulus, would it be OK if I re-wrote the help the website on it to read:

> Sets the quality of service DSCP value. Default is 128. Disable with 0.  Note: This option is ignored by Windows, but see [Tips & Tricks](Tips-Tricks-More#quality-of-service) for how to apply it.

And the Tips & Tricks preamble to read:

### Quality of Service

Jamulus uses DSCP/CS4 opportunistically to deal with buffer bloat, and uses a default DSCP/CS4 value of 128 (or 0x80). This is compatible with IPv4 and IPv6. Other values can be set using the `-Q` option, eg  `-Q [0..255]` (where 0 disables QoS). If you want to explore the effect of non-default settings, see [RFC4594](https://tools.ietf.org/html/rfc4594). However, most people will have no need to do this.

#### Using Quality of Service on Windows

Jamulus’s QoS settings (including the default) have no effect on Windows because the operating system ignores them. To enable Quality of Service for Jamulus, you must follow these instructions. Note also that you may need to repeat this procedure every time Jamulus is updated.

\[instructions for Windows\]

ghost:
> What happens if some users are on Macs?

DSCP Quality of Service is compatible with all Jamulus versions across all platforms.
I'm expecting that Macs will be identical to Linux in behaviour.

> Does this scheme require the server and all clients to be Windows-based?

No.

The 'Windows procedure' is an instruction for Windows users to make a QoS policy for Jamulus. If they don't, then their Jamulus-3.7.1 will operate with the same DSCP Quality of Service as Jamulus-3.7.0 did.
BTW if you do the 'Windows procedure' on your current Jamulus, you will get DSCP/CS4 in the IP header octet right now.

ghost:
> @DavidSavinkoff Now that I undertand more about QoS and the use of `-Q` on Jamulus, would it be OK if I re-wrote the help the website on it to read:
> 
> > Sets the quality of service DSCP value. Default is 128. Disable with 0.  Note: This option is ignored by Windows, but see [Tips & Tricks](Tips-Tricks-More#quality-of-service) for how to apply it.
> 
> And the Tips & Tricks preamble to read:
> ### Quality of Service
> 
> Jamulus uses DSCP/CS4 opportunistically to deal with buffer bloat, and uses a default DSCP/CS4 value of 128 (or 0x80). This is compatible with IPv4 and IPv6. Other values can be set using the `-Q` option, eg `-Q [0..255]` (where 0 disables QoS). If you want to explore the effect of non-default settings, see [RFC4594](https://tools.ietf.org/html/rfc4594). However, most people will have no need to do this.
> #### Using Quality of Service on Windows
> 
> Jamulus’s QoS settings (including the default) have no effect on Windows because the operating system ignores them. To enable Quality of Service for Jamulus, you must follow these instructions. Note also that you may need to repeat this procedure every time Jamulus is updated.
> 
> [instructions for Windows]

This looks very good. I read it like a lawyer. IANAL

PR #425, opened 2021-04-12
Incorporate Mastermix mode into Tips and Tricks

cwerling:
As suggest by @gilgongo here: https://github.com/jamulussoftware/jamulus/pull/1381#discussion_r611940711
Only merge if the respective PR was merged.

# Does this need translation?

- [x] Yes 
- [ ] No 

# Changes

### DIFF
--- a/wiki/en/en-Tips-Tricks-More.md
+++ b/wiki/en/en-Tips-Tricks-More.md
@@ -53,10 +53,19 @@ function loadserverstat ( $statfilename )
 
 You can save and restore the mix you have for your band rehearsals (fader, mute, pan, solo etc.) and load these any time (even while you are playing). Do this with "File > Save Mixer Channels Setup" in your client and load them using "Load Mixer Channels Setup" (or drag/drop them to the mixer window).
 
+## Delegate mixing to a mix master using the --mastermix server mode
+
+By starting a server instance with the `--mastermix` flag, the first client to connect will become the mix master.
+
+Subsequent joiners will be able to control their own as well as the mix master's pan and gain, but nobody else's (clients after version TODO will see their faders and buttons grayed out). Instead, they will hear the pan and gain value as set by the mix master.
+
+When this mix master sets a channel to SOLO, this will *not* apply to everybody's elses mix. Allowing the mix master to 'secretly' SOLO people can help creating a good mix without disturbing other clients.
+
+This mode can be valuable to e.g. choir rehearsals to offload mixing from the singers to a conductor or dedicated sound engineer and improve the quality of the overall mix for everybody.
+
 ## Converting a public server to a private one on the fly
 
 You can run a public server long enough for your band to connect, then go private by simply unchecking the 'Make my server Public' box in the server GUI. Your band mates will still be connected to the server until they disconnect. (Thanks to [David Savinkoff](https://github.com/DavidSavinkoff) for this tip!)
-
 
 ## Jamulus client Linux start script
 
--- a/wiki/en/en-Tips-Tricks-More.md
+++ b/wiki/en/en-Tips-Tricks-More.md
@@ -55,13 +55,13 @@ You can save and restore the mix you have for your band rehearsals (fader, mute,
 
 ## Delegate mixing to a mix master using the --mastermix server mode
 
-By starting a server instance with the `--mastermix` flag, the first client to connect will become the mix master.
+By starting a server instance with the `--mastermix` flag, the first client to connect will become the controller, or 'mix master' of the audio mix.
 
-Subsequent joiners will be able to control their own as well as the mix master's pan and gain, but nobody else's (clients after version TODO will see their faders and buttons grayed out). Instead, they will hear the pan and gain value as set by the mix master.
+Subsequent joiners will be able to control their own as well as the mix master's pan and gain, but nobody else's (clients will see these faders and buttons grayed out). Instead, they will hear the pan and gain value as set by the mix master.
 
-When this mix master sets a channel to SOLO, this will *not* apply to everybody's elses mix. Allowing the mix master to 'secretly' SOLO people can help creating a good mix without disturbing other clients.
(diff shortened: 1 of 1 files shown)


wiki/en/en-Tips-Tricks-More.md
```diff
@@ -53,11 +53,20 @@ function loadserverstat ( $statfilename )
 
 You can save and restore the mix you have for your band rehearsals (fader, mute, pan, solo etc.) and load these any time (even while you are playing). Do this with "File > Save Mixer Channels Setup" in your client and load them using "Load Mixer Channels Setup" (or drag/drop them to the mixer window).
 
+## Delegate mixing to a mix master using the --mastermix server mode
+
+By starting a server instance with the `--mastermix` flag, the first client to connect will become the mix master.
```
gilgongo: Suggest you say ", the first client to connect will become the controller, or 'master' of the audio mix." Otherwise you'll be in danger of defining a term using that term (ie "a mix master is a mix master"). Fine to refer to "mix master" after that though.

wiki/en/en-Tips-Tricks-More.md
```diff
@@ -53,11 +53,20 @@ function loadserverstat ( $statfilename )
...
+
+Subsequent joiners will be able to control their own as well as the mix master's pan and gain, but nobody else's (clients after version TODO will see their faders and buttons grayed out). Instead, they will hear the pan and gain value as set by the mix master.
```
gilgongo: We have usually kept the docs relevant to the current version and then update them once the new version is out.

I'm not sure I understand "...  as well as the mix master's pan and gain" though. Does that mean they can override the master mix? But that would negate the point. Or does it mean they can control the mix master's channel and theirs, but nobody else's?

wiki/en/en-Tips-Tricks-More.md
```diff
@@ -53,11 +53,20 @@ function loadserverstat ( $statfilename )
...
+
+When this mix master sets a channel to SOLO, this will *not* apply to everybody's elses mix. Allowing the mix master to 'secretly' SOLO people can help creating a good mix without disturbing other clients.
```
gilgongo: House style has been to refer to buttons in the UI with sentence case, so this would be "sets a channel to Solo" and "'secretly' solo" (the latter because its just a verb not a button that's being referred to). 

Meta: This is a good example of the use of "channel" and "people" as per the debate about when to use "channel". I think. I can't remember, and now can't find the discussion on this, but were we saying that we should use "channel" only to refer to system audio (eg "the server mixes all audio channels", "channel panning"), otherwise everyone is a "person" unless you want to refer to the UI, in which case it's a "fader" (eg "the icon above a person's fader")?

gilgongo: "can help to create a good mix" I think.

wiki/en/en-Tips-Tricks-More.md
```diff
@@ -53,11 +53,20 @@ function loadserverstat ( $statfilename )
...
+
+This mode can be valuable to e.g. choir rehearsals to offload mixing from the singers to a conductor or dedicated sound engineer and improve the quality of the overall mix for everybody.
```
gilgongo: Slightly confusing about the actors. I'd split it into two (and remove the "e.g.") as "This mode can be valuable for example in choir rehearsals to offload mixing from the singers to a conductor. Or in having a dedicated sound engineer to improve the quality of the overall mix for everybody."

gilgongo (changes requested):
Thanks! I think we might move this to a dedicated server documentation page later. But until then this is fine I think.

wiki/en/en-Tips-Tricks-More.md
pljones: The problem with "person" is a "person" can be playing several instruments and run multiple clients, each with the same name.  Whenever we need to refer to identifying controls on the mixer - i.e. when referring to actions that would analogously be undertaken on a physical mixing desk - then "channels" should be used, in my view.

gilgongo:
@cwerling Can you have a look at the suggestions on this? Thanks.

wiki/en/en-Tips-Tricks-More.md
gilgongo: > The problem with "person" is a "person" can be playing several instruments and run multiple clients,

Yes, I think it was for reasons like this that we wanted to use "person", so that it could use as opposed to "channel", for clarity. So for example, "You can group a person's channels..." and "People connected to the server". If we didn't have "person" and always used "channel" then it would be less clear in some contexts perhaps.

>  when referring to actions that would analogously be undertaken on a physical mixing desk

Although I recall we decided that we'd use "fader" in that case, eg "Lower the channel's fader" not "Lower the volume on the channel", reserving "channel" for the "system thing" if we were talking about something from the user's POV.

Perhaps bit of a subtle distinction, but it was in an effort to have three distinct terms for a human using a server ("person"), a UI element ("fader"), and the system thing ("channel") so as make things a bit clearer.

So this would mean we'd say eg, "Muting a person's channel will display a muted icon above your fader" NOT "Muting a channel will display a muted icon above your channel".

wiki/en/en-Tips-Tricks-More.md
cwerling: > Or does it mean they can control the mix master's channel and theirs, but nobody else's?

Exactly this. I know it might sound weird, but it's exactly what we need in our choir rehearsals, since 
1. everybody still has individual needs for how low the conductor/click is
2. the conductor mutes himself so he wouldn't be able to mix his own channel in the master mix

cwerling:
> @cwerling Can you have a look at the suggestions on this? Thanks.

Thanks for the suggestions and sorry for not coming back to this earlier. I think I incorporated them almost as is, but feel free to make some more corrections still!

gilgongo (approved):

PR #428, opened 2021-04-15
Added steps that work on Fedora 33

jp8:
# Does this need translation?

- [ ] Yes 
- [X ] No 

# Changes

just an update to the compile guide on linux


gilgongo:
Hi @jp8 and thanks for this. We are currently re-writing the compilation instructions for supported platforms now that binary installs are the preferred method on Linux, so will incorporate these changes into them if that's OK.

PR #433, opened 2021-04-18
Change links to edit on release branch

ann0see:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

A lot of confusion with branches has happened in the past. Therefore we should redirect people to edit on the release branch and redirect them to changes only if new content/content that needs translation is added.


hoffie (approved):
Yes, I think this makes sense, considering that most content for `changes` comes from reguar contributors who already know about it, while edits from the web page are likely to come from people who are new to the project.

ann0see:
Ok. Just merged it.

PR #415, opened 2021-04-04
Update Linux client/server for binary installs

gilgongo:
# Does this need translation?

- [x] Yes 
- [ ] No 

# Changes

Move all client/server compilation instructions to COMPILING.md. From now on, we assume binary installs, and further assume that people who choose to compile don’t need hand holding.

PR #440, opened 2021-04-25
Update Fedora build dependency

fwiep:
# Does this need translation?

- [x] Yes 

# Changes

Replaced the `libQt5Concurrent5` package (which is no longer available in Fedora) to `qt5-qtbase`, which provides the necessary libraries.

### DIFF
--- a/wiki/en/en-Installation-for-Linux.md
+++ b/wiki/en/en-Installation-for-Linux.md
@@ -64,7 +64,7 @@ sudo apt-get install build-essential qt5-qmake qtdeclarative5-dev qt5-default qt
 On **Fedora**:
 
 ```shell
-sudo dnf install qt5-qtdeclarative-devel jack-audio-connection-kit-dbus libQt5Concurrent5 jack-audio-connection-kit-devel
+sudo dnf install qt5-qtdeclarative-devel jack-audio-connection-kit-dbus qt5-qtbase jack-audio-connection-kit-devel
 ```
 
 ### QjackCtl: Optional, but recommended


wiki/en/en-Installation-for-Linux.md
```diff
@@ -64,7 +64,7 @@ sudo apt-get install build-essential qt5-qmake qtdeclarative5-dev qt5-default qt
 On **Fedora**:
 
 ```shell
-sudo dnf install qt5-qtdeclarative-devel jack-audio-connection-kit-dbus libQt5Concurrent5 jack-audio-connection-kit-devel
+sudo dnf install qt5-qtdeclarative-devel jack-audio-connection-kit-dbus qt5-qtbase jack-audio-connection-kit-devel
```
pljones, line 67: Doesn't `aqtinstall` work for Qt on Fedora (and most platforms (edit)maybe(/edit) except Windows)?

fwiep: I have only searched for a replacement package in Fedora's standard repos. `aqtinstall` seems not to be in there, but `qt5-qtbase` is.

pljones: See the autobuild script for MacOS:
https://github.com/jamulussoftware/jamulus/blob/master/autobuild/mac/artifacts/autobuild_mac_1_prepare.sh

pljones: We use it for the MacOS build -- I wondered if it would save platform-specifics having to be written for the other platforms but it seems the autobuild is very platform specific still...

https://github.com/jamulussoftware/jamulus/blob/master/autobuild/mac/artifacts/autobuild_mac_1_prepare.sh

gilgongo:
This doesn't seem to be the same as the assertion about libraries here though: https://github.com/jamulussoftware/jamuluswebsite/pull/428/files

wiki/en/en-Installation-for-Linux.md
hoffie: > Doesn't `aqtinstall` work for Qt on Fedora (and most platforms (edit)maybe(/edit) except Windows)?

I can't comment regarding Mac, but I don't think we should require additional packets or scripts (?) on Linux where dependencies can be installed by the native package managers (apt, dnf, yum, etc.). At least it should be like this for users building from source and distributors. As such, it would be the best check for us to do the same in autobuild.

hoffie:
yum/dnf support installing by path (e.g. `yum install /lib/libfoo.so`). Would this help to provide a single list for all supported Fedora versions?

wiki/en/en-Installation-for-Linux.md
softins: I always thought `aqtinstall` was just a way of being able to install a specific version of Qt into a build environment _for Mac only_, so that the installation package produced would contain the version of Qt runtimes required for backward compatibility. For Fedora/CentOS/Debian/Ubuntu, there is no need for `aqtinstall`, because we just use the packaged version of Qt provided by the distro.

gilgongo:
Hi @fwiep - thanks for this. We are in fact re-writing the installation docs as part of https://github.com/jamulussoftware/jamuluswebsite/pull/415 and I've incorporated these changes in that. I'll therefore close his PR, and if people want to discuss the aqtinstall stuff they are welcome to open a thread on Discussions about that.

PR #454, opened 2021-05-02
correcting link to linux intstallation file name

jujudusud:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

There is an error in installation-linux page link. correcting with this commit.


pljones (approved):

hoffie (approved):
Thanks!

This was in `changes` only, so the target branch seems correct. Merging.

PR #466, opened 2021-05-09
Linux install corrections

gilgongo:
# Does this need translation?

- [x] Yes 
- [ ] No 

# Changes

The install command (which could also be run by double-clicking on the file) needs an explicit `./ ` on the path. The .deb installs QjackCtl as a dependency automagically, so no need separately to install that or to launch from the command line.


wiki/en/en-Installation-for-Linux.md
```diff
@@ -13,9 +13,9 @@ Make sure you read the [Getting Started](Getting-Started) page.
 ### Debian and Ubuntu
 
 1. Download the [latest .deb file]({{ site.download_root_link }}{{ site.download_file_names.deb-gui }})
-1. Update apt: `sudo apt-get update`
-1. Install the package from where you downloaded it: `sudo apt install {{ site.download_file_names.deb-gui }}`.
-1. Since Jamulus needs the JACK server, you have to install it too. We recommend to use `QjackCtl` to configure JACK. You can install it via `sudo apt-get install qjackctl`
+1. Update apt by opening a console window (CTRL+ALT+T should work) and type: `sudo apt-get update`
+1. Navigate to where you downloaded the installer and either double-click on it, or use the command line: `sudo apt install ./{{ site.download_file_names.deb-gui }}`.
+1. Once installed, yuo can delete the file and close any console windows.
```
jujudusud, line 18: you instead of yuo line 18

PR #475, opened 2021-05-17
#457 Updated ES-Software Manual

ignotus666:
# Does this need translation?

- [ ] Yes 
- [* ] No 

# Changes

Updated the Software Manual to be level with the version in the translate branch after noticing it wasn't up to date.


hoffie:
@ignotus666 What's the reason for closing this PR? Is it obsolete for some reason?

ignotus666:
I realised the original file was ok and that this PR would actually revert it to an earlier state... I was doing stuff with multiple branches and lost track of which one contained the latest EN files. Sorry for the confusion!

PR #483, opened 2021-05-24
Updated _data files for 3_8_0

ignotus666:
# Does this need translation?

- [ ] Yes 
- [* ] No 

# Changes

### DIFF
--- a/_data/es/navigation.yml
+++ b/_data/es/navigation.yml
@@ -1,8 +1,8 @@
 nav:
 - title:                 Cómo usar Jamulus
   subfolderitems:
-    - page:              Cómo Empezar
-      url:               /wiki/Getting-Started
+    - page:              Configuración
+      url:               /wiki/Setup
       subsubfolderitems:
         - page:          Windows
           url:           /wiki/Installation-for-Windows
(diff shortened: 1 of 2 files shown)


_data/es/navigation.yml
```diff
@@ -1,17 +1,15 @@
 nav:
 - title:                 Cómo usar Jamulus
   subfolderitems:
-    - page:              Cómo Empezar
-      url:               /wiki/Getting-Started
+    - page:              Configuración
+      url:               /wiki/Setup
```
ann0see, line 5: Not sure, but I think @gilgongo would like to change it back to Getting-Started. But I think he'll do it close to the release.

ann0see (approved):
Not yet merging since there's still one open question:

_data/es/navigation.yml
gilgongo: Yes, I was going to just find/replace all the links to /wiki/Setup to /wiki/Getting-Started once I've moved the Setup content into that file so we can then remove *-Setup.md and have one less redirect.

ann0see: Ok. So this means that we can just merge this as is

PR #494, opened 2021-05-29
Add permission handling to GH Actions

ann0see:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes
Limit permissions to Jekyll workflow (already did a test)
# Related
https://github.com/jamulussoftware/jamulus/issues/1737

### DIFF
--- a/.github/workflows/jekyll.yml
+++ b/.github/workflows/jekyll.yml
@@ -2,14 +2,15 @@ name: Jekyll site CI
 
 on:
   push:
-    branches: [ changes, release, translate ]
+    branches: [ changes, release ]
   pull_request:
-    branches: [ changes, release, translate ]
+    branches: [ changes, release ]
 
 jobs:
   build:
     permissions:
       contents: read
+      pull-requests: read
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/jekyll.yml
```diff
@@ -2,14 +2,15 @@ name: Jekyll site CI
...
+    branches: [ changes, release ]
   pull_request:
-    branches: [ changes, release, translate ]
+    branches: [ changes, release ]
 
 jobs:
   build:
     permissions:
       contents: read
+      pull-requests: read
```
ann0see, line 13: Is pull-requests read needed?

.github/workflows/main.yml
```diff
@@ -10,7 +10,10 @@ on:
 
 jobs:
   jekyll:
-    runs-on: ubuntu-16.04 # can change this to ubuntu-latest if you prefer
+    permissions:
+      deployments: write
```
ann0see, line 14: Referencing: https://docs.github.com/en/rest/reference/permissions-required-for-github-apps#permission-on-deployments to be checked! I think. this might not even be necessary, since we just put html on the gh-pages branch and GitHub deploys it internally.

ann0see (commented):
I think we should have a look at the API documentation

PR #517, opened 2021-06-06
Anchor links won't work with capital letters

### LINKED ISSUE #516: Viewing the logs doesn't work as specified (wiki/en/Server-Linux.md)
https://jamulus.io/wiki/Server-Linux
wiki/en/Server-Linux.md

Error in Section "Viewing the Logs".

Currently, the shown commands for viewing the logs refer to systemd unit `jamulus` (e.g. `journalctl -u jamulus`).
It should be `jamulus-headless` - otherwise it won't work.

-- 
Note: This is my first issue on github :)

ignotus666:
# Does this need translation?

- [ ] Yes 
- [* ] No 

# Changes

Corrected anchor link in Command-Line-Options. Capital letters in anchor links cause them to not work.

Edit: also updated scripts with suggestions by @ann0see.

In other news: A side-effect of running po4a-update-templates.sh is that the timestamp gets changed in .po and .pot files, cluttering PRs with changes to 100s of files (see this one...). I've read in several places that Git ignores timestamp changes but it doesn't seem to be happening here. If in future it's done with a GH action it won't be a problem with PRs but I imagine it would seriously clutter the commit history. There must be some kind of filter that can be used but I don't know how to set that up.


ann0see:
> A side-effect of running po4a-update-templates.sh is that the timestamp gets changed in .po and .pot files, cluttering PRs with changes to 100s of files (see this one...).

Could you please raise an issue about that?


So this is save to merge, I think?

ignotus666:
Yes, I just made a couple more changes but it's tested and should be safe to merge. I've been looking more into how to solve the timestamp issue but it's beyond me - I'll open an issue for it. That said, I'm running the update script but once it gets automated it shouldn't be that big an issue as translators won't be running it and it's not something that will be done that often - only once EN files have reached a definitive (well, relatively speaking) state for translation. So it won't be necessary to trawl through PRs as they won't have those changes. But I do suppose it unnecessarily pollutes the commit history.

po4a-create-all-targets.sh
```diff
@@ -31,7 +31,7 @@ fi
 
 # Check if po4a is installed
 if ! [ -x "$(command -v po4a)" ] ; then
-    echo 'Error: please install po4a.' >&2
+    echo "Error: please install po4a." >&2
```
pljones: Technically, quotes not needed at all...  (`echo` doesn't have an arg count limit.)

I tend to use double quotes only when I expect expansion within them, otherwise sticking to single quotes is "safer".

po4a-update-templates.sh
```diff
@@ -30,7 +30,7 @@ fi
 
 # Check if po4a is installed
 if ! [ -x "$(command -v po4a)" ] ; then
-    echo 'Error: please install po4a.' >&2
+    echo "Error: please install po4a." >&2
```
pljones, line 33: Same as a `po4a-create-all-targets.sh` -- I'd question the use of double quotes in preference to single quotes.

po4a-update-templates.sh
```diff
@@ -52,9 +52,9 @@ do
     path="${dirname#$SRC_DIR/}"
 
     if [ "$dirname" = "$SRC_DIR" ]; then
-        potname=$basename.pot
+        potname="$basename".pot
```
pljones, line 55: These two need the double quotes, as you don't know if the path will contain a space, which would otherwise cause the shell some problems...

po4a-update-templates.sh
```diff
@@ -52,9 +52,9 @@ do
...
     else
-        potname=$path/$basename.pot
+        potname="$path/$basename".pot
```
pljones, line 57: I'd put the whole thing (including the trailing `.pot`) inside the double quotes, just feels tidier.

ignotus666: Thanks, I've implemented your suggestions in #522

PR #518, opened 2021-06-06
Update README.md, remove CONTRIBUTING.md

gilgongo:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

Adding the intro from CONTRIBUTING.md (which may not be correct now we have po4a) so that it can be usefully deleted.


README.md
```diff
@@ -1,6 +1,11 @@
 # Jamulus Website
 
-This is the content repository for [jamulus.io](https://jamulus.io). Text appearing in the Jamulus software application itself is part of the [main code repository](https://github.com/jamulussoftware/jamulus). The web site is built with Jekyll and Fox-CSS.
+This is the content repository for [jamulus.io](https://jamulus.io). Text appearing in the Jamulus software application itself is part of the [main code repository](https://github.com/jamulussoftware/jamulus). See also the [general overview](https://jamulus.io/wiki/Contribution) to getting involved. 
+
+
+## Want to help translate?
+
+To make updates or translations for jamulus.io you will need a GitHub account and be familiar with GitHub in general. If you want to get more involved, you’ll need to be familiar with the use of git on the command line, as well as Jekyll. Note that translating the Jamulus application itself is done by a different process, but you are welcome to help with that too. Introduce yourself on the [GitHub Discussion](https://github.com/jamulussoftware/jamulus/discussions) and say what you’d like to do.
```
gilgongo, line 8: With po4a in the mix, I wonder if maybe something like "You will need a GitHub account and be familiar with GitHub in general, and will need to use [some tool] to update .po files. Note that translating the Jamulus application itself is done by a similar process, and you are welcome to help with that too."

ignotus666: In reality, translators shouldn't need to worry about po4a, as (ideally) that would be taken care of automatically later by the repo. They certainly shouldn't need to ever run po4a-update-templates.sh unless they've updated EN files, but that's outside the scope of translation. If it's done by a GH action, po4a-create-all-targets.sh could be run whenever a translator submits .po files and they get merged. We could explain somewhere how to generate .md files locally (currently in /wiki/README), but it wouldn't be a requirement. Also, it's limited to Linux users.

ann0see (approved):

PR #522, opened 2021-06-07
Simplify .md folder deletion command and replace path with variable

ignotus666:
# Does this need translation?

- [ ] Yes 
- [* ] No 

# Changes

See title

### DIFF
--- a/po4a-create-all-targets.sh
+++ b/po4a-create-all-targets.sh
@@ -67,7 +67,7 @@ use_po_module () {
 		dirname=$(dirname "$file")
 		path="${dirname#$SRC_DIR/}"
 
-		if [ "$dirname" = "$SRC_DIR" ]; then
+		if [ "$dirname" = "$SRC_DIR" ] ; then
 			potname=${basename}
 			localized_file="$PUB_DIR/$lang/$basename.md"
 		else
--- a/po4a-create-all-targets.sh
+++ b/po4a-create-all-targets.sh
@@ -34,8 +34,8 @@ fi
 
 # Check if po4a is installed
 if ! [ -x "$(command -v po4a)" ] ; then
-	echo "Error: please install po4a." >&2
-	exit 1
+    echo "Error: please install po4a." >&2
+    exit 1
 fi
 
 # Check if source document folder exists in the right place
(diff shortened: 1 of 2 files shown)


po4a-create-all-targets.sh
```diff
@@ -48,14 +48,10 @@ fi
...
 
-while IFS= read -r -d '' dir
-do
-	lang=$(basename -s .md "$dir")
-	echo "delete $lang folder"
-	cd "$PUB_DIR"
-	rm -rf "$lang" 
-	cd ../  
-done <   <(find "$PO_DIR" -mindepth 1 -maxdepth 1 -type d -print0)
+for lang in $(ls "$PO_DIR" ) ; do
```
ann0see, line 51: Wouldn’t this delete everything (=all files) in $PO_DIR? 

You could then just `rm -rf *` I‘m a bit worried here.

ignotus666: No, it takes the directory names in PO_DIR as a reference to delete them in /wiki. That way it only deletes supported language folders (if they have a .po folder they have a counterpart in /wiki). I tested it before submitting ;)

ann0see: Yes. You’re running rm -r on $PUB_DIR

po4a-create-all-targets.sh
```diff
@@ -94,7 +94,7 @@ use_po_module () {
 while IFS= read -r -d '' dir
 do
 	lang=$(basename -s .md "$dir")
-	echo "$lang"
+	echo '$lang'
```
ann0see: Wait. I think you need double quotes here to allow the content of variables be shown not the string $lang

ann0see (changes requested):

po4a-create-all-targets.sh
ignotus666: Ah yes, I got a bit overenthusiastic replacing the double quotes ...

ignotus666: Hmmm, I'm sure I tested it before submitting though... I'll have a look tomorrow, can't check now.

pljones: Well, no quotes at all are required as it's just `echo` - but definitely not single quotes...

po4a-update-templates.sh
```diff
@@ -30,13 +30,13 @@ fi
...
 if ! [ -x "$(command -v po4a)" ] ; then
-    echo "Error: please install po4a." >&2
+    echo 'Error: please install po4a.' >&2
     exit 1
 fi
 
 # Check if source document folder exists in the right place
 if [ ! -d "$SRC_DIR" ] ; then
-    echo "Error: please run this script from the root folder"
+    echo 'Error: please run this script from the root folder'
```
pljones: This should be redirected to stderr, too (like line 33).

po4a-update-templates.sh
```diff
@@ -77,16 +77,16 @@ do
             --master "$file" \
             --master-charset "UTF-8" \
             --po "$po_file" ; then
-        echo ""
-        echo "Error updating $lang PO file for: $adoc_file"
+        echo ''
+        echo 'Error updating $lang PO file for: $adoc_file'
```
pljones: Not much of an error as it doesn't cause termination of the script.  Should it?

Also, this second line doesn't want the single quotes as it's after variable expansion.

ignotus666: Well, it continues to update the rest of the po files. Depends on whether you want it to just exit regardless or not.

I'm working on the quote drama now. Getting rid of them all except where needed.

ann0see (approved):

Issue #488, opened 2021-05-27
Use button for installation guides


ann0see:
Is there any reason why the buttons for the installation guides have been removed in the upcoming website release and are now just "invisible" links? I think we should still make them prominent on the Setup page to make the workflow easy.

gilgongo:
The reason is that I would think most people can probably skip that section since they'll have already installed Jamulus by the time they're on that page. So it doesn't need to heavily called out. In fact once we have a more prominent "Download" link on the new nav with a single download page explaining the basics, I was thinking of removing that section anyway.

ann0see:
Ok. But this is only true as soon as the downloads page exists

gilgongo:
OK sure.

ann0see:
So I would suggest to add the buttons for now to make the guides more visible

gilgongo:
Fixed with https://github.com/jamulussoftware/jamuluswebsite/pull/524

PR #523, opened 2021-06-07
Modify installation guide for Windows and Getting-Started page

ann0see:
# Does this need translation?

- [x] Yes 
- [ ] No 

# Changes

* Add links to troubleshooting page
* Add Hardware Setup paragraph to Getting-Started page

### DIFF
--- a/wiki/en/Getting-Started.md
+++ b/wiki/en/Getting-Started.md
@@ -27,7 +27,7 @@ And for now, close all other applications. It’s best to start simple at first.
 
 ## Hardware Setup
 
-If you use external audio hardware, plug that in before you start Jamulus. If you haven't set it up, please see the installation guides mentioned above.
+If you use external audio hardware, plug that in before you start Jamulus. If you haven't configured your hardware, please see the installation guides mentioned above.
 
 ## Jamulus main window
 
--- a/wiki/en/Getting-Started.md
+++ b/wiki/en/Getting-Started.md
@@ -64,7 +64,7 @@ Now use the “Connect” button in the Jamulus main window to join a server. A 
   <figcaption>Connect to a server</figcaption>
 </figure>
 
-The most important thing about servers is their “ping time”. The bigger the number, the harder it will be to play in time with others. Usually, you would select a server with a ping of less than 15ms if you can.
+The most important thing about servers is their “ping time”. The bigger the number, the harder it will be to play in time with others. Usually, you would select a server with a ping of less than 50ms if you can.
 
 Once connected to a server, make sure you can hear yourself OK and fix any input volume or other problems. You can have a look at the [troubleshooting page](/wiki/Client-Troubleshooting) for common problems. And of course, check that you are also able to obey [The Golden Rule](/wiki/Client-Troubleshooting#you-all-sound-ok-but-its-difficult-to-keep-together).
 
--- a/wiki/en/Getting-Started.md
+++ b/wiki/en/Getting-Started.md
@@ -17,13 +17,17 @@ If you have any questions, please see the [Discussions](https://github.com/jamul
 
 ## Installation
 
-To get Jamulus working with your hardware, please follow the installation guide for your platform:
+To get Jamulus working with your operating system, please follow the installation guide for your platform:
 
 * [Windows](/wiki/Installation-for-Windows)
 * [macOS](/wiki/Installation-for-Macintosh)
 * [Linux](/wiki/Installation-for-Linux)
 
 And for now, close all other applications. It’s best to start simple at first.
+
+## Hardware Setup
+
+If you use external audio hardware, plug that in before you start Jamulus. If you haven't set it up, please see the installation guides mentioned above.
 
 ## Jamulus main window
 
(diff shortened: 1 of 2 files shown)


wiki/en/Getting-Started.md
```diff
@@ -17,14 +17,18 @@ If you have any questions, please see the [Discussions](https://github.com/jamul
...
 
 * [Windows](/wiki/Installation-for-Windows)
 * [macOS](/wiki/Installation-for-Macintosh)
 * [Linux](/wiki/Installation-for-Linux)
 
 And for now, close all other applications. It’s best to start simple at first.
 
+## Hardware Setup
+
+If you use external audio hardware, plug that in before you start Jamulus. If you haven't set it up, please see the installation guides mentioned above.
```
pljones: "If you haven't set it up" - what's the "it", your hardware or Jamulus?

ann0see:
```suggestion
If you use external audio hardware, plug that in before you start Jamulus. If you haven't configured your hardware, please see the installation guides mentioned above.
```

ann0see: Maybe that’s the problem here:

you might need to configure your hardware (turn up gain,…) and set it up in JACK/ASIO4ALL/… see my proposal

gilgongo: Maybe better to add the thing about plugging in hardware to the platform-specific guides (which don't currently say this) and remove it here to avoid confusion. I think it's unlikely people with external hardware would not have it plugged in at this stage, and "configuration" is platform specific. This guide is for all platforms.

gilgongo: BTW I thought this issue was about the links on this page: https://github.com/jamulussoftware/jamuluswebsite/issues/488 If so, you may as well put them in as part of this PR.

ann0see: Adding it to every page means duplication. The main intention about that was that the most problems seem to occur with wrongly configured hardware.

ann0see: > BTW I thought this issue was about the links on this page: #488 If so, you may as well put them in as part of this PR.

Can do that.

ann0see: See: https://github.com/jamulussoftware/jamuluswebsite/pull/524 it might also go live right away (Fuzzy matching?)

gilgongo: > Adding it to every page means duplication

Not sure I understand. If we say  you should plug in your hardware before running Jamulus here, but don't say it in the section about installing Jamulus earlier, what is the benefit of saying it here for the first time?  Perhaps a minor thing, but seems a little confusing. 

On balance, I wouldn't say it at all since I highly doubt anyone would not have their hardware attached to their machine when investigating how to use Jamulus.

ignotus666: > See: #524 it might also go live right away (Fuzzy matching?)

Since they're links that don't need translation (would be copied from source segment), if you run the update and then create all target scripts, the new links will appear in translations without having to do anything. Translators could validate them eventually but it's not strictly necessary, other that it "artificially" lowers the % of translated text for threshold purposes.

ann0see: > what is the benefit of saying it here for the first time? 

My original intention here was to hint multiple user groups (somewhat experienced PC user without pro audio knowledge should get a hint at the fact that the gear plays a role; inexperienced user who is already overwhelmed by the fact that he needs to install software -> needs a step by step guide) at one of the probably most challenging UX problems: setting up the audio gear. If it was part of every OS setup page, we‘ll need to write the same text (Plug in Hardware/configure your internal sound card/…) everywhere. Maybe it’s more of related to the UX issue of Windows setup?

Maybe this needs some more general discussion/thoughts?

gilgongo: No that's OK, it's just that if we say "If you haven't configured your hardware... " then that leaves the user to ask themselves if they have configured their hardware or not, and if they think they haven't, they have to go to the installation page they probably just came from where we don't in fact tell them to plug in their hardware (not sure why, maybe because it's obvious?) and that has platform-specific instructions (well, for Windows and Linux anyway). Maybe it's just a minor thing.

wiki/en/Getting-Started.md
```diff
@@ -62,7 +66,7 @@ Now use the “Connect” button in the Jamulus main window to join a server. A
 
 The most important thing about servers is their “ping time”. The bigger the number, the harder it will be to play in time with others. Usually, you would select a server with a ping of less than 15ms if you can.
```
gilgongo:
```suggestion
The most important thing about servers is their “ping time”. The bigger the number, the harder it will be to play in time with others. Usually, you would select a server with a ping of less than 50ms if you can.
```

Increasing the reccommended ping time might encourage people to use more public servers, as @jujudusud points out.

gilgongo (approved):

wiki/en/Getting-Started.md
ann0see: 50ms is quite high… 

30ms? However, it all depends on the hardware too. 

Why don’t we just say "marked green or yellow" in the connection dialog?

jujudusud: As discussed with pljones, 50ms is the real maximum we had used for a server.
We want to encourage users to connect to the existing public server.

ann0see: Ok. So applied that and merged PR.

PR #524, opened 2021-06-09
Add buttons for installation guides to Setup

### LINKED ISSUE #488: Use button for installation guides

ann0see:
# Does this need translation?

Depends if we want to go live right away.

# Changes

This PR adds buttons to the Getting-Started page.

Fixes #488


gilgongo (approved):
This does look pretty disracting for something that I hope won't be relevant to a lot of people going to this page. But it's not a big deal.

jujudusud (approved):
I think we should wait for 3.8.1 if it is not in 2022.

PR #529, opened 2021-06-15
Simplify po4a-update-templates.sh script, remove .pot files, update READMEs

ignotus666:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

- Some of the steps in the po4a-update-templates.sh script were superfluous so I removed them.
- Removed .pot files.
- Added a message that is shown to clarify an error displayed when new files are added to /wiki/en/.
- Updated READMEs to reflect path changes and explain how to add a new language (simpler now).

I've tested it and everything seems to work as before.


wiki/README.md
```diff
@@ -20,3 +20,4 @@ In order to run the `po4a-update-templates.sh` and `po4a-create-all-targets.sh`
 
 - If you want something displayed on a new line immediately below one that does not end with a period, the second line must start with a space; e.g. if you want a line break after } or ). For code blocks introduced with ``` or ~~~, these characters must be followed by a line break (enter) and a space at the start of the next line. The closing formatting characters should be on their own line too.
 - Any corrections made to languages other than EN **must** be done to their .po files. Directly editing .md files won't break anything but any changes done that way will be lost the next time `po4a-create-all-targets.sh` is run.
+- Headings introduced with # must have double spacing before the text body.
```
gilgongo, line 23: Is this something we can check for in linting? I'm sure I'd not be able to remember to do this,

ignotus666: I must admit the wording is a bit messy and could do with a review. These started out as "notes to self" and might not be fully accurate. There were I think 4 examples where po4a messed up the formatting and it was always related to how it interprets where a paragraph ends. If you want I could provide the specific examples to make it clearer.

ignotus666: Just to add - I cloned your adm-man branch and processed all the wiki/en files with po4a to see if any of the new stuff came out wrong. It's all fine, with no weird formatting, so I don't think it's going to be too much of an issue.

ann0see:
Something is wrong here: I ran both scripts and the resulting .md files now have a lang: en attribute, not e.g. de --> wrong translation! Do we need to manually change something?

ann0see (changes requested):
See above

ignotus666:
> Something is wrong here: I ran both scripts and the resulting .md files now have a lang: en attribute, not e.g. de --> wrong translation! Do we need to manually change something?

That's weird - I don't see that happening. To be clear: you see this in the translated .md files in /wiki?

ann0see:
Yes. The lang attribute for de is wrong and I think it shows spanish instead of english for untranslated files.

ignotus666:
I've found the problem. I take it you're using the po4a version from the repos (0.57?). I just tested with that version (I'd been using the latest compiled from the GH page) and what you describe happens. So it looks like >=0.63 is required. Would this be a problem?

*Edit: I also tried 0.62-1 from the Debian-main repo and that doesn't work either. If this will present a problem I'll roll back to using the previous setup with the .pot files.

ignotus666:
Since I'm not entirely sure whether it works without .pot files by design or due to a quirk, I think it's best to stick to "standard procedure" (use .pot files) and keep compatibility with older versions of po4a. So I'll close this PR and prepare a new one.

ignotus666:
Ugh... I just did some more tests and it seems any version <0.63 messes up the yaml front matter section. It seems to be a bug that was fixed in 0.63. All my testing up until now had been done inadvertently with 0.63/0.64 which is why I never noticed this. I'd installed 0.57 via Synaptic thinking it would override it, to ensure it would work with older repo versions, but I was on 0.64 all along...

@ann0see is this a screw-up or nothing major? :-(

ann0see:
If it’s just die to an old version of po4a we can easily state that a new version should be used and give an error/warning of an old version is found

ignotus666:
Ok that's good, but I was thinking about if we used GH actions - the versions that have to be used need to be compiled as they're not available through any repo. I thought it might be an issue but I don't really know how that works.

I've created a new branch and will submit a new PR from there (I think I'm not meant to be making PRs from 'release' am I?). I'll add a release version check to the scripts then and add that info to the relevant READMEs.

ann0see:
Yup. You shouldn’t raise PRs directly from release/master/main/… 

I usually do the following:

1. Add the upstream repo as separate remote
`git remote add upstream https://GitHub.com/jamulussoftware/jamuluswebsite` 
2. fetch the upstream repo: `git fetch upstream` 
3. Checkout the release/… branch: `git checkout upstream/release`
4. create a branch from it: `git checkout -b <name>` 
5. And then push everything to my own repo

PR #533, opened 2021-06-20
Simplify scripts, remove .pot files, update READMEs

ignotus666:
# Does this need translation?

- [ ] Yes 
- [* ] No 

# Changes

Removed superfluous commands from scripts and .pot files. Added a check to make sure po4a version 0.63 or higher is installed. Updated READMEs to reflect changes.


ann0see:
I'm now trying to install po4a (latest version) on Debian testing. It seems to take quite some time. Do you know if there really isn't any binary available? We can't compile it every time via GH actions...

How did you get the updated version of po4a?

In theory we could run the po4a scripts on macOS...

ignotus666:
That's strange - it only takes about 20 seconds for me to compile it (once the dependencies are installed). I couldn't find any binaries for 0.63 or later. It is possible to use it without compiling though using:

`PERLLIB=~/git-checkouts/po4a/lib ~/git-checkouts/po4a/po4a-<command>`. Replace "git-checkouts" with path to where po4a has been cloned. This has to be added to the lines in the scripts where the po4a-translate and po4a-updatepo commands are.

ignotus666:
I've created a .deb package with 0.64-alpha. I'm not at all experienced doing it but I just tested it on a clean Mint install and it seems to work ok.

[po4a_deb.tar.gz](https://github.com/jamulussoftware/jamuluswebsite/files/6687009/po4a_deb.tar.gz)

ann0see (approved):
Seems to be ok with your .deb file. Would like to have a review by someone else too.

PR #537, opened 2021-06-23
Additions and corrections to contribute docs

ignotus666:
# Does this need translation?

- [ ] Yes 
- [*] No 

# Changes

Corrected "adminstration" -> "administration".
Corrected spaces before units and some other errors
Added note about abbreviations

### DIFF
--- a/contribute/en/Administration.md
+++ b/contribute/en/Administration.md
@@ -1,8 +1,8 @@
 ---
 layout: wiki
-title: "Adminstration"
+title: "Administration"
 lang: "en"
-permalink: "/contribute/Adminstration"
+permalink: "/contribute/Administration"
 ---
 # Administration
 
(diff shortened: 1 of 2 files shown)


ignotus666:
I wanted to make a separate PR for the corrections to wiki/en docs against 'release', as I will also update the .po files to avoid turning those paragraphs into fuzzies. But I'm unsure whether this would create a conflict with #533 so I'll wait until that gets merged or discarded.

contribute/en/Administration.md
```diff
@@ -121,7 +121,7 @@ When a new Issue or Pull Request comes in, do the following:
 We use the Discord/Matrix chat for quicker, loose interaction.
 All decision and results should still be documented in permanent Github Issues, Discussions or Comments.
 
-## Decision making
+## Decision-making
```
pljones, line 124: No hyphen.

ignotus666: There are differing views on that but I was following the usage [here](https://dictionary.cambridge.org/dictionary/english/decision-making), [here](https://www.collinsdictionary.com/dictionary/english/decision-making) and [here](https://www.merriam-webster.com/dictionary/decision-making), where the noun form uses a hyphen.

pljones (approved):

ann0see:
Why doesn’t this go to release?

ignotus666:
Good question. For some reason I was convinced these documents had been added to next-release but didn't exist yet in release... I'll make a new PR.

contribute/en/Administration.md
pljones: Adjectival I'd agree with (_decision-making process_; the only example given in Collins, which is poor...) but straight noun form it makes me feel it's wrong...  Cambridge seems much more set on the hyphen...  Apparently it's because I'm old fashioned.  Which I guess is now "old-fashioned"...

Issue #525, opened 2021-06-10
Tracking outdated information 


ann0see:
This issue is to track outdated information on the website (as requested by @jujudusud)
 **This only tracks and should track the english version**
- [x] 1. Jamulus main window screenshot doesn't show the latest changes
- [x] 2. Main window screenshot is outdated
- [x] 3. Commands on the Choosing-a-Server-Type should be jamulus (lower case), not Jamulus (upper case)
- [x] 4. Client-Troubleshooting: Getting blasted by somebody joining your jam with feedback or noise? --> GUI has changed. We need to add that this is an advanced option
- [x] 5. Command line options (a lot of things are undocumented; we're already aware of that)
- [ ] 6. Jamulus Demos (move this to the KB?)
- [x] 7. FAQ: Diagram still uses old icons
- [x] 8. Hardware setup who still knows that this page exists ;-)
- [x] 9. Installation guides: The buttons at the bottom should be named Setup not Onboarding
- [x] 10. Installation for Linux: too short. We should at least hint at the compilation instructions or add a headline about "other distros"
- [x] 11. Running a private server, server win/mac/linux...: Should be included in an upcoming server guide
- [x] 12. Manual: Feedback Protection should be explained differently: Attempts to detect audio feedback loops or loud noise in the first three seconds after you connected to a server. Once detected, this feature will activate the "Mute Myself" button and mute you in your own mix. An error message will tell you that you should fix the problem (or similar)
- [x]  13. Manual: We're not consistent with systemctl vs service. The former is more modern.
- [x]  14. ~~We still use Central Server: https://jamulus.io/de/wiki/Server-Win-Mac#other-options~~ Upddated with SrvAdm manual in next release.

These are the issues I found from my first round through the website.

jujudusud:
Commenting for now only:
Why did jamulus should be lower case ?
Yes we should put "Jamulus Demos" in KB
FAQ Diagrams you can use FR screenshots images (with translation work in EN) (the public one needs to be identical to private in the client screen)

ann0see:
> Why did jamulus should be lower case ?

since the .deb file installs the binary as `jamulus` not `Jamulus`

jujudusud:
Ok, but it is not the standard build option. Anyway ... I'm ok with this explanation. Disapointed but ok ...
[...]

PR #542, opened 2021-06-27
Updating images + labels

gilgongo:
# Does this need translation?

- [] Yes 
- [x] No 

# Changes

New screenshots to reflect last release. Changing "Onboarding" to "Setup". As per https://github.com/jamulussoftware/jamuluswebsite/issues/525

Other languages are up to date.


ann0see:
wrong
Something is wrong here...

gilgongo:
That's weird. Is that the same for the other images I updated? Also, how do you get this dark mode to work? Is that a brower setting?

ann0see:
No. I don’t think so. Dark mode is a browser setting (which then applies specific CSS). In chrome you can set "prefers-color-scheme" manually.

gilgongo:
OK the png had some strange alpha channel setting on it. Can you try again?

ann0see:
Hmm. No. It still seems to be like that...

You can try to set dark mode via Chrome/-ium Dev Tools>Rendering> Emulate CSS media feature prefers-color-scheme>prefers-color-scheme: dark

gilgongo:
OK I may have uploaded the same file again by mistake using the web UI. I've tried again from the CLUI.

ignotus666:
Wouldn't it be better to separate out the image updates and make a PR for them to 'release'? That way they can go live now instead of waiting for the next release. ES (and I think FR) already have those images up to date.

gilgongo:
Oh I didn't check if they were up to date on the other languages. So we could merge this OK to release then.

ignotus666:
I was going to say that since it has to be translated, the Onboarding -> Setup change should go to 'next-release', but 'release' gets merged into 'next-release' once in a while, doesn't it? It'd make no difference for translations then.

gilgongo:
Ah yes - now that po4a is in the house we can merge next-release into release whenever we want, is that right?

ignotus666:
Yup, I can't think of a reason why not. Running po4a-update-templates.sh will sync the .po files to whatever state the EN wiki docs are in.

jujudusud:
Be careful. Remember that "next release" allows you to collect all the changes before a release and that it allows you to trace the changes made in EN before translating...
I don't think it's a good idea to make changes that don't have to go into the online site right away in "release" because it doesn't allow you to make a small change in the site.

gilgongo:
OK but in this case (for this PR) EN is actually more out of date than the other languages.  But I get what you mean about next-release.

BTW the checks seems to be taking ages in this...

jujudusud:
This PR is part of small changes that have to be made and pushed  to the online site right away. ;-)

ignotus666:
> allows you to collect all the changes before a release and that it allows you to trace the changes made in EN before translating

Is there a particular reason why we need to be able to trace all the changes made to EN docs in a single place (squash&merge?) ?

jujudusud:
> 
> Is there a particular reason ...

Only to be sure that all the changes needed in the next release are done before translating. I personally think that it is safer/easyier to do this in one branch and not multiple branches as it has been done for 3.8.0 unfortunatly.

ignotus666:
I think we need a separate branch from 'release' so that new/unfinished content doesn't go live straight away. AFAIK all content changes are being done to 'next-release'. 'Release' is only receiving corrections.

PR #546, opened 2021-06-29
preferred buffer delay 128 samples en + fr

### LINKED ISSUE #545: According to the website [software manual](https://jamulus.io/wiki/Software-Manual), 64 samples is the preferred setting to use for the Buffer.
According to the website [software manual](https://jamulus.io/wiki/Software-Manual), 64 samples is the preferred setting to use for the Buffer.
However the text in the GUI dialog indicates 128 samples is the preferred option.

Which is the correct one?

One of them should change....

_Originally posted by @henkdegroot in https://github.com/jamulussoftware/jamulus/discussions/1895_

jujudusud:
# Does this need translation?

- [x] Yes -- en, es and fr are ok. de, it, pt have to be updated.
- [ ] No 

# Changes

Modifiying the preferred buffer delay setting in website Software Manual file.


jujudusud:
I can add all other languages if translators sends the files (or the diff / patch).

ignotus666:
The .md versions of translated texts in wiki/ should **never** be edited. Translators should only ever modify .po files.

The correct procedure in this case would be to modify the .md file in English, run the template update script, edit the .po files for the translations and finally run the create targets script.

jujudusud:
@ignotus666 that's exactly what I did.
The script modify all the other languages. In order to push it online I only commit an push the .md files.
Do you want me to push also the .po ?

ignotus666:
Yes, otherwise the po files are out of sync.

PR #549, opened 2021-06-29
Software-Manual FR update for 3.8.1

jujudusud:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

Updated Software_Manual files of FR language for 3.8.1 release.

### DIFF
--- a/translator-files/po/fr/Software-Manual.po
+++ b/translator-files/po/fr/Software-Manual.po
@@ -5,7 +5,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: \n"
-"POT-Creation-Date: 2021-06-06 08:34+0200\n"
+"POT-Creation-Date: 2021-06-30 00:17+0200\n"
 "PO-Revision-Date: 2021-05-31 23:05+0200\n"
 "Last-Translator: daryl <darylo1@hotmail.com>\n"
 "Language-Team: English <kde-i18n-doc@kde.org>\n"
@@ -13,8 +13,10 @@ msgstr ""
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"Plural-Forms: nplurals=2; plural=(n > 1);\n"
 "X-Generator: Lokalize 19.12.3\n"
+"X-Language: fr_FR\n"
+"X-Source-Language: C\n"
 
 #. type: YAML Front Matter: lang
 #: wiki/en/Software-Manual.md:1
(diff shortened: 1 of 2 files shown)


translator-files/po/fr/Software-Manual.po
```diff
@@ -241,9 +247,16 @@ msgstr "## Effet réverbération"
...
+#| "Reverb can be applied to one local mono audio channel or to both channels "
+#| "in stereo mode.  The mono channel selection and the reverberation level "
+#| "can be modified. For example, if a microphone signal is fed in to the "
+#| "right audio channel of the sound card and a reverb effect needs to be "
+#| "applied, set the channel selector to the right and move the fader upwards "
+#| "until the desired reverb level is reached."
 msgid ""
-"Reverb can be applied to one local mono audio channel or to both channels in "
-"stereo mode.  The mono channel selection and the reverberation level can be "
+"Adds reverb to your local mono audio channel, or to both channels in stereo "
```
ann0see, line 258: Did this get produced automatically?

ignotus666: I'm assuming @jujudusud ran po4a-update-templates, which transferred the recent changes to Software-Manual.md to the .po files and updated them. The lines starting with #| are "previous untranslated strings" and will not show up in the final document.

ignotus666 (approved):
Looks ok to me. I see @jujudusud ran po4a-update-templates.sh and then selectively pushed only the files related to French (not saying that's bad - in fact it avoids a cascade of changes). OTOH I think it's preferable to wait until the translation period starts "officially" to avoid potentially wasting effort translating docs that might change again, but that's up to each person I suppose.

I do however think it might be time to consider automating the scripts with gh actions so that they get run "centrally" and no one has to worry about remembering to run them or not locally. Should I open an issue perhaps? There are a number of things to consider first, I think.

ann0see:
Ok. Feel free to open an issue. I assume these scripts can be run here: https://github.com/jamulussoftware/jamuluswebsite/blob/daf42e7d22f3da57e3bdb15cc028f8fc736c8bbb/.github/workflows/jekyll.yml#L20

and here: 

https://github.com/jamulussoftware/jamuluswebsite/blob/daf42e7d22f3da57e3bdb15cc028f8fc736c8bbb/.github/workflows/main.yml#L29

Are you ok with a merge of this PR?

ignotus666:
> Are you ok with a merge of this PR?

You mean this one? Yes, I think it's good to go.

> I assume these scripts can be run here:
> and here:

I made the scripts to run only when they actually need to, i.e. templates only get updated if docs in wiki/en/ are edited and targets only get created if files in po/** have been edited. These and other things are what I wanted to comment on - I'll open an issue.

PR #550, opened 2021-06-30
Add translation stats script

ignotus666:
# Does this need translation?

- [ ] Yes 
- [* ] No 

# Changes

Added a script that gets called from either po4a-update-templates.sh or po4a-create-all-targets.sh. It produces a 'statistics' text file in translator-files/ showing the updated translation status of all .po files for each language.

### DIFF
--- a/po4a-stats.sh
+++ b/po4a-stats.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+# You need gettext
+# You may set the following variables:
+# SRC_DIR folder for original source English .md files
+# PO_DIR directory where .po files are stored
+
+# Folder where source English .md files are
+#SRC_DIR="./wiki/en"
+
+# Directory where the po file folders are
+#if [ -z "$PO_DIR" ] ; then
+#	PO_DIR="./translator-files/po"
+#fi
+
+# Remove stats file before creating new one
+rm -f "translator-files/statistics"
+
+produce_stats () { 
+# Determine file names
+	while IFS= read -r -d '' file ; do
+		basename="$(basename -s .md "$file")"
+		
+		# Stats printed to translator-files/statistics.txt
+		echo "$lang"/"$basename".po >> translator-files/statistics
+		msgfmt --statistics "$PO_DIR/$lang/$basename".po &>> translator-files/statistics
+		echo '' >> translator-files/statistics
+		
+	done <   <(find -L "$SRC_DIR" -name "*.md"  -print0)
+	echo '' >> translator-files/statistics
+}
+
+# Run produce_stats on each language folder
+while IFS= read -r -d '' dir ; do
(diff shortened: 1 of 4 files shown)


po4a-stats.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+# You need gettext
```
ann0see: You should check for it ;-)

translator-files/statistics
```diff
@@ -0,0 +1,475 @@
+pt:
+pt/Compiling.po
```
ann0see, line 2: This or other status files in shouldn’t go into the repo (we don’t want it to be logged/created and committed all the time since it just shows a status; nothing directly essential to the page creation). In future we might want to create a note card on the tracking board which logs the status of translations.

po4a-stats.sh
ignotus666: Since this script is called by another that already checks, I don't see the need. I shouldn't have included that reference here.

translator-files/statistics
ignotus666: The idea was to offer an up-to-date overview of the status without having to load the files anywhere or run a script. If it can be shown elsewhere (and in a more visually appealing way) I'm all for it.

ann0see: Something in GH actions probably?

I wouldn’t like to have it in the repo (online) personally. However offline should be ok. 
We could add this file to a [.gitignore](https://www.atlassian.com/git/tutorials/saving-changes/gitignore)?

ignotus666: If the po4a scripts are run via gh actions, then this stats file gets produced as a side effect, nothing extra needs to be done. Sure, we can add it to gitignore.

po4a-stats.sh
ann0see: Ok. Then just remove this comment.

ignotus666: Done

translator-files/statistics
ignotus666: I was under the impression that including something in .gitignore meant that it wasn't tracked any more, but still "existed". But if it won't be visible on the repo, there's not really any point in having it - there are other ways to view the translation status offline. This was simply to offer an overview without having to clone the repo or do anything.

ann0see: > This was simply to offer an overview without having to clone the repo or do anything.

Yes that would be nice; however I don’t like the idea of having volatile files on the repo. I‘d rather not add it to the repo but focus on alternatives via GH Actions.

ignotus666: No problem - so we can close this then

ann0see: See https://github.com/jamulussoftware/jamuluswebsite/issues/554 and my comment on DC element. It’s only partly related but goes into a similar direction.

PR #556, opened 2021-07-03
Use LFS to save space

### LINKED ISSUE #554: Documentation repository might grow due to new image library
The new way of adding images results in all versions of images being saved by git. As time goes, the repo will get larger and larger (since the old images are still there.)

After thinking a bit more, I now remember why I created the assets repo: "assets" should contain all the binary data which might make a repo bigger and bigger. 

I‘d suggest moving the pictures over to the assets repo.

Once in my past with another git repo this issue resulted in gigabytes needed to be cloned.

ann0see:
You'll need to use git lfs now: https://git-lfs.github.com/

# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

THIS IS JUST A DRAFT FOR LFS


hoffie:
Hrm, is this really needed, i.e. do we hit any limits or performance problems?
I know LFS, but I would rather have expected to use it for something like larger binaries, e.g. ISOs, packages, etc.
I don't think screenshots and .po files should be that large?

Update: After having read #554 (which I should have done first), I understand the rationale. Still -- should we really cover the .po files as well?

ann0see:
I think it depends. .po files get changed very often and can always be recreated. 

On the other hand, they are just text files.

gilgongo:
Am I right in thinking that the most we can expect the images in the repo to grow is about 250Mb per year if we have about 20 images at 50k each, for 5 languages on about 5 branches and change all the images every time we release about 5 times per year? In practice it might be more like 50-100Mb though, so maybe we'd get to 1Gb in 10-20 years?

Just trying to size the problem.

ann0see:
Probably? Although you might think it’s not much, it‘s still a waste of disk space and makes git clone take forever (especially for lower bandwidth connections).

gilgongo:
Given that we don't edit images that much, would not moving them to the assets repo be a bit less complicated than migrating to LFS? I think having LFS installed locally also becomes a dependency of editing the website too. Does it work if you upload a new version of an image via github web?

EDIT: I see everyone would need to run `git lfs prune` as cache will not be cleaned up automatically. Also if someone without Git LFS installed commits a file which should be associated with LFS, what happens? Does the complexity warrant the benefit?

ann0see:
Yes, the assets repo solution should work too. Upload images -> link them on this repo.

gilgongo:
I think we should probably just add images to the assets repo then, because LFS seems a bit complicated/fragile.

ann0see:
Yeah. However, the ghuserimages way would also work. We could just open an issue and copy and paste all images in there?

gilgongo:
Do you mean to use ghuserimages in includes (which was my first thought orginally to solve the space problem)? That way, if an image is used more than once, we don't have to update it in multiple places. So the binary images in `/jamuluswebsite/assets/img/` would be replaced with include files that contained the URL to the actual image on ghuserimages. Not sure if that would work though as I've not tested it :-)

ann0see:
Yes. Something like that

gilgongo:
OK I'll see if I can explore that then.

hoffie:
Just for reference in case the LFS topic is every re-visited: We should be sure if and how much it would cost or how quickly we would reach the included free quota.
While certain paid features such as Github Actions are free for public repositories, I do not find a hint that this is the case for LFS. In addition, I'm seeing non-zero usage (although still small and within the free included limits) of LFS quota in the jamuluswebsite repo:

Screenshot 2022-01-24 at 11-05-10 Build software better, together
https://github.com/organizations/jamulussoftware/settings/billing

@ann0see Not sure if this is expected, maybe still some left-over to clean up?

https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-storage-and-bandwidth-usage
https://docs.github.com/en/billing/managing-billing-for-git-large-file-storage/about-billing-for-git-large-file-storage

ann0see:
No idea. Probably it’s a left over

hoffie:
I have not succeeded in finding out where those objects are, what they are called or how they could be removed.
There might be even a chance that LFS objects on forks of jamulus do count against our storage quota. Maybe you can re-check in your repo, plus, you have some more context from the past? @ann0see

gilgongo:
Do images in the issues/PR descriptions use LFS?

ann0see:
I think it’s not even enabled in my repo
E4344FB3-2A56-403A-BCA0-94877270AC65

ann0see:
https://www.jasongaylord.com/blog/2021/08/27/remove-git-lfs-from-repository

https://github.com/git-lfs/git-lfs/issues/3026

hoffie:
Not sure if "Include ... in archives" is the relevant part.

What I have found so far, it is impossible to explicitly remove Git LFS entries. The only way seems to be to drop all references and hope for Githubs GC to kick in at some point.

I suspect that this very PR is the reason that the references are still kept.
Should I try to get rid of it by changing the de81e951e1882fb602e635d3a5cfe90abe597e1d?

https://stackoverflow.com/questions/34579211/how-to-delete-a-file-tracked-by-git-lfs-and-release-the-storage-quota
https://docs.github.com/en/repositories/working-with-files/managing-large-files/removing-files-from-git-large-file-storage#git-lfs-objects-in-your-repository

ann0see:
Yes, you can try?

hoffie:
No success. :(

I had amended the commit from this PR to avoid looking like LFS via `sed -re 's/lfs/DISABLED-L-F-S/g' -i $(grep -ril 'lfs' * .gitattributes )`. The resulting commit obviously has a diferent id as it has different content.

I tried pushing to the virtual branch which one can use with `git fetch`, but it looks like this only created a new branch (which I have no cleaned up again).
The original branch no longer exists, so I don't think it makes sense to push there.

```
$ git push upstream pr556:pull/556/head
[...]
remote: Resolving deltas: 100% (2/2), completed with 2 local objects.
remote: 
remote: Create a pull request for 'pull/556/head' on GitHub by visiting:
remote:      https://github.com/jamulussoftware/jamuluswebsite/pull/new/pull/556/head
```

I'm giving up.

ann0see:
I've now deleted my repo. But it still doesn't seem to work...

So it's certainly not on my repo.

ann0see:
What happens if I delete this PR? Will try

ann0see:
Doesn't seem to be possible. Contacted GitHub support.

ann0see:
Ok. They removed the LFS files and they no longer show up in the org settings.

PR #560, opened 2021-07-07
Move non-translated EN docs and remove unneeded .po files

### LINKED ISSUE #559: Move EN files that don't need translation
There are a significant number of files in wiki/en that don't require translation. As po4a can't tell them apart from files that do, it creates .po files for them - in every language. This leads to unnecessary clutter and means translators need to know what to translate and what to leave be. I moved all the EN files that don't need translating to a separate "en-no-translate" folder in wiki/, removed all the unneeded .po files (45 of them...), and built the site locally to check whether moving the files had any negative effect - and it seems it doesn't.

This makes it simpler for translators ("just translate everything in your language po folder"), it speeds up the po4a scripts as they don't have to process as many files (it's a noticeable difference), and of course we get rid of clutter. 

Opening this issue in case anyone can think of a good reason _not_ to do this, or a better way.

Edit: I just had a look at the 'next-release' branch, and 14 out of 32 files in wiki/en/ don't need translation. Moving them out would mean 70 fewer .po files.

ignotus666:
# Does this need translation?

- [ ] Yes 
- [ *] No 

# Changes

As described in #559, docs that don't need translation have been moved to a 'en-no-translate' folder and all the corresponding unnecessary .po files (75) have been removed.

I also ran the po4a scripts so the current .po files and their counterparts in wiki/ are up to date and reflect the current state of translations.


ann0see:
Since it’s going to next-release it’s save to merge.

Issue #519, opened 2021-06-06
Exclude translator_files from being pushed to gh-pages


ann0see:
Currently the translator_files folder and both scripts are pushed to the gh-pages branch. Since they are not needed there, we could probably somehow exclude them.

ignotus666:
Put the scripts in a folder and prepend it (and translator-files) with an underscore? I tested that and it seems to work.

ann0see:
Argh yes this should work. I should have known that.

ignotus666:
Paths in the scripts have to be changed accordingly, too. [This](https://github.com/ignotus666/jamuluswebsite/tree/GH_actions_test2) branch has "excluded" folders for the scripts and po4a .deb package ("po4a-tools") and "translator-files". The paths in these scripts have been modified. Putting the scripts in a folder caused them to not work via GH actions (they did locally) due to how paths are established in actions, so the scripts have also been modified to work in that scenario.

Just documenting this info here for future reference. I don't trust myself not to forget about it.

PR #566, opened 2021-07-14
Exclude translator-files and scripts from gh-pages

ignotus666:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

The translator-files folder is excluded from gh-pages, as are the scripts, which have been moved to a folder '_po4a-tools'. The scripts have also been edited to work with the changed paths, and a couple of lines were added so they work in future with GH actions.

Addresses #519


_po4a-tools/po4a-update-templates.sh
```diff
@@ -46,15 +51,15 @@ fi
 
 while IFS= read -r -d '' file ; do
     # Determine target file/folder names
-    basename=$(basename -s .md "$file")
+    basename="$(basename -s .md "$file")"
```
ann0see: Not sure about the quoting here.

_po4a-tools/po4a-create-all-targets.sh
```diff
@@ -7,6 +7,11 @@
 # PUB_DIR directory to publish the localised files in
 # THRESHOLD translation % below which translated .md files are not generated
 
+# Get absolute path and CD to parent directory of script. Will work locally without it, but not via GH actions.
```
ann0see: This looks complicated. Where do you need to cd into?

../ ?

Maybe we can use GitHub actions variables?

_po4a-tools/po4a-update-templates.sh
ignotus666: I'm pretty sure I was asked to add them to the same expression in create-all-targets, (by @pljones perhaps? not sure), so added them here too. But I could also be mistaken. The scripts work though.

ann0see: They do probably work. Syntactically I‘m not sure if they’re correct.

_po4a-tools/po4a-create-all-targets.sh
ignotus666: This was because for some reason, when you put the scripts in a folder (i.e. not in the root directory), they will work locally -changing the paths in them of course - but not when launched by a GH action. I presume because for some reason the script "thinks" it's being run from another directory, but don't ask me why... These lines get the absolute path for wherever the scripts are (locally or in a GH action) and cd's into the folder holding them, in this case "_po4a-tools", and runs the scripts from there. But... I'm just a monkey hammering at a keyboard, if you know a better way, great.

ann0see: Hmm yes. I think it’s a problem with the path you’re in if you run the script. If you’re not in the folder where the script is saved/root directory/whatever, the relative paths are wrong. 

Probably it’s similar to https://unix.stackexchange.com/questions/295806/how-to-execute-a-script-located-in-the-same-directory-as-the-current-script
and not easily solvable.

ignotus666: Yes, when I was looking into it I found loads of examples, all using a similar command. This article explains what's going on, the title nicely condensing what I was struggling to explain: https://www.networkworld.com/article/2703490/unix---when-a-bash-script-asks--where-am-i--.html.

I've made a couple of changes, the variable name and the comment, hopefully making it a bit less cryptic.

_po4a-tools/po4a-update-templates.sh
ignotus666: So I went back to see if I'd been told to add those quotes, and it doesn't seem like it... so I removed them from both scripts.

ann0see (approved):

Issue #553, opened 2021-07-01
Run po4a scripts via gh actions


ignotus666:
As commented elsewhere, it would be good to automate running the po4a scripts via a gh action. I did some initial tests and created a couple of .yml files that can be found [here](https://github.com/ignotus666/jamuluswebsite/tree/GH_actions_test/.github/workflows) (targets.yml and templates.yml). What they do:

- po4a-update-templates.sh only gets triggered when edits are pushed to files in wiki/en/
- po4a-create-all-targets.sh only gets triggered when edits are pushed to files in translator-files/po/**

I did it that way because I think it's pointless to trigger the po4a scripts if anything else is edited.

po4a and its dependencies are cached to speed things up (vs installing every time). This cache gets purged if not used for 7 days or more, so if it can't be found, a new one is automatically created. The catch is that a po4a .deb package not available from any repos is required. I put one I created in the root folder but I assume it's not the best practice.

Hopefully some or all of the scripts can be used (a few things would need to be renamed in any event). As set up on that branch, they work. However, due to my very poor grasp of this gh action business, there are probably more sensible ways to go about it. This issue is to get the ball rolling and discuss the best approach.

ignotus666:
I managed to combine both .yml scripts into one. Depending on what files have been edited (en/*.md or po/**.po), the corresponding po4a script is run. See the script [here](https://github.com/ignotus666/jamuluswebsite/blob/GH_actions_test2/.github/workflows/po4a.yml). It's made possible using the [Paths Changes Filter](https://github.com/marketplace/actions/paths-changes-filter).

ann0see:
Added to my todo list; once I find time, I‘lll look at code and give some thoughts on this.

ann0see:
Ok. So your scripts basically install all dependencies, run the scripts and then push everything? 

The best thing which could happen would be to only create the .md files in the wiki folder right before Jekyll runs via CI. We’d 
 not even add them to the live repo. This removes the possibility to edit .md files on the live site accidentally entirely.

ignotus666:
> Ok. So your scripts basically install all dependencies, run the scripts and then push everything?

Yeah.

> The best thing which could happen would be to only create the .md files in the wiki folder right before Jekyll runs via CI. We’d
> not even add them to the live repo. This removes the possibility to edit .md files on the live site accidentally entirely.

So you mean run the create targets script as a step in the creation of the Jekyll site?

ann0see:
[...]

PR #568, opened 2021-07-20
Add new GH actions and stats script

### LINKED ISSUE #557: Show up to date translation status in repo
The idea was to offer an up-to-date overview of the status without having to load the files anywhere or run a script. If it can be shown elsewhere (and in a more visually appealing way) I'm all for it.

_Originally posted by @ignotus666 in https://github.com/jamulussoftware/jamuluswebsite/pull/550#discussion_r661704253_

ignotus666:
# Does this need translation?

- [ ] Yes 
- [* ] No 

# Changes

Here are the GH actions scripts that include the po4a processing. I'll keep this as a draft as a few things will need to be reviewed and changed. 

The main change apart from the scripts is that translated .md files are no longer in the repo - they are only produced just prior to building and zipping/deploying the site.

I also sneaked in the script to produce stats again. The stats information file is not in the repo either but could be viewed via a link, which hasn't been created yet.

For info about what the actions do now, see issue #553.


.github/workflows/jekyll.yml
```diff
@@ -1,20 +1,81 @@
...
+            templates:
+              - 'wiki/en/*.md'
+      - name: Update templates
+        if: steps.filter.outputs.templates == 'true'
+        run: ./_po4a-tools/po4a-update-templates.sh
+# This step only runs if a PR is merged:
+      - name: Push changes to repo if action triggered on:push and templates need updating
+        if: ${{ github.event_name == 'push' && steps.filter.outputs.templates == 'true' }}
+        run: |
+            git config --local user.name "ignotus666"
```
ignotus666: This will have to change

.github/workflows/main.yml
```diff
@@ -1,26 +1,64 @@
...
+        with:
+          base: ${{ github.ref }}
+          filters: |
+            templates:
+              - 'wiki/en/*.md'
+      - name: Update templates
+        if: steps.filter.outputs.templates == 'true'
+        run: |
+          ./_po4a-tools/po4a-update-templates.sh
+          git config --local user.name "ignotus666"
```
ignotus666: Same here

_po4a-tools/po4a-create-all-targets.sh
```diff
@@ -9,15 +9,14 @@
...
-
 cd "$SCRIPT_DIR"
 
 ####################################
 # INITIALISE VARIABLES
 ####################################
 
 # Set % threshold for translated .md files to be created
-THRESHOLD="80"
+THRESHOLD="0"
```
ignotus666: This has been set to 0 because if higher, some untranslated documents (from newer EN .md files) prevent the site from showing in that language. It's because of some dependency between .md files.

_po4a-tools/po4a-stats.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# Stats file location
+STATS_FILE="../wiki/Statistics.md"
+
+# Print yaml front matter and table title/header
+echo '---
+layout: wiki
+title: "Statistics"
+lang: "en"
+permalink: "/wiki/Statistics"
```
ignotus666, line 12: I added this despite not being sure where this file can be made available.

.github/workflows/jekyll.yml
ann0see: Yes. And we should probably use another way to commit changes. There should be an official way to add commits by a GH actions step.

ignotus666: Done.

ignotus666:
To add a new language, it's possible to do it by creating a folder in /po with the language code, putting an empty temporary file inside (as you can't create empty folders, e.g. `temp.po~`), and then using as the title for the PR: "Add new language". This triggers the script in the action that creates all the .po files for the new language and adds them to the PR. The temp file is deleted automatically. I tested it and it works nicely.

I was looking at what would need to be changed in the readmes and realised adding a new language was missing in the actions. What do you think?

ann0see:
Will have a look at this once I find some time.

ignotus666:
sync.yml can't run parallel to main.yml as the latter may sometimes push commits that need to be included in sync.yml. What about including sync as a job in main.yml (using `needs:`)?

ann0see:
Yes. We’re also using needs: on the autobuild on the main repo.

.github/workflows/jekyll.yml
```diff
@@ -1,20 +1,76 @@
...
-
+    branches: [ next-release, release ]
+  push:
+    branches: [ next-release ]
 jobs:
-  build:
+  Jekyll_site_CI:
     permissions:
-      contents: read
+      contents: write
```
ann0see, line 10: We‘ll need to add new GH actions to the repo whitelist.

.github/workflows/jekyll.yml
```diff
@@ -1,20 +1,76 @@
...
-    branches: [ changes, release, translate ]
   pull_request:
-    branches: [ changes, release, translate ]
-
+    branches: [ next-release, release ]
+  push:
+    branches: [ next-release ]
 jobs:
-  build:
+  Jekyll_site_CI:
```
ann0see:
```suggestion
  jekyll_site_ci:
```

Personal preference.

.github/workflows/jekyll.yml
```diff
@@ -1,20 +1,76 @@
...
+        id: cache-po4a
+        with:
+          path: "~/po4a"
+          key: ${{ runner.os }}-po4a
+      - name: Install or retrieve po4a from cache
+        env:
+          CACHE_HIT: ${{steps.cache-po4a.outputs.cache-hit}}
+        run: |
+          if [[ "$CACHE_HIT" == 'true' ]]; then
+            sudo cp --force --recursive ~/po4a/* /
```
ann0see: Why do you need —force and sudo here?

.github/workflows/jekyll.yml
```diff
@@ -1,20 +1,76 @@
...
+          else
+            sudo apt install -yq gettext libsgmls-perl libyaml-tiny-perl opensp
```
ann0see: So these are the dependencies?

.github/workflows/jekyll.yml
```diff
@@ -1,20 +1,76 @@
...
+            wget -O po4a.deb https://github.com/jamulussoftware/assets/raw/main/po4a/po4a_0.64-alpha.deb
+            sudo dpkg -i po4a.deb
+            mkdir -p ~/po4a
+            for dep in po4a libcroco3 libosp5 sgml-base gettext libsgmls-perl libyaml-tiny-perl opensp; do
+                dpkg -L $dep | while IFS= read -r f; do if test -f $f; then echo $f; fi; done | xargs cp --parents --target-directory ~/po4a/
```
ann0see: Could you please explain what and why you’re doing this ? Do you check if all dependencies were installed correctly? Why is this needed?

.github/workflows/jekyll.yml
```diff
@@ -1,20 +1,76 @@
...
+        if: steps.filter.outputs.templates == 'true'
+        run: ./_po4a-tools/po4a-update-templates.sh
+        
+# This step only runs if a PR is merged:
+      - name: Push changes to repo if action triggered on:push and templates need updating
+        if: ${{ github.event_name == 'push' && steps.filter.outputs.templates == 'true' }}
+        uses: EndBug/add-and-commit@v7
+        with:
+          default_author: github_actions
+          message: 'Templates updated'
```
ann0see:
```suggestion
          message: 'AUTO: Update templates'
```

.github/workflows/jekyll.yml
```diff
@@ -1,20 +1,76 @@
...
+# Create translated .md files. Never pushed to the repo.
+      - name: Create translated .md docs and stats
+        run: ./_po4a-tools/po4a-create-all-targets.sh
+
+# Build, zip and upload site
+      - name: Build the site in the jekyll/builder container
+        run: |
+          docker run \
+          -v ${{ github.workspace }}:/srv/jekyll -v ${{ github.workspace }}/_site:/srv/jekyll/_site \
+          jekyll/builder:latest /bin/bash -c "chmod a+w /srv/jekyll/Gemfile.lock && chmod 777 /srv/jekyll && jekyll build --future"
```
ann0see: I know it was a copy and paste from a GH sample I added, but chmod 777 is probably not right. Less permissions are probably possible.

.github/workflows/main.yml
```diff
@@ -1,26 +1,64 @@
...
-      # It is highly recommended that you only run this action on push to a
-        # specific branch, eg. master or source (if on *.github.io repo)
-
 jobs:
-  jekyll:
+  build_and_deploy_site:
     permissions:
       contents: write
-    runs-on: ubuntu-16.04 # can change this to ubuntu-latest if you prefer
+    runs-on: ubuntu-latest
```
ann0see, line 10: Did that work for you?

_po4a-tools/po4a-stats.sh
ann0see: Yes. Adding this file to the site is probably better.

.github/workflows/jekyll.yml
ignotus666: I suppose --force is not strictly necessary if using sudo, but sudo **is** necessary to copy the files into usr/bin, usr/lib, etc. I can remove --force, though I don't see any harm in leaving it either.

.github/workflows/jekyll.yml
ignotus666: Yes. But some of these in turn have their own dependencies, which leads to the next comment.

.github/workflows/jekyll.yml
ignotus666: If `"$CACHE_HIT" == 'false'`, this creates a list of all the dependencies installed by the previous commands, required for a po4a installation, and copies them all to the cache folder (`~/po4a/`). This cache folder is what is stored by and retrieved later from git if `"$CACHE_HIT" == 'true'`.

.github/workflows/main.yml
ignotus666: It works in that the workflow runs successfully. I can't see whether the site is deployed properly I suppose because I'd have to change some configuration files to deploy it on my repo.

In fact if I tried to use ubuntu-16-04 I got errors saying it was deprecated.

.github/workflows/jekyll.yml
ignotus666: Tell me what you'd change it to.

_po4a-tools/po4a-stats.sh
ignotus666: Here's the thing though: the version displayed on the site would be the translation status of 'release'. What about if a translator wants to know about the status in 'next-release'? The stats file gets created for both branches, but the one for 'next-release' gets "lost" as it isn't published anywhere.

ignotus666:
> To add a new language, it's possible to do it by creating a folder in /po with the language code, putting an empty temporary file inside (as you can't create empty folders, e.g. `temp.po~`), and then using as the title for the PR: "Add new language". This triggers the script in the action that creates all the .po files for the new language and adds them to the PR. The temp file is deleted automatically. I tested it and it works nicely.
> 
> I was looking at what would need to be changed in the readmes and realised adding a new language was missing in the actions. What do you think?

Any thoughts on this?

ignotus666:
> Yes. We’re also using needs: on the autobuild on the main repo.

I was asking also because if we include it as a job in main.yml you lose the `workflow_dispatch` option - but I don't know if that's really important.

_po4a-tools/po4a-stats.sh
ann0see: Hmm yes. If we succeed to have an issue which gets updated updated automatically, this would also work. 

https://github.com/marketplace/actions/create-or-update-comment

better: https://stackoverflow.com/questions/66756072/updating-github-issues-from-github-actions to avoid needing to add unknown code.

ann0see:
> Any thoughts on this?

Yes, that’s ok. Not perfect since you‘d commit an unneeded file, but ok. What about checking the main Jekyll.yaml file for new languages?

.github/workflows/main.yml
ann0see: Hmm. For me ubuntu-latest have an error.

.github/workflows/jekyll.yml
ann0see: 700 or 770?

Or r,w,x for owner rx for group and others?

ignotus666: 755?

ignotus666:
>What about checking the main Jekyll.yaml file for new languages?

Do you mean checking `languages:` in _config.yml? It'd be cool that if a new language were added to the array, it'd trigger the creation of the .po files for it, but that's way beyond my deficient skills.

ann0see:
Yes, that would be great. 

Either you do some regex/grep/sed/… magic (not exactly sure about the regular expression) or something from this post via python: https://stackoverflow.com/a/15113891 

Proposal:

1. get languages array from _main.yml
2. For each array item, check if folder exists in _translator-files/… if false: create it, if true: 
 create new po files

.github/workflows/jekyll.yml
ann0see: Yes

ignotus666: Done

ignotus666:
> Proposal:
> 
> 1. get languages array from _main.yml
> 2. For each array item, check if folder exists in _translator-files/… if false: create it, if true:
>    create new po files

I had a look into this and as I suspected, it's beyond my abilities. But realistically speaking, how many times would this function be used, from here until the end of time - 5? 6? I'd love to be proven wrong but I think that's even being optimistic. So to get this moving I propose chucking in the "add folder + file" hack for now and if someone can implement a more elegant solution in the future, then all the better.

.github/workflows/main.yml
ignotus666: Yep, just tested it again and the workflow fails with ubuntu-16.04. Po4a won't work properly and the jekyll build fails.

.github/workflows/jekyll.yml
ann0see: Thanks

ann0see:
Yes, that’s ok.

_po4a-tools/po4a-create-all-targets.sh
ignotus666: Changed it back to 80 since 'release' is not affected by the .md file dependencies yet.

_po4a-tools/po4a-stats.sh
ignotus666: This is also beyond my capabilities. The stats file for 'next-release' will be in the zipped site anyway, so not entirely lost...

ignotus666:
I've added the stuff for new languages and updated the READMEs to document that and to reflect the changes brought by implementing the new actions. I also moved some of the information about translating the website as I think it was a bit scattered.

These actions have been tested to death on my repo, but the one thing I couldn't view was the actual website deployment - the workflow runs without errors, but still...

.github/workflows/jekyll.yml
ann0see: Yup. Please remove --force if it's not needed.

.github/workflows/jekyll.yml
ann0see: Ok. Leaving this open for other reviewers. Maybe you should add a comment in the code here.

.github/workflows/jekyll.yml
```diff
@@ -1,20 +1,76 @@
...
+# Paths changes filter. If there's been a change to any files defined in the filter, the step (update templates) runs:
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          base: ${{ github.ref }}
+          filters: |
+            templates:
+              - 'wiki/en/*.md'
+      - name: Update templates
+        if: ${{ steps.filter.outputs.templates == 'true' || github.event.pull_request.title == 'Add new language' }}
```
ann0see: You shouldn't only run this if a PR is named "Add new language". Everybody could trigger it (not only people with write/maintenance access), and what happens if someone labels his PR  "Add chinese translation"?

I think it should be more generic: Maybe check for a specific PR label ("new language") or similar.

.github/workflows/main.yml
```diff
@@ -1,26 +1,64 @@
...
+        id: cache-po4a
+        with:
+          path: "~/po4a"
+          key: ${{ runner.os }}-po4a
+      - name: Install or retrieve po4a from cache
+        env:
+          CACHE_HIT: ${{steps.cache-po4a.outputs.cache-hit}}
+        run: |
+          if [[ "$CACHE_HIT" == 'true' ]]; then
+            sudo cp --force --recursive ~/po4a/* /
```
ann0see: --force: remove if possible.

.github/workflows/main.yml
```diff
@@ -1,26 +1,64 @@
...
+          else
+            sudo apt install -yq gettext libsgmls-perl libyaml-tiny-perl opensp
+            wget -O po4a.deb https://github.com/jamulussoftware/assets/raw/main/po4a/po4a_0.64-alpha.deb
+            sudo dpkg -i po4a.deb
+            mkdir -p ~/po4a
+            for dep in po4a libcroco3 libosp5 sgml-base gettext libsgmls-perl libyaml-tiny-perl opensp; do
+                dpkg -L $dep | while IFS= read -r f; do if test -f $f; then echo $f; fi; done | xargs cp --parents --target-directory ~/po4a/
+            done
```
ann0see: Comments missing (as above)

_po4a-tools/po4a-stats.sh
ann0see: Ok.

ann0see:
>  but the one thing I couldn't view was the actual website deployment

You could change the links in a separate branch and enable GH pages on your repo.

ann0see:
@pljones I know, you're not too much into the website stuff, but could you please have a short look at the scripts?

.github/workflows/jekyll.yml
ignotus666: I'm not sure I understand. Do you want to make it easier to trigger this or more restrictive? As it is, anyone can trigger it by creating the folder + empty file and the PR title "Add new language". It's documented in the README in _translator-files/ so people know exactly how to do it. As commented elsewhere, this is going to be a pretty rare event, and anyone wanting to do it just has to be directed to the relevant README. If you have to instruct them to add a specific label instead of a PR title, you're not really simplifying it, are you? It's just a different instruction.

That said, I have no problem changing how this is done - I simply need to understand better what the goal is.

.github/workflows/main.yml
ignotus666: Done (also in jekyll.yml).

.github/workflows/main.yml
ignotus666: Done in both files.

ignotus666:
> You could change the links in a separate branch and enable GH pages on your repo.

I gave it a go but keep getting a 404 error when trying to open the website. I'm missing something but am running out of willpower today...

.github/workflows/jekyll.yml
ann0see: Goal: no need to have a fixed PR title
Solution: don’t check for a title, but for a fixed label (which e.g. maintainers would add).
Open question: do we have security issues/bad behaviour if a PR author (=untrustworthy) opens a PR which triggers adding a new language

I hope it’s clearer now

.github/workflows/main.yml
ann0see: Thanks

ann0see:
No problem ;-). It‘s already late. 

BTW: thank you for all the work you’ve spent on po4a!

.github/workflows/jekyll.yml
ignotus666: Ok - so you'd like to make it more restrictive, i.e. it gets triggered when a maintainer adds a label? Yes, I agree that there's the risk a cheeky person could open a PR, but... that's the case at present, for any public repo. The new language files get added to their PR, they're not automatically added to our repo, so if it's an unwanted/untrusted PR you just have to reject it.

ignotus666:
> BTW: thank you for all the work you’ve spent on po4a!

I just hope to god it's worth it ;-)

.github/workflows/jekyll.yml
ignotus666: Ok, so adding a new language triggered by adding a label (e.g. "new language") works too, limiting the scope of who can trigger the action. Either way though, I think the approach is flawed if the PR author has to enable actions for their repo (since adding a language pushes new files to their repo)... But only triggering it on push won't work either because empty folders containing empty files don't get pushed from one branch to another (while they can be pushed from a local branch), so another mechanism would have to be devised...

 Also, as an aside, zipping the site every time jekyll.yml is run means we'd be ending up with a whole load of zipped files being uploaded. Adding a language would cause 3 uploads (PR, add label, merge). 

It's back to the drawing board I think - suggestions are welcome!

ann0see: > Either way though, I think the approach is flawed if the PR author has to enable actions for their repo (since adding a language pushes new files to their repo)... But only triggering it on push won't work either because empty folders containing empty files don't get pushed from one branch to another (while they can be pushed from a local branch), so another mechanism would have to be devised...

Good points. Didn't think about the issue of needing to enable actions. Actions do also run on our side and can push to the repo of the new user, however this might not be wanted.

What about checking for a git tag (but this will then only work for the main repo)?

new-lang-[language name] --> Creates folders and files --> pushes them to the repo --> now a person forks the repo --> translation work --> open PR --> merge new language?

ignotus666: Hmm - I was hoping for something -a couple of lines- that could be added to the workflow to automate adding a new language because at the end of the day, doing it manually is so trivial that I think coming up with convoluted workarounds to do something that will seldom be needed is a bit of a waste of everyone's time... In the time I've taken just to think about how to do this I could have added 200 more languages =)

Another idea: dorny/paths-filter@v2 has an interesting filter that can be triggered when you add/modify/delete files in a specified path. I set it to trigger update-templates if files are added (not modified or deleted) to _translator-files/po/ and it works nicely. It can be set to do this on:push so it only creates the .po files on our repo when the PR is merged and doesn't require activating actions by the PR author. So:

- User creates empty lang folder + temp file and makes PR. We could also do this for them, but whatever.
- When PR is merged, the language files are created and pushed to the main repo.
- User pulls update/forks from main repo and translates.

The nice thing is it's only 3 lines added to the script.

ann0see: Hmm. Yes, I think your suggestion is the easiest. Let’s just do that.

ignotus666: Done. The procedure is as explained above. No special PR titles or anything - the .po files get created and added to the main repo when merging a PR submitted with a new language folder containing a dummy temp.po~ file.

ignotus666: BTW, I intentionally didn't include the "add new language" function in main.yml as I think languages should only be added to 'next-release'.

Edit: I was mistaken when I said
>"But only triggering it on push won't work either because empty folders containing empty files don't get pushed from one branch to another

They do get pushed and it does work. And calling a folder empty when it contains a file (even if the file is empty itself) is not correct ;). Will change that in the readme.

ignotus666: > What about checking for a git tag (but this will then only work for the main repo)?
> 
> new-lang-[language name] --> Creates folders and files --> pushes them to the repo --> now a person forks the repo --> translation work --> open PR --> merge new language?

I had a go at this:
- User creates an issue with the new language in square brackets in the title (e.g. "Add Basque [eu]")
- Adding a "new language" label triggers a workflow that extracts the language from the square brackets, creates the folder/files and pushes them to the 'next-release' branch (not to 'release').
- Ready for the user to fork and translate the new language.

I think it's basically what you were proposing - I thought it would be harder than it actually was. I did do it in a separate workflow script though, and I omitted the site zipping and uploading, but left the building part as a check.

The only thing that doesn't seem to work is retrieving the cached dependencies - it says "issues" event types are not supported. It still runs fine though, just takes a little longer.

So there's that option too. I do think it's a much neater solution than what I have now. So if you OK it I'll add it to this PR and edit the README instructions accordingly.

You can check out the script [here](https://github.com/ignotus666/jamuluswebsite/blob/release/.github/workflows/add-lang.yml).

ann0see: Yes, this sounds good. You just need to make sure that the string in brackets doesn’t open a security hole (you need to validate it).

ignotus666: Uh... ok (just looked up what that means). So should I check if it matches a two, lower case-letter string or something like that?

Edit:
This: test for a language code that's either in the 'xx' or 'xx_XX' format? The latter in case someone adds e.g. 'pt_BR'.

ann0see: Exactly. 
en_US,en, should be allowed, but not ../

You can probably just modify the regular expression in the sed command.

ignotus666: Done. Restricted to 'xx' and 'xx_XX' formats - anything else will cause it to exit.

ignotus666: Added the script to this PR.

.github/workflows/add-lang.yml
```diff
@@ -0,0 +1,73 @@
...
+        id: cache-po4a
+        with:
+          path: "~/po4a"
+          key: ${{ runner.os }}-po4a
+      - name: Install or retrieve po4a from cache
+        env:
+          CACHE_HIT: ${{ steps.cache-po4a.outputs.cache-hit }}
+
+# If CACHE_HIT: true, retrieve the cache. If not, install po4a and its dependencies and copy them all to a folder (~/po4a/) to be cached:
+        run: |
```
ann0see: Since we’re using this three times now, would it be possible to have it in a shell script?

.github/workflows/add-lang.yml
```diff
@@ -0,0 +1,73 @@
...
+            done
+          fi
+
+# Get language from issue title and create files for it:
+      - name: Make new lang directory and create .po files
+        env:
+          ISSUE_TITLE: ${{ github.event.issue.title }}
+        run: |
+          NEW_LANG=$(sed 's/.*\[\([^]]*\)].*/\1/' <<< "$ISSUE_TITLE")
+          if [[ $NEW_LANG  =~ ^[a-z]{2}(_[A-Z]{2})?$ ]]; then
```
ann0see: I think we could also allow en_us —> also allow lower case letters with the regex. But maybe that’s inconsistent?

ann0see (commented):
Just a comment.

.github/workflows/add-lang.yml
ignotus666: Yes, that makes sense. Done.

.github/workflows/add-lang.yml
ignotus666: Hmm, the xx_XX format is what is used in the application and I'd rather stick to that for consistency. Same for underscore vs hyphen.

ann0see:
Let's do the following: Squash and merge this PR, test on jamulus.io and if anything goes completely wrong, I'll revert it.

ignotus666:
Fingers crossed!

ann0see:
It's taking much longer to build now...

ignotus666:
It never took that long in my tests

ann0see:
Just re-started it.

ann0see:
Ok. The site seems to be working, but there are merge conflicts between release and next-release.

ann0see:
I think the url to the statistics site should be /contribute/statistics?

ignotus666:
> but there are merge conflicts between release and next-release.

I think that is to be expected - when release and next-release have the same directory structure and files (in en-no-translate, /wiki, etc) I think it should go well. It did in my tests anyway.

> I think the url to the statistics site should be /contribute/statistics?

Since it's a file that's never actually stored anywhere I didn't think it mattered - but I can change the path if you want.

ann0see:
/wiki/ is usually for the stuff users see. /contribute/ is for the stuff maintainers/contributors see/need to know about.

BTW, I think chmod 777 is needed (we're getting an error for jekyll building)

ann0see:
Ok. Fixed that.

ignotus666:
Great. I would do what is necessary to make 'next-release' ready to become 'release' (remove unneeded.po files, remove translated language folders from wiki/, etc.), deploy it, and make 'next-release' a copy of that. Then we could do a call for translations and start afresh from there. Otherwise I think we're just going to keep chasing conflicts whenever something changes in 'release'. It will mean some existing translations will fall a bit behind, but it's not that much and we could bring them back up to date with a test round of translations.

ann0see:
Yes. Do you really need to do that much on next release? I think next-release is almost releasable? Tagging
@gilgongo here. 

What we want to do is a dry run of the release process, I think?

ann0see:
Ah. Ok. Just saw the files on next-release. 

It’s not releasable because of the shared-commands issue: https://github.com/jamulussoftware/jamuluswebsite/pull/573/checks?check_run_id=3214409686

gilgongo:
We could I suppose put that content into the files and just put comments around it to remind anyone updating it that they need to do so in both places? Well, actually three places if you include the CLI.

ignotus666:
Lowering the threshold to 0 (even if temporarily) will solve the issue. That, or making sure that the linked files are always translated together.

ignotus666:
It's also possible to only apply a 0 threshold to specific files (shared-commands and software-manual in this case) and 80 to the rest by adding this in po4a-create-all-targets.sh:

```
if [[ $basename == 'Software-Manual' || $basename == 'Shared-Commands' ]] ; then
    THRESHOLD="0"
else
    THRESHOLD="80"
fi
```
But I don't know if that's too messy. Works though.

Edit:
What about adding something to the names of .md files that share content so a wildcard can be used in the script? E.g. "Shared-Commands-linked.md" and "Software-Manual-linked.md" so:

```
THRESH_VAL="80" # To keep the variable at the top of the file

if [[ "$basename" == *"-linked" ]] ; then
    THRESHOLD="0"
else
    THRESHOLD="$THRESH_VAL"
fi
```
Edit 2:
Overthinking this by far... we only need Shared-Commands.md to have a 0 threshold so `if [[ "$basename" == "Shared-Commands" ]]` etc. would do.

Aaaand edit 3:
What about this: leave it as `if [[ "$basename" == 'Shared-'* ]]` and if any files are added in future with the same function, name them starting with "Shared-" too.

PR #565, opened 2021-07-14
Sync release to changes

ann0see:
#558 was merged and resulted in conflicts. Since the po4a scripts were run (on release and next-release), .po files are at different stages and the conflicts need to be resolved.

This PR should fix https://github.com/jamulussoftware/jamuluswebsite/actions/runs/1030148493 which failed due to conflicts.

PR #570, opened 2021-07-26
Sync release to changes #565 

ann0see:
Tries to fix conflicts. See #565


ann0see:
Getting this error:

Liquid Exception: Could not locate the included file 'Shared-Commands.md' in any of ["/wiki/fr"]. Ensure it exists in one of those directories and, if it is a symlink, does not point outside your site source. in wiki/fr/Software-Manual.md

ann0see:
getting a 404 for french server administration page...

In fact, the french translation seems to be gone?!

ignotus666:
Don't panic - this is because 'Shared-Commands.md' hasn't been translated into French, therefore doesn't pass the 80% threshold, and because other files that **have** been translated depend on it, the deployment of that language fails. It's why I dropped the threshold to 0 in #568. After giving it some thought, I think it wouldn't do any harm to drop the threshold parameter entirely (or leave it at 0), since they way po4a works outdated information will always get filtered out anyway. But that's another discussion.

ann0see:
After a bit of research through the files, I think that's the problem. Can we find a solution for Shared-Commands which does not break the site (even if we set the threshold)?

ignotus666:
Let's ping @gilgongo about that because I think he's the one who wrote those docs.

In short: if a doc that includes 'Shared-Commands' has been translated, but 'Shared-Commands' hasn't, that language fails to deploy entirely.

jujudusud:
> Don't panic - this is because 'Shared-Commands.md' hasn't been translated into French, ...

Do you need a fr translation for this file ?

ignotus666:
Hmm - this PR is going to merge into 'release' a ton of updates to EN files that haven't been translated yet. I understand if it's necessary to resolve the conflicts mess, but just pointing it out.

gilgongo:
If  'Shared-Commands.md' hasn't been translated into French - can it be? 

I didn't want to have to maintain exactly the same info in two seperate places.

ignotus666:
@ann0see I just did a test and figured out how to merge 'release' into 'next-release':

- Copy/replace the _translator-files/po folder from release to next-release.
- Rename 'translator-files' to '_translator-files' in next-release.
- Copy/overwrite 'Installation-for-Macintosh' and 'Installation-for-WIndows' from wiki/es/ in release to the same folder in next-release.

It should now let you merge without conflicts.

The next step is to delete all the unneeded .po files again from next-release and running po4a-update-templates.sh. I also tested this and there are no conflicts merging next-release into release after that.

ann0see:
Isn’t that what this PR also tries to do?

ignotus666:
No, you seem to be merging next-release into release (the other way round).

ann0see:
Oh, thank you!!!

Was probably not paying enough attention here. Should be fixed now?

ignotus666:
The list of changes is just too enormous to check... what changes did you make, and do we know if this fixes the conflicts for 'release' -> 'next-release'?

ann0see:
Just
 git checkout next-release
 git merge release

and fixed the conflicts in the .po files (as mentioned in the other PR). 

This moves the (translator) files around and does not update the .po files.

ignotus666:
Oh, so this is as if you'd already merged release into next-release then?

ann0see:
Locally. Yes. And this PR will do it online

ignotus666:
Great, then go ahead. It also means the Shared-Commands drama isn't so urgent because it isn't going live.

ann0see:
Ok. Merged

ignotus666:
I think these conflicts when merging release -> next-release are going to keep happening as long as there is the directory structure and .po file mismatch between them. In my tests it doesn't happen when they are aligned in those aspects, but when I replicate the current status of those two branches conflicts keep happening.

ann0see:
Hmm. That’s not good, but we need to live with it

PR #577, opened 2021-08-03
Prepare next-release branch

ignotus666:
# Does this need translation?

- [ ] Yes 
- [ *] No 

# Changes

This PR makes 'next-release' ready to become 'release' and be deployed.
It also includes a workaround so that any file beginning with 'Shared-' (shared between documents) does not need to pass the 80% threshold for publication, thus avoiding problems caused when these files are not yet translated.


_translator-files/po/de/QOS-Windows.po
```diff
@@ -51,25 +51,24 @@ msgid ""
...
+"&nbsp;Local Computer Policy<br> &nbsp;&nbsp;Computer Configuration<br> &nbsp;"
+"&nbsp;&nbsp;Windows Settings<br> &nbsp;&nbsp;&nbsp;&nbsp;Policy-based QoS "
+"(click)<br> Looking at the third panel (Actions) of the Local Group Policy "
+"Editor<br> &nbsp;Policy-based QoS<br> &nbsp;&nbsp;More Actions<br> &nbsp;"
+"&nbsp;&nbsp;Create new Policy (click)<br> &nbsp;&nbsp;&nbsp;&nbsp;Policy "
+"Name: Jamulus<br> &nbsp;&nbsp;&nbsp;&nbsp;Specify DSCP value: 32<br> &nbsp;"
+"&nbsp;&nbsp;&nbsp;Next<br> &nbsp;&nbsp;&nbsp;&nbsp;This QoS policy applies "
+"Only to applications with name Jamulus.exe<br> &nbsp;&nbsp;&nbsp;&nbsp;"
+"Next<br> &nbsp;&nbsp;&nbsp;&nbsp;Next<br> &nbsp;&nbsp;&nbsp;&nbsp;UDP<br> "
+"&nbsp;&nbsp;&nbsp;&nbsp;Finish<br>"
```
ann0see, line 68: Worth another issue. Maybe we should use some pictures here.

ann0see (commented):
Just scanned it.

ann0see:
Moving this on, I think we should have a look a the release process for the website and do a dry-run?

_translator-files/po/de/QOS-Windows.po
ignotus666: That section is a real pain to translate...

ignotus666:
Yeah, I think so.

ann0see:
Just opened #579

PR #585, opened 2021-08-06
Remove text wrap, ES translation complete, delete po4a.deb

ignotus666:
# Does this need translation?

- [ ] Yes 
- [ *] No 

# Changes

- Got rid of the text wrap in .po files.
- ES translation fully up to date.
- The po4a.deb file had sneaked in due to an oversight. Should get deleted now after it's retrieved by the GH action.
- FR, DE and PT translations should be as they were, ready to be translated/updated.
- Edited add-lang.yml so the title format check happens _before_ installing all the po4a stuff. 

Will check the zipped site now to make sure everything is OK and will report here.


ignotus666:
Everything seems to be in order.

ann0see:
Thanks. Could you please split ES translation and the other work into different PRs for the next time? This makes reviewing a lot easier.

.github/workflows/add-lang.yml
```diff
@@ -9,12 +9,29 @@ jobs:
...
+          ISSUE_TITLE: ${{ github.event.issue.title }}
+          ISSUE_NUM: ${{ github.event.issue.number }}
+        run: |
+          NEW_LANG=$(sed 's/.*\[\([^]]*\)].*/\1/' <<< "$ISSUE_TITLE")
+          if [[ $NEW_LANG =~ ^[a-z]{2}(_[A-Z]{2})?$ ]]; then
+            echo "new_lang=$NEW_LANG" >> $GITHUB_ENV
+            echo "issue_num=$ISSUE_NUM" >> $GITHUB_ENV
+          else
+            echo Language code in wrong format
+            exit 1
```
ann0see, line 25: Good idea to have it here. I think we should have a look at all these scripts from a security PoV. I think @hoffie knows some more about bash & security?

ann0see (approved):

ignotus666:
> Could you please split ES translation and the other work into different PRs for the next time? This makes reviewing a lot easier.

Yeah sorry about that, got a bit carried away... in my defence, the only part requiring a review are the scripts, and they were at the top ;)

PR #589, opened 2021-08-09
Optimise GH action 

ignotus666:
# Does this need translation?

- [ ] Yes 
- [ *] No 

# Changes

**NOTE**: this PR is superseded by #590 - it can be merged before that one, or ignored and merge #590 directly.

Changed jekyll.yml script so it just builds, zips and uploads the site on pull request, and does not run at all if not necessary (when merging changes other than those to EN .md files).

Also updated some variable names and comments so they are clearer to understand.

Issue #521, opened 2021-06-07
Timestamp changes in .po and .pot files pollute commits


ignotus666:
When the po4a-update-templates.sh script is run, the timestamp changes in all .po and .pot files. GH interprets this as a modification and lists all those files (almost 200) as having been modified, when this modification is inconsequential. This means an endless list of edits is shown, polluting the commit history and making it hard to pick out relevant changes. There must be a way to tell GH to ignore changes to this line (it's just one) but alas my skills are lacking.

ignotus666:
Using sed to delete the timestamp line in the header right after the .po files are processed by `po4a-update-templates.sh`, this problem goes away. I was under the mistaken impression that po4a needed the timestamps for some purpose, but it turns out it's a different kind of timestamp that can be optionally generated - and is not applicable here. Did some testing and didn't notice any negative consequences.

PR #590, opened 2021-08-13
Wrapping and timestamp fixes, updated READMEs

ignotus666:
# Does this need translation?

- [ ] Yes 
- [ *] No 

# Changes

**NOTE**: THIS SHOULD ONLY BE MERGED AFTER #589 IS - OR DISCARD THAT AND DIRECTLY MERGE THIS ONE.

I divided them into 2 PRs to make reviewing easier, but whatever - this one includes all the other PR's changes plus some more.

- Corrected the no-wrap command as the previous one was not working as expected.
- Added command to remove the timestamp line starting with "POT-Creation-Date:" in .po files. This mostly solves the issue of commits being flooded with meaningless edits of this line. Before, this line changed in every single .po file, every time `po4a-update-templates.sh` was run - even if no .po files had been edited. Now, though some editors (namely Poedit) add it back in, it will only happen to the files that are actually edited via translation, vastly reducing the problem. Addresses #521.
- As a consequence of the first point, the list of recommended editors has been reduced to 2 (Poedit and OmegaT), as the rest mess up the line wrapping.
- Made the po4a installation error messages a bit clearer.
- Edited the README and OmegaT instruction files in `_translator-files` to update the instructions regarding the preferred editors.

Leaving this here as I'll be gone until September and will likely not be available in the meantime. The wrapping and timestamp issues were seriously bugging me and I wanted to get them out of the way before leaving.

### DIFF
--- a/contribute/en/po4a-Setup-Explained.md
+++ b/contribute/en/po4a-Setup-Explained.md
@@ -15,7 +15,7 @@ First though, a brief overview.
 
 In broad terms, there are three types of activity that take place on the website repository:
 
-- Editing/adding/updating translatable documentation in English, the source language that all translations are based on. Translators now need to know _where_ these changes have taken place, preferably without having to actually scan the documents in English or trawling through PRs.
+- Editing/adding/updating translatable documentation in English, the source language that all translations are based on. Translators now need to know _where_ these changes have taken place, preferably without having to actually scan the documents in English or trawl through PRs.
 - Translating these documents into each respective language. As translations now take place in special .po files, these need to be converted back into .md format for the website.
 - Editing other content in the repository that is not necessarily the documentation itself (scripts, config files, images, etc.). This activity is largely unaffected by the new setup.
 
--- a/contribute/en/po4a-Setup-Explained.md
+++ b/contribute/en/po4a-Setup-Explained.md
@@ -96,7 +96,7 @@ This is the more complex script. A distinction between 'pull request' and 'push'
 
     5- Translated .md files are created with `po4a-create-all-targets.sh`.
 
-    6- Site is built, zipped and uploaded as an artifact. This is so the contributor can download and view the site locally, to check translations as they would appear on the website or the overall appearance of the website if other content has been edited (config files, images, etc.) before the PR is merged. Only happens if the trigger was a PR.
+    6- Site is built, zipped and uploaded as an artifact. This is so the contributor can download, unzip twice and view the site locally (e.g. with a [local webserver](https://pythonbasics.org/webserver/)), to check translations as they would appear on the website or the overall appearance of the website if other content has been edited (config files, images, etc.) before the PR is merged. Only happens if the trigger was a PR.
 
 - **Push events**. Only applicable to 'next-release':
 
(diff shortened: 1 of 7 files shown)


contribute/en/po4a-Setup-Explained.md
```diff
@@ -0,0 +1,117 @@
...
+This document is a description of the po4a-based processes that are behind the website translation and documentation maintenance workflow. It mainly focuses on the scripts automating these processes, describing the function of each one of them in an attempt to provide an understanding of how they all fit together.
+
+First though, a brief overview.
+
+## Contributor activities
+
+In broad terms, there are three types of activity that take place on the website repository:
+
+- Editing/adding/updating translatable documentation in English, the source language that all translations are based on. Translators now need to know _where_ these changes have taken place, preferably without having to actually scan the documents in English or trawl through PRs.
+- Translating these documents into each respective language. As translations now take place in special .po files, these need to be converted back into .md format for the website.
```
ann0see:
```suggestion
- Translating these documents into each respective language. As translations take place in special .po files, these need to be converted back into .md format for the website.
```

contribute/en/po4a-Setup-Explained.md
```diff
@@ -0,0 +1,117 @@
...
+- Editing other content in the repository that is not necessarily the documentation itself (scripts, config files, images, etc.). This activity is largely unaffected by the new setup.
```
ann0see: Do you really want to refer to "new setup"? This implies that we need to remove the reference to "new" someday…

I‘d mainly see this file as an explanation of the process which can be edited/adapted in future and be linked to for new contributors/maintainers/translators/… who need to know something about the po4a process.

contribute/en/po4a-Setup-Explained.md
```diff
@@ -0,0 +1,117 @@
...
+
+    2- Checks po4a cache (as above).
+
+    3- If the above filter is 'true', update/add .po files with `po4a-update-templates.sh`. If not, skip.
+
+    4- The output from the previous step is never pushed (even if it is 'true') when the trigger is a PR, as this would mean pushing the changes to the contributor's repository, which in turn would require them to enable GitHub actions beforehand. As this cannot be assumed to always be the case, it is left until their PR is merged and then the output is pushed to the Jamulus repository.
+
+    5- Translated .md files are created with `po4a-create-all-targets.sh`.
+
+    6- Site is built, zipped and uploaded as an artifact. This is so the contributor can download and view the site locally, to check translations as they would appear on the website or the overall appearance of the website if other content has been edited (config files, images, etc.) before the PR is merged. Only happens if the trigger was a PR.
```
ann0see: Maybe give a hint that the user should extract the .zip file and run it either with a local web server or jekyll serve. Just double-clicking on index.html doesn’t work.

ann0see (approved):
Wow! Thank you very much - especially for the detailed description of the po4a process!
Approving it, although I‘d like to have a review from someone else - especially concerning the deletion of the update date in the po files.

contribute/en/po4a-Setup-Explained.md
ignotus666: I see what you mean. I don't think I'll have much access to computers until September, so feel free to edit stuff. If you want, ping me, I'll have a look and commit the edits myself; whatever.

As for the sed command to delete the timestamp line, I know it's a bit hacky, but I would ask that any objections offer an alternative. It does prevent some 85 lines from being needlessly edited every time the script runs. I thoroughly tested it and there were no negative consequences if that's what you're concerned about.

ann0see (approved):

PR #595, opened 2021-08-29
Update 2021-01-05-Jamulus-Sound-Devices.md

TheDoomGoblin:
Making sure that the Zoom L-8 is configured to 48kHz to avoid Jamulus ASIO Driver error message

# Does this need translation?

- [ ] Yes 
- [X] No 

# Changes

Additional information regarding Zoom L-8 settings needed to avoid Jamulus ASIO Driver error message


ann0see (changes requested):
Great! Thank you. I think you shouldn’t add it as Windows only, but as

**Note:** under all OS. Probably it’s a device setting on the mixer and nothing you need to set via your PC?

ann0see (approved):
Thank you!

ann0see:
Thanks for working on that. You should be able to see the changes in about 3-5 min!

PR #596, opened 2021-09-05
Document IPv6 support

ann0see:
# Does this need translation?

- [x] Yes 
- [ ] No 

# Changes

Add Statement to server manual

### DIFF
--- a/wiki/en/Running-a-Server.md
+++ b/wiki/en/Running-a-Server.md
@@ -52,7 +52,7 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
 
 - Running a Private Server at home will require you to [port forward](#running-a-private-server) on your router. When running a Private Server, port forwarding should not be necessary in most cases, but it's advisable to do so because some networks may not work properly with Jamulus in its default mode.
 
-- Jamulus only has limited IPv6 support which needs to be enabled with a command line argument on the client and server. There are plans to expand IPv6 support for connections started from the server list.
+- Jamulus only has limited IPv6 support which needs to be enabled with a command line argument on the client and server. There are plans to expand IPv6 support.
 
 
 # Server Types
--- a/wiki/en/Running-a-Server.md
+++ b/wiki/en/Running-a-Server.md
@@ -52,7 +52,7 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
 
 - Running a Private Server at home will require you to [port forward](#running-a-private-server) on your router. When running a Private Server, port forwarding should not be necessary in most cases, but it's advisable to do so because some networks may not work properly with Jamulus in its default mode.
 
-- Jamulus only has limited IPv6 support which needs to be enabled with a command line argument on the client and server. There are plans to expand IPv6 support.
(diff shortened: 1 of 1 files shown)


wiki/en/Running-a-Server.md
```diff
@@ -52,7 +52,7 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
 
 - Running a Private Server at home will require you to [port forward](#running-a-private-server) on your router. When running a Private Server, port forwarding should not be necessary in most cases, but it's advisable to do so because some networks may not work properly with Jamulus in its default mode.
 
-- Jamulus doesn't currently support IPv6
+- Jamulus only has limited IPv6 support which needs to be enabled with a command line argument on the client and server. There are plans to expand IPv6 support.
```
ann0see, line 55: Not sure in how much detail we want to go here

pljones: I think what's there looks okay.  At this stage, anyone wanting to use the option needs to know how to go read the command line options page without being told exactly what to do.

gilgongo: I've not been following the IPv6 ticket very closely, but is it worth saying "Jamulus only has limited IPv6 support _for private servers only_, which ..." - or is that not helpful?

pljones: "private" and "public" and "central" got deprecate, so "no" :)

It could have "Server registration is currently not supported." added.

ann0see:
```suggestion
- Jamulus only has limited IPv6 support which needs to be enabled with a command line argument on the client and server. There are plans to expand IPv6 support for connections started from the server list.
```

wiki/en/Running-a-Server.md
```diff
@@ -52,7 +52,7 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
 
 - Running a Private Server at home will require you to [port forward](#running-a-private-server) on your router. When running a Private Server, port forwarding should not be necessary in most cases, but it's advisable to do so because some networks may not work properly with Jamulus in its default mode.
 
-- Jamulus doesn't currently support IPv6
+- Jamulus only has limited IPv6 support which needs to be enabled with a command line argument on the client and server. There are plans to expand IPv6 support for connections started from the server list.
```
pljones: There's more to it than that.  I'd rather not over-simplify things just for the sake of more words.

ann0see: Yes, it’s more than that. 
@gilgongo should we revert my proposal then?

gilgongo: OK sure, it's only really a half-feature at the moment and we don't want to make a big thing about it yet.

ann0see:
Should be squash and merged.

wiki/en/Running-a-Server.md
pljones: `Jamulus only has limited IPv6 support which needs to be enabled with a command line argument on the client and server.` alone, would do for now.  _Maybe_ say "Directory servers are not supported" (which would cover server registration with a directory server over IPv6 and running a directory server on IPv6).

pljones (approved):

PR #600, opened 2021-09-07
Draft: Preliminary work on JSON-RPC docs

dtinth:
# Work in progress

Corresponding Jamulus PR: https://github.com/jamulussoftware/jamulus/pull/1975

# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

### DIFF
--- a/wiki/en/JSON-RPC.md
+++ b/wiki/en/JSON-RPC.md
@@ -49,6 +49,25 @@ Jamulus will also send **notifications** to the consumer:
 
 <!-- prettier-ignore-end -->
 
+---
+
+> 🤔 Should I list all the methods in the documentation or should I just do this…?
+
+## Available methods
+
+To see what methods are available, take a look at these files:
+
+- [`clientrpc.cpp`](https://github.com/dtinth/jamulus/blob/json-rpc/src/clientrpc.cpp)
+- [`serverrpc.cpp`](https://github.com/dtinth/jamulus/blob/json-rpc/src/serverrpc.cpp)
+
+For requests from the consumer to Jamulus, search for `pRpcServer->HandleMethod`.
+
+For notifications from Jamulus to the consumer, search for `pRpcServer->BroadcastNotification`.
+
+---
+
+> 🤔 Here’s a draft list of all methods. But keeping them up to date may be a pain.
+
 ## Common methods
 
 ### `jamulus/getMode`
--- a/wiki/en/JSON-RPC.md
+++ b/wiki/en/JSON-RPC.md
@@ -1,0 +1,202 @@
+---
+layout: wiki
+title: "JSON-RPC interface"
+lang: "en"
+permalink: "/wiki/JSON-RPC"
+---
+
+# JSON-RPC interface
+
+A JSON-RPC interface is available for both Jamulus client and server to allow programmatic access. To use the JSON-RPC interface, run Jamulus with `--jsonrpcport <port>` option. This will start a JSON-RPC server on the specified port on the localhost.
+
+## Wire protocol
+
+The JSON-RPC server is based on the [JSON-RPC 2.0](https://www.jsonrpc.org/specification) protocol, using [streaming newline-delimited JSON over TCP](https://clue.engineering/2018/introducing-reactphp-ndjson) as the transport. There are three main types of messages being exchanged:
+
+- A **request** from the consumer to Jamulus.
+- A **response** from Jamulus to the consumer.
+- A **notification** from Jamulus to the consumer.
+
+## Example
+
+After opening a TCP connection to the JSON-RPC server, the following **request** can be sent:
+
+<!-- prettier-ignore-start -->
+
+```json
+{"id":1,"jsonrpc":"2.0","method":"jamulus/getMode","params":{}}
+```
+
+<!-- prettier-ignore-end -->
+
+The request must be sent as a single line of JSON-encoded data, followed by a newline character. Jamulus will send back a **response** in the same manner:
+
+<!-- prettier-ignore-start -->
+
+```json
+{"id":1,"jsonrpc":"2.0","result":{"mode":"client"}}
+```
+
+<!-- prettier-ignore-end -->
+
+Jamulus will also send **notifications** to the consumer:
+
+<!-- prettier-ignore-start -->
+
+```json
+{"jsonrpc":"2.0","method":"jamulusclient/chatTextReceived","params":{"text":"<font color=\"mediumblue\">(01:23:45 AM) <b>user</b></font> test"}}
+```
+
+<!-- prettier-ignore-end -->
+
+## Common methods
+
+### `jamulus/getMode`
+
+Returns the current Jamulus mode.
+
+**Parameters:** None (empty object)
+
+**Returns:**
+
+| Member | Type   | Description                                         |
+| ------ | ------ | --------------------------------------------------- |
+| mode   | string | The current Jamulus mode (`"client"` or `"server"`) |
+
+## Client methods
+
+### `jamulusclient/getClientInfo`
+
+Returns the status of the client.
+
+**Parameters:** None (empty object)
+
+**Returns:**
+
+| Member    | Type    | Description                                    |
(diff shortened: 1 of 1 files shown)


wiki/en/JSON-RPC.md
```diff
@@ -0,0 +1,221 @@
+---
+layout: wiki
+title: "JSON-RPC interface"
+lang: "en"
+permalink: "/wiki/JSON-RPC"
+---
+
+# JSON-RPC interface
+
```
gilgongo, line 9: As I assume this page will grow in time (and is already quite extensive), you may want to add a table of contents.

```suggestion
 {:.no_toc}

This manual documents the JSON-RPC-based protocol by which you can control the Jamulus client and server.

<details markdown="1">

<summary>Table of contents</summary>

* TOC
 {:toc}

</details> 
```

ignotus666:
Are we going to translate this? It's very technical, and I'd assume anyone using it would be familiar with the terminology in English, but I'll go with the consensus. If we're not, it should go in `wiki/en-no-translate/`.

gilgongo:
I was assuming we would (I'd actually checked the box on behalf of @dtinth), but if not then the adopted convention is that we'd not put it not on the website but in an .md file in the main repo I think? I didn't know about `wiki/en-no-translate/` though!

ignotus666:
The reason I raise the issue of translation is that it contains _a lot_ of vocabulary that is very technical and specific, which I'm not entirely sure has translations that are commonly used in different languages. I'm not a programmer, but as English is effectively the _lingua franca_ for programming, translations heavy in technical terms may even be harder to understand than the original.

Yes, `wiki/en-no-translate/` is where files that don't get translated go, so the .po folders don't get cluttered with unnecessary files. Links to them work exactly the same as to the files in `wiki/en/` (i.e., you don't need to specifiy the `en-no-translate` path).

gilgongo:
OK in that case we can just put it in `wiki/en-no-translate/` and it can be on the website.

ignotus666:
It should be moved there before merging though, otherwise it's going to create a bunch of unwanted .po files. Tagging @dtinth

If other @jamulussoftware/translators want to chime in though that would be great.

gilgongo:
Oh BTW OT but just seen this line in `_config.yml`

`exclude_from_localization:   ["assets", "images", "css", "README.md", "manifest.json", "humans.txt", "sitemap.xml", "robots.txt", "404.html", "CNAME", "LICENSE.md"]`

Is there a reason `wiki/en-no-translate`/ isn't listed there?

(Also notice that the `site` variable seems to have been removed - anyone know why?)

ignotus666:
I suppose `wiki/en-no-translate/` could go in there, yes. As 'site' gets prepended with an underscore (_site) it doesn't get included in the site at all, so needn't be included in the list. That's my guess anyway.

dtinth:
@ignotus666 

> The reason I raise the issue of translation is that it contains _a lot_ of vocabulary that is very technical and specific, which I'm not entirely sure has translations that are commonly used in different languages. I'm not a programmer, but as English is effectively the _lingua franca_ for programming, translations heavy in technical terms may even be harder to understand than the original.

I totally agree... I will move this to `wiki/en-no-translate`.

wiki/en/JSON-RPC.md
dtinth: @gilgongo I am thinking about whether we need an extensive documentation for JSON-RPC at all…

My concern is that this would double the work required to make changes to the JSON-RPC interface. First you would need to make changes in `jamulus` project, and then you would need to document the method here in `jamuluswebsite`.

I think that adds substantial maintenance burden. I want to avoid having to make changes to 2 repositories and filing 2 PRs. So, upon second thought, I am thinking of doing either of these:

- Point the reader to read `clientrpc.cpp` and `serverrpc.cpp`. Since the code is quite self-documenting already, it may be a redundant to have an English documentation at all.

- Add some code comments directly in `clientrpc.cpp` and `serverrpc.cpp` and create a Python script to generate a Markdown documentation based on these comments (e.g. `tools/generate_rpc_docs.py`). The documentation would live directly in jamulus repository, e.g. `RPC.md`, and jamuluswebsite would just link there. Maybe some GitHub Actions could be created to ensure that this documentation is up-to-date. This would also lets reader view the documentation of a specific release.

  The comment might look like this:

  ```cpp
  /* For notifications */
        /// @rpc_notification jamulusclient/clientListReceived
        /// @brief Emitted when the client list is received.
        /// @param {array} params.clients - The client list.
        /// @param {number} params.clients[*].id - The channel ID.
        /// @param {string} params.clients[*].name - The musician’s name.
        /// @param {string} params.clients[*].skillLevel - The musician’s skill level (beginner, intermediate, expert, or null).
        /// @param {number} params.clients[*].countryId - The musician’s country ID (see QLocale).
        /// @param {string} params.clients[*].city - The musician’s city.
        /// @param {number} params.clients[*].instrumentId - The musician’s instrument ID (see CInstPictures::GetTable).

  /* For methods */
        /// @rpc_method jamulusclient/setSkillLevel
        /// @brief Sets your skill level.
        /// @param {string} params.skillLevel - The new skill level (beginner, intermediate, expert, or null).
        /// @result {string} result - Always "ok".

  ```

Suggestions or opinions?

gilgongo: I'm not qualified to say, but if it's self-documenting, then less work is always better/safer.

ann0see:
Going to re-open this. @dtinth @gilgongo I raised the question if this should go to the KB or not

wiki/en/JSON-RPC.md
ann0see: Could we add a more extensive documentation to the file (md table like here and then link the comment from here)? I don’t think people should be forced to read the code. But they can of course read comments - although that’s not optimal (consistency, style) but ok for an experimental feature.

ann0see:
Suggestion:
- Add a short document to the wiki linking to the jsonrpc code file
- Add a well formatted documenting comment in the code
- Link the comment from the website

gilgongo:
See my comment on https://github.com/jamulussoftware/jamulus/pull/1975#issuecomment-1041198431 - I'm note sure I underand how this PR relates to that one.

wiki/en/JSON-RPC.md
hoffie: We should also consider the audience of this documentation: It will be developers (not necessarily Jamulus developers or C++ developers). They should already be familiar with reading code or at least comments in the code. I like the idea of having a single place to handle both the code and the directly related docs.

I think we should still keep the first part in the "wiki" (everything including the **Example** paragraph?), put the method documentation right next to the code and link from the wiki page to the code.

dtinth:
@ann0see @gilgongo @hoffie My original plan was to put the documentation in `jamulus` repo itself, e.g. in `JSON-RPC.md`, so that its documentation evolves alongside the codebase (similar to COMPILING.md) — that's why this PR was previously closed. I would rather not have to submit PRs to 2 different projects to update them. What do you think?

gilgongo:
> @ann0see @gilgongo @hoffie My original plan was to put the documentation in `jamulus` repo itself, e.g. in `JSON-RPC.md`, so that its documentation evolves alongside the codebase (similar to COMPILING.md)

Yes, I think the "rule of thumb" is that if the information is aimed at **developers** messing with "code", then it goes into the repo (like you say, as COMPILING.md is). If it's to do with the "end usage" of Jamulus (so in this case what the effect of turning on the RPC option is, and some basic info on how to use it), then it's on the website (AKA "the wiki", although I'm not sure why we call it that).

hoffie:
Yes, sounds good to me. End-user-relevant things (CLI option, pointer to deeper docs) here in `jamuluswebsite`, deeper docs right next to (or even -- in the form of comments -- in) the code.

ann0see:
Ok. So next steps:
1. Add a short entry to the website "we support JSON-RPC and please see the following comment in the code for available options" and CLI documentation
2. Document all methods in the code.

ann0see:
Closing

Issue #586, opened 2021-08-06
Limit jekyll site zip uploads to PR


ignotus666:
I was wondering whether it might make sense to only zip and upload the site on pull request - to make it available to download and check - and just build the site as a check (no zipping and uploading) when the same PR is merged. Currently it's uploading 2 basically identical artifacts for every PR/merge.

ann0see:
Yes. That would make a lot of sense. If the code is exactly the same, there’s no need to run the check twice. 
BTW: https://github.com/jamulussoftware/jamulus/pull/1921#issuecomment-890499912 is related

ignotus666:
Shall I leave out only the site zipping and uploading or the build too? I suppose both - we could bundle them in a separate job with `needs:` and only trigger that job if on: pull_request.

Edit: no, a separate job won't work as it needs to build and zip the output of the previous steps, which haven't been merged anywhere... Adding `if: ${{ github.event_name == 'pull_request' }}` to the last three steps (build, zip, upload) would do it.

I'm not confident enough about my knowledge to poke my head into that thread, but couldn't you just make a list of the files you _don't_ want to trigger the workflow with, using a negative pattern in paths? E.g.: 

```
paths:
  - '!CONTRIBUTING.md'
  - '!README.md'
  - etc.
```

ann0see:
Yes:

```
- '!LICENSE.md'
- '!README.md'
- '!.gitignore'
- '!.nojekyll'
- '!.remarkrc'
- '!.spellchecker.json'
- '!CONTRIBUTING.md'
```

…but with paths-ignore: and without the !.

So the build would not run if these files were changed. However, Peter disliked my proposal in the mentioned PR. I just linked it for your reference

ann0see:
> Shall I leave out only the site zipping and uploading or the build too?

Avoid as much as possible ;-).

ignotus666:
> Avoid as much as possible ;-).

Will do.

> I just linked it for your reference

Yeah I know, hence why I wasn't going to showcase my ignorance there :)

ignotus666:
Looking at how to optimise which steps are run and when, I scribbled down a table to visualise it better. I can't be bothered to make a pretty one in markdown so a photo will have to do:

![IMG_20210809_172430](https://user-images.githubusercontent.com/4263412/128731798-16025f3b-6076-49fe-8443-448dae407084.jpg)

[...]

PR #601, opened 2021-09-08
Only create targets on PR

ignotus666:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

The script was still unnecessarily creating the target .md files on merges to 'next-release'. See table in #586.


gilgongo:
Should this be tagged for next release or can it go in now?

ignotus666:
It can go in now. It gets synced to 'next-release' so both branches will have it.

PR #603, opened 2021-09-08
Remove Demos to KB

gilgongo:
# Does this need translation?

- [x] Yes 
- [] No 

# Changes

As per No.6: https://github.com/jamulussoftware/jamuluswebsite/issues/525

I removed the link to Demos on the home page for now too.


ignotus666:
If Demos.md isn't going to be translated, the corresponding Demos.po files should also be removed from `_translator-files/po/LANGS/`.

gilgongo:
Oh ok.

PR #624, opened 2021-09-23
Move to MIT licence

gilgongo:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

As per [discussion here](https://github.com/jamulussoftware/jamulus/discussions/1872) and on Discord, we don't have any GPL code in the repo, and all scripts etc. are our own work. We prefer MIT over GPL so let's make it so.

BTW this of course applies to the REPO not the CONTENT of the website, which is CC-BY-SA

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Jamulus
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.


ann0see (approved):

gilgongo:
OK so @pljones has raised a good point: 

> If [the website is] how we maintain the primary support site for the software, it MUST be as safe from abuse as the software.
> 
> If someone forks Jamulus, and wants to fork the support site too, then they need to know they'll be free to continue to take changes from upstream on their fork. MIT would raise the potential that that could stop happening. jamulus.io could go stale (for whatever reason), the main support site could move elsewhere and changes might stop being made public in source form.  If that then, in turn, went stale, it would be very problematic to support users going forward. 

He advocates the GPL to protect us from this scenario. I understand that @ann0see would prefer MIT due to its permissiveness though. 

So it comes down to what we're afraid of:

If we just want to disclaim any liability for what's in the repo ("I ran `jekyll --serve` on a clone of your site and it erased my hard disk!!") but let anyone do whatever they want with it, then MIT is the way to go (or possibly Apache2 if we are afraid of patent trolls ("We hereby demand a fee for your use of the method `codey code` which we, Evil Corp, have the patent rights for.").

If we want to avoid the scenario that @pljones mentions, then GPLv2 or v3 would do. As would I think the AGPL, which may be better for a website repo (cf. gnu.org: _"[the AGPL]...allow[s] users who interact with the licensed software over a network to receive the source for that program. We recommend that people consider using the GNU AGPL for any software which will commonly be run over a network."_)

Setting as draft for now, but now that I've thought more about it, my own opinion is that we don't really care if anyone is put off by the restrictions of the GPL (we're not trying to get people to use the website repo). But we **would** care if Jamulus users lost support materials, however unlikely that may be.

So I'd favour the AGPL.

ann0see:
I‘m still in favor of something permissive. What’s about Apache licenses?

gilgongo:
@ann0see Apache is basically the same as MIT but more specifically worded and with a protection against patent trolls. 

As @pljones says, Apache and MIT would allow monetising a fork of the repo, including the build process, with no requirement to contribute back any changes to the code used to build the site (e.g. to add in the ad-rotators and money-walled "enhanced content", etc). It's this lack of "contributing back", not the possible monetisation, that's the issue.

But even if we don't mind about contributing back, if we have a more permissive licence then we have to accept that Jamulus might lose control of its support materials to a commercial entity one day (once we're gone, that is). It's a "long game" thing - but needs to be decided now since migration later (from MIT to AGPL, say) might not be possible. 

EDIT: Also note that the current CC-BY-SA licence for the content is in line with what the AGPL would give us for the repo. I see no need to change that.

ann0see:
Adding my thoughts here:

The main problem with permissive licenses would be the ability of people to fork the repo and make their fork (especially their own work) proprietary? This doesn’t mean that the main site would become proprietary too, would it?

Do we see the ability of someone adding proprietary changes as an issue? 

Personally, I‘d say as long as the FLOSS part we wrote is still healthy, available, and a proprietary version doesn’t cause any harm, it can still co-exist. No need to force someone in open sourcing their work if they don’t want to. If Jamulus.io gets stale the problem could be the lack of free website updates? There’s still the ability to have a free fork based on the stale site?

As far as I know, the GPL just requires giving users who get the "program" access to modified code there’s no need to actively raise PRs or similar; however everyone can take the modified code and integrate it into the software. This would not be the case for MIT/Apache licensed code which could have been put under another license. 

Does a license exist which allows developers to write proprietary parts (better additions) as long as they don’t touch the floss part?

If we decide on the GPL, I agree that we should probably choose the AGPL

gilgongo:
> Does a license exist which allows developers to write proprietary parts (better additions) as long as they don’t touch the floss part?

Sounds like the LGPL maybe? 

https://en.m.wikipedia.org/wiki/GNU_Lesser_General_Public_License

pljones:
> No need to force someone in open sourcing their work if they don’t want to. 

If it's independent, that's quite true.  If it depends on someone else's work, then why shouldn't they be required to contribute as freely as the original contributors - i.e. on an equal basis?

gene96817:
The perspective I am familiar with would allow anyone to monetize open-source code. However, any derivative work should be stay open-sourced. This means if they make any improvements or bug fixes it stays with the project. This means someone could write some new (incompatible) feature that would still belong in the open-source library.

Why would we want a permissive license that would let someone build new features that are proprietary. It is already possible to have a fork with unique features. If someone has an interesting new idea built on Jamulus, that derivative work should belong to the community. After all, it is an idea inspired by Jamulus. It builds on the work of the Jamulus community.  The work cannot exist without the work of the Jamulus community. (That is why it is a derivative work.)

ignotus666:
> Does a license exist which allows developers to write proprietary parts (better additions) as long as they don’t touch the floss part?

This is the key issue for me. If someone wants to make separate additions proprietary, I could accept that. But if they could modify our work and then close it all off, I wouldn't be ok with it. Some of my own scripts derive from AGPL'd work and although I've substantially modified them (though it's still debatable whether they could be considered to be entirely separate) I'd be uncomfortable if subsequent derivations could be made proprietary.

gene96817:
Determining "don't touch the floss part" is tricky. We should keep to the usual definition where the license covers "derivative works". Even that is not guaranteed to be foolproof. It should be left to the inventor of new functions to solve the question of whether their work falls under derivative work or not. @ignotus666 if the new function is truly not related to Jamulus functions (therefore not a derivative work) the current license should be good enough. We should not (as amateur lawyers) create exemptions that could undercut the license schemes. A lot of really smart people with experienced IP lawyers have worked on the different license schemes. I think the current license is a very good one to build on all the efforts of the Jamulus developer community.

ignotus666:
I basically mean that if someone were to modify any of the existing content in any way, I think those files should remain open source. If they were to add additional scripts or whatever, even if it were to enhance the current setup (therefore derivative), I wouldn't be that bothered. But I think we're talking about extremely unlikely scenarios here in any event. I initially stated elsewhere that I couldn't really give a toss, but after further thought, if only on principle, I'd vote for GPL or AGPL. Ok, that might restrict a hypothetical future developer's options but hey... To put it bluntly, in practice I don't think it would mean pissing in anyone's cornflakes.

gilgongo:
@ignotus666 

> if only on principle, I'd vote for GPL or AGPL.

Not LGPL? From what I can tell, that does what you all want - maybe I'm mistaken.

"LGPL ... allows the work to be linked with (in the case of a library, "used by") a non-(L)GPLed program, regardless of whether it is licensed under a license of GPL family or other licenses"

ignotus666:
> Not LGPL?

Yes, sorry, I got lost among all the initialisms... I'd be ok with that.

gilgongo:
OK so if the answer to "_Does a license exist which allows developers to write proprietary parts (better additions) as long as they don’t touch the floss part?_" is "**Yes, it's the LGPL**" then I think we have a rather obvious conclusion to this discussion :-)

ann0see:
Yes. Probably then I‘d vote for the LGPL

gilgongo:
OK so I'll close this PR and open one for that then.

PR #625, opened 2021-09-30
Move to LGPL

gilgongo:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

As per discussion here https://github.com/jamulussoftware/jamuluswebsite/pull/624

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,504 @@
+                  GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 2.1, February 1999
+
+ Copyright (C) 1991, 1999 Free Software Foundation, Inc.
+ 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ Everyone is permitted to copy and distribute verbatim copies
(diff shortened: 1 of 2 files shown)


ann0see (approved):
I hope it’s ok - and we can legally change it if there are severe problems.

Linking https://en.m.wikipedia.org/wiki/Contributor_License_Agreement
and following thought:
We might still need some clause like only if the 5 most active contributors (=most modifications) to the code in question can be put under a different FLOSS license. Just to be sure a license change is not impossible with a big majority of contributors but hard to achieve if it goes into the wrong direction.

The background is still the AppStore thingy.

gilgongo:
I'm not sure what you mean. We can't add any extra restrictions to the LGPL - that's what  the Apple thingy was all about. We agreed to disclaim the one specific case of redistribution by Apple. What we're talking about here is distribution by **us.** 

Right now, and unless anyone can find any other problems, the LGPL does what I think you and others want. If we ever wanted to migrate the licence to something else (and this applies to the app repo too) then that's another discussion entirely.

PR #627, opened 2021-10-23
Update to symlink version

ignotus666:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

This PR does a number of things - I think it's best to get it in early (before creating or editing any new 'next-release' branch anyway) as there are changes to the file structure that could cause conflicts otherwise. It also needs to go to 'release' because of how some GH actions scripts work (namely 'add-lang' as it only triggers if it's on the 'main' branch). 

These are the changes:

- All translatable files that were scattered around and not within the `/wiki/en/` folder (those in `/_data`, `/_includes`, the `1-index.html` files) are now converted to .po format and are contained in each respective language folder. Translators can now focus solely on that folder and don't need to check anywhere else.
- This has been done by moving the EN version of these files to `/wiki/en/misc` and using symlinks to them in their former locations.
- A new script has been added to `/_po4a-tools` that takes care of creating all the appropriate files when a new language is added. It also adds the new language code to the array in `_config.yml` and creates a screenshots folder for it, so when a new language is added, (almost) everything is taken care of - again, the new translator only has to translate what's in their po folder and if they can, replace the screenshots with others in their language (this has been documented).
- Updated `README`s to reflect these changes.
- Several po4a and GH actions scripts have been updated. E.g. the line wrapping in .po files was not properly fixed but hopefully it has been sorted now definitively (my tests seem to confirm this).
- The po4a deb package has been updated to the latest release on my fork of the assets repo, so if/when this PR gets merged, I'll merge that too (it has to be merged right before this PR).

This has been fairly extensively tested and works, but any and all suggestions/questions are welcome.


gilgongo (approved):
I'm not in much of a position to test this really, but unless there's some risk in the approach, we should probably merge so that it's not a blocker for subsequent stuff.

ignotus666:
As the GH actions script that could trigger a proper test only works if it's on the main branch, such a test can only be carried out on either a fork or here after merging this PR. I tested it about a zillion times on my fork first, but then again, you never know ;) Adding a new language (opening an issue with the new lang [xx_YY] + adding the 'new language' label) would test all the most significant and "risky" changes done (GH scripts + po4a scripts). 

**EDIT** : No tests can be done until a 'next-release' branch has been created.

Also, important: there's a PR [here](https://github.com/jamulussoftware/assets/pull/3) that must be merged before this one.

PR #631, opened 2021-10-28
Move pngs in root to assets/img/_favicon

ignotus666:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

Move icons from `/` to `assets/img/_favicon` to declutter the root folder. A ruby script in `_plugins` moves them all back to the root folder when the jekyll site is deployed, so in essence it works just as before.


ignotus666:
Works locally but the zipped site here doesn't... closing.

PR #634, opened 2021-10-30
Update 2021-01-05-Jamulus-Sound-Devices.md

leoinlios:
Added section for device [Roland TD-27]

# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

### DIFF
--- a/_posts/2021-01-05-Jamulus-Sound-Devices.md
+++ b/_posts/2021-01-05-Jamulus-Sound-Devices.md
@@ -254,6 +254,16 @@ _More testing required._
 
 ***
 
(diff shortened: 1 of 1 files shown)


leoinlios:
The comment about not allowing you to listen to the computer's own audio is inaccurate. The module does allow you to listen to your computer audio, and you can perfectly hear yourself play with the server delay. What the module does NOT allow you to do is mute your direct monitoring, i.e. mute the sound as it comes out of the module directly to your phones.

gilgongo:
Apologies - yes, if you can't mute your direct sound then it's not advisable to use it. I'll clarify that.

PR #635, opened 2021-11-03
Create CLUI help page

gilgongo:
# Does this need translation?

- [x] Yes 
- [ ] No 

# Changes

Currently, the command line options for client and server are split between the Client and Server manuals, with common options handled as an include. However, the link to https://jamulus.io/wiki/Command-Line-Options given in `--help` currently redirects to the Server Admin manual. 

This PR creates a single page for CLUI help that we can link to from the `--help` page


gilgongo:
Oops forgot to switch to release before I made this branch.

pljones:
You've got four commits on this branch.  If you want the effect of one or more of those commits on another branch, you can create that branch, then patch it with just the commit/s you're after using git.  (I think it's `cherry-pick` but I may be wrong.)

As ever with git, there are other ways to do this.

gilgongo:
Thanks but @softins kindly offered to sort it out ... :-)

pljones:
> Thanks but @softins kindly offered to sort it out ... :-)

😄  That's one way of doing it..!

gilgongo:
OK he sorted it out so un-drafting for review.

gilgongo:
@pljones You made a comment about the page layout for the help. Is this what you had in mind? BTW also for server and client options is it correct to say, "Command-line options will not alter settings made in the GUI."?

pljones:
> @pljones You made a comment about the page layout for the help. Is this what you had in mind? BTW also for server and client options is it correct to say, "Command-line options will not alter settings made in the GUI."?

If you're running _with_ the GUI, client and server, any settings visible in the GUI when you quit are written to the inifile.

If you're running the server _without_ the GUI, the inifile is neither read nor written.

If you're running the server _with_ the GUI, command line options override the settings in the inifile (i.e. if the custom directory is "worldjam.vip" in the inifile and you use `-e localhost`, the GUI custom directory will show "localhost" and when you quit, "localhost" will get saved to the inifile if you didn't change it).

For the client, none of the command line options has a GUI equivalent nor is stored in the inifile, so, for the client, it isn't relevant to mention it.

gilgongo:
OK so assuming your third point is correct, I think we can remove that line in the client manual. For the server manual would the following work? 

"Command-line options will be overriden by their corresponding GUI controls until you re-start the server" ?

pljones:
Command line options _set_ the GUI control values.  So the GUI doesn't override the command line in that sense.  _User action_ via the GUI (after start up) overrides _user action_ via the command line (before start up).

The result of the last user action ends up in the inifile.

So "until you restart your server" isn't that clear either.  It depends on the user action taken to start up the server.  If they don't specify the same command line options, they'll get a different outcome.

If you _only_ start the GUI server from a desktop shortcut, then it's probably a good idea not to put options that affect the inifile into the desktop shortcut command, as you'll possibly get confusing results.

gilgongo:
So as not to get too complicated in the explanation, would the following work?

"Command line options will set the server's defaults at startup. You can override them with their corresponding GUI controls while the server is running."

gilgongo:
@ignotus666 @jujudusud What should we do about this? It's a bit of a problem that people following the link in `--help` see the server manual page. And the next release is a bit far off... But I doubt that many people click that link.

(BTW having said there is no actual translation in this PR, there is one line on 239 in Running-a-Server, and the note about that in Software-Manual has been removed)

ignotus666:
I have no problem with updating the translation and it seems @jujudusud doesn't either - but if we do go ahead and agree to update that now:
* this PR has to go against `release`.
* ideally we'd translate the .po files and push them to your branch, to merge everything at once, but we can't translate any .po files until _after_ this PR gets merged (due to how the GH jobs work - updated .po files only get pushed to the target branch on merging a PR).
* this means there will be a gap of time between merging this PR and updating the translations (as opposed to merging everything at once), though I don't think that's too serious.

gilgongo:
OK so if I merge this to release, then wait for you to push to my branch, then I make a PR to release again from there?

ignotus666:
> OK so if I merge this to release, then wait for you to push to my branch, then I make a PR to release again from there?

We could do that, remember to "fetch & merge" after merging your PR to `release` though, otherwise your branch won't contain the updated .po files. We could open a PR directly to release but I suppose doing it through your branch would make it "cleaner" by merging all the updated translations at the same time.

gilgongo:
OK here goes...!

gilgongo:
Ahh! Wrong branch aaagh!

gilgongo:
Ugh. Maybe we can just leave it there for the next release.

ignotus666:
Sure, whatever.

jujudusud:
> @ignotus666 @jujudusud What should we do about this?

From my point of view, a spécific page is not the solution. I think that we have to point the help page to the Client wiki page.
In the other hand, we just have to had a link to the server page in the client one saying : "The specific server options are in the server page." Nothing less, nothing more.

gilgongo:
Not sure I understand. The Command-Line-Options page needs to exist because it may be linked to from other places outside the site (maybe!).

Be that as it may, what would you do about the shared commands? I'm not sure why the current solution doesn't work. It's only for the `--help` page and the includes allow us to have the same content in 3 separate places.

PR #637, opened 2021-11-04
Document use of TMs

ignotus666:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

Documents the use of translation memories - useful for avoiding having to re-translate strings that are identical (or almost) in different documents.


gilgongo (approved):
Sorry, forgot to merge this earlier.

PR #643, opened 2021-11-28
Add (hidden) iOS installation page

ann0see:
# Does this need translation?

- [x] Yes 
- [ ] No 

# Changes

Adds a (hidden) iOS installation page just for future use.

### DIFF
--- a/wiki/en/Installation-for-iOS
+++ b/wiki/en/Installation-for-iOS
@@ -23,11 +23,7 @@ If you have don't own a PC, we suggest you to buy a [Raspberry Pi](https://www.r
 
 ## Install for iOS (for experienced users only)
 
-If you do want to try Jamulus on iOS:
-
-1. Install XCode from the AppStore
-1. Connect your device to your Mac
-1. [Compile Jamulus for iOS](https://github.com/jamulussoftware/jamulus/blob/master/COMPILING.md#ios)
+If you do want to try Jamulus on iOS, please have at the [compilation instructions for iOS](https://github.com/jamulussoftware/jamulus/blob/master/COMPILING.md#ios).
 
 ## Feedback and development
 
--- a/wiki/en/Installation-for-iOS
+++ b/wiki/en/Installation-for-iOS
@@ -1,0 +1,41 @@
+---
+layout: wiki
+title: "Installation for iOS"
+lang: "en"
+permalink: "/wiki/Installation-for-iOS"
+---
+
+{% include breadcrumb.html root="Using Jamulus" branch1="Getting Started" branch1-url="Getting-Started" %}
+
+
+# Installation for iOS
+
+Make sure you've already read the [Getting Started](Getting-Started) page.
+
+## Things to note about iOS
+Although Jamulus on iOS works quite well on new devices if everything is set up correctly, we strongly recommend only doing so if you have an audio interface and an USB-Ethernet adapter. To connect both devices to your iOS device, you should use something like the [Lightning to USB Camera Adapter](https://www.apple.com/uk/shop/product/MD821ZM/A/lightning-to-usb-camera-adapter) and a (powered USB-hub).
+If you have don't own a PC, we suggest you to buy a [Raspberry Pi](https://www.raspberrypi.org/){: target="_blank" rel="noopener noreferrer" } which is an inexpensive and small device that performs very well with Jamulus. iOS support is just a proof of concept and needs certain skills.
+
+## Install the iOS PoC (for experienced users only)
+
+**Attention:** in order to install Jamulus on your device, you must run macOS. Also - if you don't pay for a apple developer account, Jamulus will only run for at most 7 days on your device.
+If you do want to try Jamulus on iOS:
+
+1. Install XCode from the AppStore
+1. Connect your device to your Mac
+1. [Compile Jamulus for iOS](https://github.com/jamulussoftware/jamulus/blob/master/COMPILING.md#ios)
+
+## Feedback and development
+
+iOS support isn't mature yet and far from ready to be published on the AppStore. Feel free to help out:
+Just head over to the [Jamulus GitHub repo](https://github.com/jamulussoftware/jamulus/).
+
+Please also read the [APPLEAPPSTORE.LICENCE.WAIVER](https://github.com/jamulussoftware/jamulus/blob/master/APPLEAPPSTORE.LICENCE.WAIVER) concerning Jamulus being published via the Apple AppStore.
+
+***
+
+## All installed?
+
+Jamulus has been installed and can be used now. You can now take a look at the
+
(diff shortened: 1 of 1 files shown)


ann0see:
@ngocdh @emlynmac

Just created an iOS page for the documentation. I think we should promote the mobile builds somewhere not that prominent.

wiki/en/Installation-for-iOS
```diff
@@ -0,0 +1,41 @@
...
+
+{% include breadcrumb.html root="Using Jamulus" branch1="Getting Started" branch1-url="Getting-Started" %}
+
+
+# Installation for iOS
+
+Make sure you've already read the [Getting Started](Getting-Started) page.
+
+## Things to note about iOS
+Although Jamulus on iOS works quite well on new devices if everything is set up correctly, we strongly recommend only doing so if you have an audio interface and an USB-Ethernet adapter. To connect both devices to your iOS device, you should use something like the [Lightning to USB Camera Adapter](https://www.apple.com/uk/shop/product/MD821ZM/A/lightning-to-usb-camera-adapter) and a (powered USB-hub).
```
gilgongo:
```suggestion
Jamulus on iOS works quite well on new devices if everything is set up correctly. However, we strongly recommend you use an audio interface and an USB-Ethernet adapter. To connect these to your iOS device, you should use something like the [Lightning to USB Camera Adapter](https://www.apple.com/uk/shop/product/MD821ZM/A/lightning-to-usb-camera-adapter) and a (powered USB-hub).
```

I wasn't sure what "only doing so" referred to. Also slightly confusing use of "device" if it's not clear what type of device is being referred to (MacOS, iOS, some other device?)

wiki/en/Installation-for-iOS
```diff
@@ -0,0 +1,41 @@
...
+If you have don't own a PC, we suggest you to buy a [Raspberry Pi](https://www.raspberrypi.org/){: target="_blank" rel="noopener noreferrer" } which is an inexpensive and small device that performs very well with Jamulus. iOS support is just a proof of concept and needs certain skills.
+
+## Install the iOS PoC (for experienced users only)
```
gilgongo:
```suggestion
## Install for iOS (for experienced users only)
```

Unless we want to say "proof of concept". The acronym isn't that common I don't think.

wiki/en/Installation-for-iOS
```diff
@@ -0,0 +1,41 @@
...
+
+**Attention:** in order to install Jamulus on your device, you must run macOS. Also - if you don't pay for a apple developer account, Jamulus will only run for at most 7 days on your device.
+If you do want to try Jamulus on iOS:
```
gilgongo: Not sure why is this here and not in https://github.com/jamulussoftware/jamulus/blob/master/COMPILING.md#ios ? Seems oddly split into two (and I don't understand it - how do you actually get it onto your iOS device? But then I'm not familiar with iOS apps)

wiki/en/Installation-for-iOS
```diff
@@ -0,0 +1,41 @@
...
+
+Make sure you've already read the [Getting Started](Getting-Started) page.
+
+## Things to note about iOS
+Although Jamulus on iOS works quite well on new devices if everything is set up correctly, we strongly recommend only doing so if you have an audio interface and an USB-Ethernet adapter. To connect both devices to your iOS device, you should use something like the [Lightning to USB Camera Adapter](https://www.apple.com/uk/shop/product/MD821ZM/A/lightning-to-usb-camera-adapter) and a (powered USB-hub).
+If you have don't own a PC, we suggest you to buy a [Raspberry Pi](https://www.raspberrypi.org/){: target="_blank" rel="noopener noreferrer" } which is an inexpensive and small device that performs very well with Jamulus. iOS support is just a proof of concept and needs certain skills.
+
+## Install the iOS PoC (for experienced users only)
+
+**Attention:** in order to install Jamulus on your device, you must run macOS. Also - if you don't pay for a apple developer account, Jamulus will only run for at most 7 days on your device.
```
gilgongo: Should this not be in the "points to note above?

wiki/en/Installation-for-iOS
```diff
@@ -0,0 +1,41 @@
...
+{% include breadcrumb.html root="Using Jamulus" branch1="Getting Started" branch1-url="Getting-Started" %}
+
+
+# Installation for iOS
+
+Make sure you've already read the [Getting Started](Getting-Started) page.
+
+## Things to note about iOS
+Although Jamulus on iOS works quite well on new devices if everything is set up correctly, we strongly recommend only doing so if you have an audio interface and an USB-Ethernet adapter. To connect both devices to your iOS device, you should use something like the [Lightning to USB Camera Adapter](https://www.apple.com/uk/shop/product/MD821ZM/A/lightning-to-usb-camera-adapter) and a (powered USB-hub).
+If you have don't own a PC, we suggest you to buy a [Raspberry Pi](https://www.raspberrypi.org/){: target="_blank" rel="noopener noreferrer" } which is an inexpensive and small device that performs very well with Jamulus. iOS support is just a proof of concept and needs certain skills.
```
gilgongo, line 17: Not sure this is needed TBH. If we cross-reference here to other platforms, why aren't we doing it on the other "install" pages?

ann0see: We‘re saying the same on the Android page? I‘d say this is mobile specific.

wiki/en/Installation-for-iOS
ann0see: Yes.

wiki/en/Installation-for-iOS
ann0see: See my other PR. Maybe that’s enough? We might be able to remove the Xcode step then.

wiki/en/Installation-for-iOS
gilgongo: Oh I didn't realise that. I don't really associate Rpi with mobile. BTW is there a reason why we're saying for iOS you need an audio interface and an USB-Ethernet adapter, while on Android we're not saying that?

ann0see: Good point. It can be added of course. 
Personally, I have the feeling that Jamulus runs better on iOS than Android. So having the section on Android is even more important

ann0see: The reason why I wrote the Raspberry Pi thing was to warn users and provide an easier solution. I assume many people would like to use Jamulus on mobile since it’s what the world is like today. 

Background:
 Many musicians I know demand mobile access (many lack computer knowledge, don’t have a pc,…). For the choir I sing in there were stories like (no. I can’t give my child access to my PC. They might delete my work data. They’ve got an iPad for school work. That’s enough!)
The easiest way would be to provide a Raspberry Pi which has everything set up, so the child can’t delete the important data.

gilgongo (approved):

PR #644, opened 2021-11-28
Create a basic download page

ann0see:
# Does this need translation?

- [x] Yes 
- [ ] No 

# Changes

Very basic version of https://github.com/jamulussoftware/jamuluswebsite/issues/322
not yet tested.

### DIFF
--- a/2-download.html
+++ b/2-download.html
@@ -1,7 +1,33 @@
 ---
-layout: redirect
-redirect: "/"
+layout: wiki
+title: "Download Jamulus"
 lang: "en"
 permalink: "/download"
 category: "hidden"
 ---
+
+# Quick download links for Jamulus
+
+Make sure you read the [Getting Started](Getting-Started) page and follow the installation guides if needed.
+This is just a collection of download/upgrade links to speed up the installation process. Just click on the version you want to install.
+
+## Direct download links
+If you wish, you can read the changelog of the latest Jamulus version on the [GitHub release page]({{ site.download_overview_link }}).
+
+### Windows
+[Download for Windows (ASIO)]({{ site.download_root_link }}{{ site.download_file_names.windows }}){: .button}
+[Download for Windows (JACK)]({{ site.download_root_link }}{{ site.download_file_names.windows-jack }}){: .button}
+
+### macOS
+[Download for macOS]({{ site.download_root_link }}{{ site.download_file_names.mac }}){: .button}
+[Download for macOS (legacy)]({{ site.download_root_link }}{{ site.download_file_names.mac-legacy }}){: .button}
+
+### Debian/Ubuntu
+[Download .deb (GUI)]({{ site.download_root_link }}{{ site.download_file_names.deb-gui }}){: .button}
+[Download .deb (headless)]({{ site.download_root_link }}{{ site.download_file_names.deb-headless }}){: .button}
+
+### Android
+[Download APK]({{ site.download_root_link }}{{ site.download_file_names.android }}){: .button}
+
+## Alternative mirror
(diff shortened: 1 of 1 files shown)


2-download.html
```diff
@@ -1,7 +1,33 @@
 ---
-layout: redirect
-redirect: "/"
+layout: wiki
+title: "Download Jamulus"
 lang: "en"
 permalink: "/download"
 category: "hidden"
```
ann0see, line 6: Still hidden to not show it on google. However, as far as I remember, I've liked this page in the software.

ignotus666: If this needs translation, it should be moved into /wiki/en (e.g. /wiki/en/misc) so the .po files can be created for each language. If for some reason it needs to exist in the root dir, it has to be moved anyway and a symlink created.

gilgongo: Should there also be a link to replace the OS-specific ones on /wiki/en/misc/1-index.html?

ann0see: Not that up to date with the structural changes of the last few months. Probably symlinks should be created then?

ignotus666: Not sure I follow... Basically, any docs requiring translation need to be in /wiki/en/ (or any sub-directory in there) as that's where po4a looks for translatable documents. Those that for whatever reason need to be located elsewhere are replaced with symlinks to the "original" files, which are moved into /wiki/en/. But if the location of a file doesn't need to be fixed, it's less messy to just move it into /wiki/en/. In any event, if symlinks are added, the po4a-add-language.sh script needs to be updated (I'd do that).

ann0see: Hmm. Since it’s since the file is linked in the software, it should be in the root. So we can‘t move it.

gilgongo: What do you mean by "linked in the software"?

ann0see: Sorry, I thought I linked upgrade.html to download.html because of https://github.com/jamulussoftware/jamulus/pull/1980

however, jamulus.io/download and jamulus.io/upgrade should link to this download page once it’s finished

ignotus666: With all the language versions of upgrade.html and download.html the root folder is going to get quite cluttered.

gilgongo: >  jamulus.io/download and jamulus.io/upgrade should link to this download page once it’s finished

Not sure I follow, but is it that we **can** put them in en/wiki and have them translated as normal, but if we want "special" URLs for them  (ie jamulus.io/download as opposed to jamulus.io/wiki/download) we need to have redirects in the root, is that right? These can be symlinks to the language versions though, no?

ignotus666: > Not sure I follow, but is it that we can put them in en/wiki and have them translated as normal, but if we want "special" URLs for them (ie jamulus.io/download as opposed to jamulus.io/wiki/download) we need to have redirects in the root, is that right?

Correct

>These can be symlinks to the language versions though, no?

Yes. The thing is that we end up with another 12 files (symlinks) in the root folder - increasing in number if more languages are added.

gilgongo: I see. And I'm pretty sure GH Pages doesn't do server-side redirects either, so I guess there's no alternative.

ann0see: Do you think it's worth cluttering the root? Or do we find a better way to solve this problem?

ignotus666: Can't the link in the app be changed for the next release?

gilgongo: It could. But we'd also like to have jamulus.io/download too...

ann0see:
Closing as this would need more work...

PR #650, opened 2021-12-18
Corrections for 3.8.1

### LINKED ISSUE #647: User Manual requires update to match 3.8.1
**URL**
[User Manual](https://jamulus.io/wiki/Software-Manual) https://jamulus.io/wiki/Software-Manual

**Describe the bug**

_Delay and Buffers LEDs section_
Delay and Buffers LEDs section requires updating to match the current software. The Buffer LED is not named Buffer, there is only LED for Jitter.
The Delay details can be enhanced to include text about the calculation of the ms value being displayed here.

_Measurements section_
The Measurements section no longer exists in the Audio/Netwerk setup tab. Only the Audio Stream Rate still applies to this.
Screenshot and text requires updating.

Details about ping should probably move to the main window section, however the ping time description requires updating as well.
The current text seems to fit better to the overall delay, as the ping time does not include the audio stream.

_Advanced setup_
Screenshot does not match current software, the description below the screenshot has been update to match but the screenshot still shows the "Custom Central Server Address" text.

gilgongo:
- [x] Yes 
- [ ] No 

# Changes

Corrections/updates as per #647


wiki/en/Software-Manual.md
```diff
@@ -26,9 +26,11 @@ This manual documents the Jamulus client application for use by musicians and si
 	<figcaption>Your local mix when connected to a Server</figcaption>
 </figure>
 
-## Delay and Buffer LEDs
+## Ping, Delay and Jitter
 
-**Delay** shows the status of the current audio latency:
+**Ping** shows your network latency in milliseconds, the lower the better. Ping time contributes to overall delay (see below). The most probable causes for a high ping are that your distance to the server is too large or your internet connection is not sufficient.
```
ignotus666: > or your internet connection is not sufficient. 

This sounds a bit vague and might be taken to mean that increasing bandwidth will solve the problem - which it often won't.

gilgongo: Yes, the wording here is the original from the "measurements" section we had but I wondered about that. Maybe just leave it as _"The most probable causes for a high ping are that your distance to the server is too large."_ Otherwise, we'll start getting into troubleshooting routers and whatnot here.

ignotus666: Yes, I don't think it's worth getting into all the reasons why.

wiki/en/Software-Manual.md
```diff
@@ -28,7 +28,7 @@ This manual documents the Jamulus client application for use by musicians and si
 
 ## Ping, Delay and Jitter
 
-**Ping** shows your network latency in milliseconds, the lower the better. Ping time contributes to overall delay (see below). The most probable causes for a high ping are that your distance to the server is too large or your internet connection is not sufficient.   
+**Ping** shows your network latency in milliseconds, the lower the better. Ping time contributes to overall delay (see below). The most probable causes for a high ping are that your distance to the server is too large.
```
ignotus666: Maybe: 

"The most probable cause~~s~~ for a high ping time ~~are~~ is that your distance to the server is too large."

Issue #646, opened 2021-12-12
Add Dutch [nl]


henkdegroot:
# Language you want to translate to
Dutch

# Description
I would like to add Dutch translation to the Jamulus web-site

ignotus666:
Hi @henkdegroot, thanks for volunteering! As per the above message, the files for Dutch are now ready for translation. If you have any questions about the process, please don't hesitate to ask.

henkdegroot:
Many thanks, I will start working the translation.

Does this mean that the Dutch text will only appear with the next Jamulus version or can it be that the web-site be updated before  that?

ignotus666:
It would be published with the next release. The way it's set up, translations happen on the 'next-release' branch, which gradually moves ahead of the 'release' branch as changes are made and the app release approaches. Edits/corrections can be made to existing languages on 'release', but unfortunately adding a new language to it is not currently "supported" - I'm not 100% sure but I think it could potentially lead to conflicts between branches.

henkdegroot:
No problem, thanks for clarifying. At least that sounds like I have a bit of time to complete the translation.

ignotus666:
Also: from now until there is a new release, things will inevitably change, meaning that some parts of translations done now will become obsolete - in order to avoid this, it might be best to wait until the freeze period starts a few weeks before the release. There's nothing stopping you from starting now if you prefer to give yourself more time, but be aware that there will be changes. Typically it isn't that much - a few things are removed, others added and others moved around. Tip: make sure you use a translation memory (instructions in the README).

henkdegroot:
Sure things will change, but many will remain the same as well. So won't be wasting time and it is a good learning exercise.
I am familiarizing myself with OmegaT, which seems to work very nice. Just had a bit a pain setting up all prerequisites on the machine I use for this. Also finding some minor issues while I am working on this and am checking already if they have been addressed or not (before potentially creating a new issue to get it resolved).

PR #655, opened 2021-12-19
NL translation

henkdegroot:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

Initial Dutch translation added to next-release branch

Fixes: #646


ignotus666:
Wow, that was quick, great work, I see you even created the screenshots! I've had a brief look at a local build of the website and it seems to be all ok - I don't speak a word of Dutch so can't comment on that, but all the formatting looks good and so do the links.

henkdegroot:
Well to be honest, I did start translating a bit before I requested the new language as I wanted to make sure I could build locally to  check and verify. Also the process is documented well and easy to follow, so not much issue. The main issue I had was with some odd characters (like ë and ï) which OmegaT kept changing into double characters, hence the reason I moved to PoEdit.

ignotus666:
> The main issue I had was with some odd characters (like ë and ï) which OmegaT kept changing into double characters

That's interesting, I was wondering why you'd made the switch. OmegaT is quite a mature and widely-used tool so I wonder if there is some configuration that needs to be done somewhere to properly display those characters. Spanish uses diaeresis on some letters but I never ran into problems.

Glad to hear you found the documentation easy to follow - you're the first person to start a translation from scratch with the "new" system, so any feedback is welcome :)

henkdegroot:
The issues I had with OmegaT may have been caused by the fact that I use a virtual linux machine for this and copy files to/from my windows machine. I also did not realize that the "source" folder is not updated (or at least not always) with the translation but only the "target" folder. I used the editor to my first round of translation and then noticed the issues with these characters. This was all on the linux machine. Next I tried PoEdit and did not have any problem updating the incorrect characters. I did not want to spent more time on this and since PoEdit works fine, I will keep using that.

PR #671, opened 2021-12-31
Move to use GH user-images

### LINKED ISSUE #554: Documentation repository might grow due to new image library

gilgongo:
# Does this need translation?

- [x] Yes 
- [ ] No 

# Changes

As per: https://github.com/jamulussoftware/jamuluswebsite/issues/554

This puts all the main images into include files that use the GH user-images server to stop the website repo growing infinitely large over time.


ann0see (commented):
Thanks! It’s an interesting approach btw. Why don’t you use variables like for the nav defined in the _data directory?

ann0see:
Sorry. Forgot about the linked issue 😳. Probably it’s easier to have includes?

gilgongo:
I'm not sure how to do the translated versions though. Seems a bit unfair to leave the translators to do it.

gilgongo:
@ignotus666 Is it possible to spare the translators the work of converting the images to includes? If I was to make this change to the other languages myself, what would that entail? I guess I'd need to learn how to use a .po editor?

ignotus666:
When you're done with all the screenshots in each language and that's merged, the .po files will have been updated. Yes, then it's a matter of going through the affected files in each language with a .po file editor (e.g. Poedit) and updating the links. The readme in ./_translator-files explains how to get started using it - it's really quite simple, not very different to Qtlinguist. For your purposes, you just need to know `ctrl + b` to copy source to translation (then you just have to edit the language code) and `ctrl + enter` to validate. When finished, `Save` and close. Changed segments are easy to find, scroll down until you find one with the text in a different colour (orange I think). Anyway, just ask if in doubt!

ignotus666:
I think the process to upload the screenshots should be documented too. Right now if a new language is added, the `en` screenshots folder gets copied and renamed with the new language prefix. Then it was up to the translator to create the actual screenshots. I suppose that can stay the same (the idea is that EN screenshots are shown until the new ones are created, even if they edit the links in the docs before having created them), so now all that's needed are instructions on what to do with the new screenshots and the .inc files.

gilgongo:
OK good - I'll have a look at using Poedit (and adjust the docs too).

gilgongo:
OK I've updated the .po files for each language. 

Also BTW I thought it might be better to use the en version of the LED images as they don't need translating. Dunno if that makes things inconsistent in the .po files, but does mean that if we do change them then they all automatically update in the translations.

ignotus666:
There was a conversation about sharing screenshots that didn't need translating. My view is that having some that need to be left with -en in the link but not others means you have to make it clear to translators and it increases the chances of mistakes. Otherwise you can just say "change the language code to yours in all links" and that's it.

gilgongo:
Yes - pros and cons. Is it possible to symlink them back to the English ones?

ignotus666:
I cloned your branch, created a symlink to `led-green.inc`, moved it into `es-screenshots`, and finally changed the link in the Spanish .po file to use `-es` in it. It works.

gilgongo:
OK good (I'm symlinking the `img` dir into `_includes` come to think of it too) - I'll put them in as symlinks then, since that would be one less thing to worry about if we changed those images.

gilgongo:
OK that's done. Feel free to test.

ignotus666:
Checked locally - it all looks good. And the size of the website artifact is down by almost 70% !

gilgongo:
OK good - I also updated the README.md files for both the repo and the _translators which I hope is clear enough for people to understand.

ignotus666:
Yeah, saw those and they look ok.

PR #677, opened 2022-01-09
Server UI Changes

gilgongo:
# Does this need translation?

- [x] Yes 
- [ ] No 

# Changes

A number of changes across the docs are implied by changes to the server UI (notably, but not exclusively, the abolition of the concept of "private servers") in https://github.com/jamulussoftware/jamulus/pull/2199 

See also the discussion on vocabulary here: https://github.com/jamulussoftware/jamulus/discussions/2228


wiki/en/FAQ.md
```diff
@@ -28,9 +28,9 @@ Bear in mind also that all participants should follow [The Golden Rule](/wiki/Cl
 
 ### How do I know if I can join a server? Are there rules?
 
-In general, if somebody sets up a public server, they accept that anyone can play on it. Jamulus has no password protection or other authentication mechanisms. However, some servers may state their policies in the welcome message you will see in the chat window.
+In general, if somebody sets up a Registered Server, they accept that anyone can play on it. Jamulus has no password protection or other authentication mechanisms. However, some servers may state their policies in the welcome message you will see in the chat window.
```
pljones: > if somebody sets up a Registered Server, they accept that anyone can play on it

Not quite right.  You can register a Server with a Directory and still have it Private - i.e. server firewall only allows certain client connections, Directory firewall only allows certain client connections and Directory only allows certain Server registrations.

If you Register a Server, you accept that anyone with access to the Directory can see that your server is registered.  For the Public Directories, this means _everyone_ running Jamulus.

wiki/en/FAQ.md
```diff
@@ -28,9 +28,9 @@ Bear in mind also that all participants should follow [The Golden Rule](/wiki/Cl
...
 
-You can also set up a private server on Jamulus and give others your address to connect to that. Have a look at [this page for more information](/wiki/Running-a-Server).
+You can also set up Unregistered Server on Jamulus and give others your address to connect to that. Have a look at [this overview for more information](/wiki/Running-a-Server#server-types).
```
pljones: > You can also set up Unregistered Server

I'd say something more like:

You do not have to Register your Server - you can give the address to others and they can connect directly.

wiki/en/FAQ.md
```diff
@@ -57,7 +57,7 @@ No. If you just intend to connect to other people’s servers, then all you need
 <img src="{% include img/en-screenshots/diagram-overview.inc %}" loading="lazy" alt="Diagram showing how Jamulus works">
 
 
-Jamulus works on the client-server principle. Everybody’s audio is sent to a server, mixed and processed there. Afterwards, the audio is sent back to each client. If a server is made public and registered on a public directory, its information will be broadcast to all clients.
+Jamulus works on the client-server principle. Everybody’s audio is sent to a server, mixed and processed there. Afterwards, the audio is sent back to each client. If a server is registered in a Jamulus directory, the server's information will be broadcast to all clients using that directory.
```
pljones: > the server's information will be broadcast

Hm.  I don't like "broadcast" here, now I see it.  The client _asks_ the directory for the information.  Each client gets a _separate_ response from the directory, it's not a single broadcast signal.  "provided" would be better.

wiki/en/FAQ.md
```diff
@@ -66,10 +66,10 @@ Adding video support adds a lot of complexity. You can use other software like J
...
+Normally, network address translation (NAT) firewalls prevent incoming requests initiated from outside the local network. Inbound traffic is only possible for packets relating to an outbound request (strictly speaking, “related” and “established” packets to an initial outbound connection). In Registered Server mode, when your Jamulus server connects to a Directory it of course initiates an outbound connection. From then on, the Directory sends (established/related) connection “pings” (not ICMP pings) to your server at regular intervals to keep the relevant NAT port(s) open on your router/firewall.
 
-However, in private mode, clients have to **initiate** connections into the server’s network. NAT firewalls prevent this, so you need to tell them to allow incoming connection requests on the Jamulus port using port forwarding.
+However, in Unregistered mode, clients have to **initiate** connections into the server’s network. NAT firewalls prevent this, so you need to tell them to allow incoming connection requests on the Jamulus port using port forwarding.
```
pljones, line 74: I'd add another paragraph (between the two?).  Something like this (although unproven...):

"Note that this does not work for all routers in all cases.  If you have some client users who cannot find your server in a directory, it may mean you need to set up port forwarding."

(or rephrase to fit better at the end...)

wiki/en/Privacy-Statement.md
```diff
@@ -7,32 +7,38 @@ permalink: "/wiki/Privacy-Statement"
 
 # Privacy Statement
 
+## Definition of Terms
+
+- "**Server**" The Jamulus server software, as opposed to the host machine/OS it is running on.
+- "**Client**" The Jamulus software used to connect to a Server
+- "**Directory**" A Jamulus Server configured to display a list of Servers to Clients
```
pljones: "display" is _wrong_.  "supply" is correct and not technical.

wiki/en/Privacy-Statement.md
```diff
@@ -7,32 +7,38 @@ permalink: "/wiki/Privacy-Statement"
 
 # Privacy Statement
 
+## Definition of Terms
+
+- "**Server**" The Jamulus server software, as opposed to the host machine/OS it is running on.
```
pljones: I'd add "in general" unless you've been _incredibly_ careful :).

wiki/en/Privacy-Statement.md
```diff
@@ -7,32 +7,38 @@ permalink: "/wiki/Privacy-Statement"
...
+- "**Client**" The Jamulus software used to connect to a Server
+- "**Directory**" A Jamulus Server configured to display a list of Servers to Clients
+
 ## Web site
 
 The website at jamulus.io is served using GitHub Pages. We do not collect your personal data or set tracking cookies. When you connect to the jamulus.io site, your IP is sent to GitHub Pages.
```
pljones: Should `jamulus.io` be a link?  Should we supply a link to Github's information about Github Pages?

wiki/en/Privacy-Statement.md
```diff
@@ -7,32 +7,38 @@ permalink: "/wiki/Privacy-Statement"
...
 
-When you connect to a public server, your profile is also available to third parties from the Directory Server to which that server is registered. This is can be for informational purposes about the status of the public Jamulus network (for example, [here](https://explorer.jamulus.io/)), but may not be limited to that. Profile information is not otherwise logged or stored by the Jamulus server you are connected to, or by the Jamulus Directory Server, but may be stored or processed by third parties.
+When you connect to a Server, your profile may also be available to third parties from the Directory to which that Server is registered. This is can be for informational purposes about the status of the public Jamulus network (for example, [here](https://explorer.jamulus.io/)), but may not be limited to that. Profile information is not otherwise logged or stored by the Jamulus Server you are connected to, or by the Jamulus Directory, but may be stored or processed by third parties.
```
pljones: Typo:
> This is can be for

wiki/en/Privacy-Statement.md
```diff
@@ -7,32 +7,38 @@ permalink: "/wiki/Privacy-Statement"
...
 
-The IP addresses of all public servers registered with the Directory Server can also be seen by third parties for informational or other purposes (for example [here](https://explorer.jamulus.io/)). Your public IP address is otherwise not logged or stored by Jamulus, but may be stored or processed by third parties.
+The IP addresses of all Servers registered with the Directory can also be seen by third parties for informational or other purposes (for example [here](https://explorer.jamulus.io/)). Your public IP address is otherwise not logged or stored by Jamulus, but may be stored or processed by third parties.
 
 ### Audio Recordings
 
-You will see a notice if you are connected to a Jamulus server while server recording is turned on. Recordings of each player are stored by the server separately as .WAV files and only the server operator has access to them.
+You will see a notice if you are connected to a Jamulus Server while server recording is turned on. Recordings of each player are stored by the server separately as .WAV files and only the server operator has access to them.
```
pljones: > Server while server recording

How about "when" rather than "while"?  It's more absolute.

pljones: > only the server operator has access to them.

Unless they choose to make them available to third parties.

wiki/en/Running-a-Server.md
```diff
@@ -23,7 +23,7 @@ permalink: "/wiki/Running-a-Server"
 
 The short answer is **no**. For various reasons, you should try to use other people's servers, at least at first while getting your sound set up with the Jamulus client.
 
-Note that you can have a "private" session with other people on a Public Server by simply soloing each other. You will then not be able to hear anyone else if they enter your server.
+Note that you can have a "private" session with other people on a Server by simply soloing each other. You will then not be able to hear anyone else if they enter your server.
```
pljones: Is it worth adding that this doesn't affect things like chat and profile information being seen by others?  Or is that too obvious?

wiki/en/Privacy-Statement.md
gilgongo: Where would "in general" go here?

wiki/en/Running-a-Server.md
```diff
@@ -50,55 +49,47 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
 
 - Running a server may require you to adjust any firewalls running on or outside of your machine or cloud host.
 
-- Running a Private Server at home will require you to [port forward](#running-a-private-server) on your router. When running a Public Server, port forwarding should not be necessary in most cases, but it's advisable to do so because some networks may not work properly with Jamulus in its default mode.
+- Running an Uregistered Server at home will require you to [port forward](#running-an-unregistered-server) on your router. When running a Registered Server, port forwarding should not be necessary in most cases, but it's advisable to do so because some networks may not work properly with Jamulus in its default mode.
```
pljones, line 52: Ah, I see the port forwarding "problem" is mentioned here already :) .

wiki/en/Running-a-Server.md
gilgongo: I think the understanding of "mute" should be enough, but we could add something to clarify.

wiki/en/Running-a-Server.md
```diff
@@ -50,55 +49,47 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
...
-You can run your server in one of three "modes" (either at home or on a 3rd party host):
-
-### 1. Public
+You can run your server in different ways (either at home or on a 3rd party host):
 
-Use this mode when you want anyone to join your server, whoever they are.   
+### 1. Registered 
 
-Your server will be listed in the Directory that clients use by default. Musicians can then discover and connect to your server.
+Your server will be listed in a Directory. By default, Jamulus has a list of Public Directories that clients can connect to. If you register with one of these, anyone can then discover and connect to your server.
```
pljones: Check for trailing whitespace?

wiki/en/Running-a-Server.md
gilgongo: Yes, a relatively recent addition. These docs haven't been changed much otherwise in almost two years I think.

gilgongo:
I'll wait a bit for the main PR to mature before updating the screenshots.

wiki/en/Running-a-Server.md
```diff
@@ -50,55 +49,47 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
...
 </figure>
 
+### 2. Unregistered 
 
-
-### 2. Private
-
-This is the default when starting a server for the first time. Private Servers are not listed by Directories, so only musicians who know your server's address to will be able to connect to it. This is useful because Jamulus does not (in any server mode) let you control who can connect to a server.
-
+This is the default when starting a server for the first time. Unregistered Servers are not listed by Directories, so only musicians who know your server's address to will be able to connect to it. This is useful because Jamulus does not let you control who can connect to a server.
```
pljones: > This is the default when starting a server for the first time.

Not yet :).  By default - if running with the UI - you're registered with Any Genre 1.

If you run headless, you'll not register unless you supply a directory address (because headless servers don't have inifiles, not supplying a directory is the only way to say "don't register me with Any Genre 1").

wiki/en/Running-a-Server.md
```diff
@@ -50,55 +49,47 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
...
+### 3. Custom Directory
 
+If you want to run a number of Servers, possibly also behind a firewall or on a LAN, you may want to run your server as a Directory. Examples include online events, music associations, sectional rehearsals or music lessons for schools.
 
-### 3. Directory
-
-For when you want to run a number of Private Servers, possibly also behind a firewall or on a LAN. Examples include online events, music associations, sectional rehearsals or music lessons for schools.
-
-To view servers listed by a custom Directory Server, musicians must enter the address in their client’s "Custom Directory Server" settings field.  They will then see a Connection Setup list which is generated by your Directory.
+To view servers listed by a Custom Directory, musicians must enter its address in their client’s "Custom Directory Server" settings field.  They will then see a Connection Setup list which is generated by your Directory. Custom Directories otherwise work for clients in the same way as Public Directories.
```
pljones: > client’s "Custom Directory Server" settings field

image
"Custom Directories", as we store multiple at the client.

pljones: > They will then see a Connection Setup list

image
It's just the "Connect" window, where the server list is supplied by the chosen Directory.

wiki/en/Running-a-Server.md
```diff
@@ -50,55 +49,47 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
...
 
 Up to 150 servers can then register with your Directory by setting your Directory address in their `--directoryserver` option at startup.
 
-To run a server as a Directory, it should be configured with `--directoryserver localhost` (that is, making itself the Directory to query for servers).
+To run a server as a Directory, it should be configured with `--directoryserver localhost` (that is, making itself the Directory to query for servers). When running a server with the GUI, select "Custom" from the Directory drop-down list.
```
pljones: > select "Custom" from the Directory drop-down list

... and enter "localhost" as the Custom Directory server address.

wiki/en/Running-a-Server.md
```diff
@@ -50,55 +49,47 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
...
-To view servers listed by a custom Directory Server, musicians must enter the address in their client’s "Custom Directory Server" settings field.  They will then see a Connection Setup list which is generated by your Directory.
+To view servers listed by a Custom Directory, musicians must enter its address in their client’s "Custom Directory Server" settings field.  They will then see a Connection Setup list which is generated by your Directory. Custom Directories otherwise work for clients in the same way as Public Directories.
 
 Up to 150 servers can then register with your Directory by setting your Directory address in their `--directoryserver` option at startup.
 
-To run a server as a Directory, it should be configured with `--directoryserver localhost` (that is, making itself the Directory to query for servers).
+To run a server as a Directory, it should be configured with `--directoryserver localhost` (that is, making itself the Directory to query for servers). When running a server with the GUI, select "Custom" from the Directory drop-down list. 
 
-#### Points to note about Directories
+#### Points to note about Custom Directories
```
pljones: > Custom

Doesn't make a difference.  Several of the Public Directories use `--listfilter` (using the undocument version filter).

wiki/en/Running-a-Server.md
```diff
@@ -50,55 +49,47 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
...
 
 - If you want to control which servers can register with your Directory, you can enable a whitelist with the `--listfilter` command line option. See the section on command line options later in this document.
 
-- When running a public Directory Server behind a NAT firewall on a private network, use the `--serverpublicip` option to specify the public IP address of the server(s) being listed by your Directory. This is necessary to allow clients on the public Internet to connect to them via NAT. Note that for the servers using this option, you will still need proper port forwarding in your router/firewall.
+- When running a Directory behind a NAT firewall on a private network, use the `--serverpublicip` option to specify the public IP address of the server(s) being listed by your Directory. This is necessary to allow clients on the public Internet to connect to them via NAT. Note that for the servers using this option, you will still need proper port forwarding in your router/firewall.
```
pljones: >  public IP address of the server(s) being listed by your Directory

Only applies to servers on the same LAN / with the same public IP.  The phrasing implies it applies to _all_ registered servers.

wiki/en/Running-a-Server.md
```diff
@@ -150,21 +141,17 @@ Note also that mean ADSL2 transfer rate is 10 Mbit/s for downstream and 1 Mbit/s
 
 <figure><img src="{% include img/en-screenshots/server-window-setup.inc %}" style="width:80%; border:5px solid grey;" loading="lazy" alt="Image of the Jamulus server setup window"></figure>
 
-### Make My Server Public
-
-By default, you will be running a Private Server and need to [read these instructions](#running-a-private-server) to have others connect to you in this mode.
+### The Directory list 
 
-When making your server public, you should see a confirmation message saying whether your server has registered successfully. If not, and you leave your server running, it will keep trying to register until a free slot becomes available.
+**None**: By default, you will not be connected to a directory and will be running in unregistered mode.  [Read these instructions](#running-an-unregistered-server) to have other people connect to your server in this mode.
```
pljones: See previous note - currently the GUI registers with Any Genre 1.

wiki/en/Running-a-Server.md
```diff
@@ -150,21 +141,17 @@ Note also that mean ADSL2 transfer rate is 10 Mbit/s for downstream and 1 Mbit/s
...
 
+**Genre**: To allow other people to see your server on one of the built-in public directories, select your desired genre directory. You should see a confirmation message saying whether your server has registered successfully. If not, and you leave your server running, it will keep trying to register until a free slot becomes available.  
 
-### Genre
-
-Because there is a limit of 150 servers per Directory, you need to select which Directory Server you want to advertise your server on. Select a genre if you intend to limit players to that style (players can then see your server in their genre list). This will make your server appear in the relevant server list for clients.
-
+**Custom**: This sets your server to be a directory on which other servers can be listed. Note that when setting this mode, you may want also to specify the location of server list file (See "Options" below) unless you want to disconnect all listed servers when restarting Jamulus.
```
pljones: > **Custom**: This sets your server to be a directory on which other servers can be listed. 

Wrong.

Custom allows you to set and use a Custom Directory server address.  Nothing more than that.

"Registering" with localhost is what makes a server a directory.

wiki/en/Running-a-Server.md
pljones: (I can't remember if I kept this existing behaviour -- I think I did...  It's probably worth including the change if I didn't already...)

wiki/en/Running-a-Server.md
```diff
@@ -181,28 +168,31 @@ This sets the path to where server's recording will be stored. With this path se
...
+# Running an Unregistered Server
 
-It is highly recommended to test your server in **public mode first** so as to narrow down any subsequent problems in private mode.
+It is highly recommended to test your server on a **Public Directory** first so as to narrow down any subsequent problems in uregistered mode.
 
-### Port forwarding
+### Port forwarding 
 
-Normally, people from outside your home network cannot see things inside it. So if you want to operate a Jamulus server at home, you need to open a hole in your router to let Jamulus clients connect to it.
+If you are running a server on your home network, people from outside will not be able to see things inside it. So if you want to operate a Jamulus server, you need to open a hole in your router to let Jamulus clients connect to it
```
pljones: More trailing whitespace - and missing punctuation at end of sentence.

wiki/en/Server-Troubleshooting.md
```diff
@@ -19,17 +19,17 @@ permalink: "/wiki/Server-Troubleshooting"
 
 </details>
 
-## Servers - Public
+## Servers - Registered 
 
 ### Why doesn't my server show up in the list? Why isn't it registering?
 
 If you are registered OK (you can [see it here](https://explorer.jamulus.io/)) and you or your friends can't see your server, you may need to wait, or start your client with the `--showallservers` option and try connecting from there ([see command line options](Software-Manual#command-line-options)) on how to start your client with a config option).
```
pljones: It's worth mentioning again that the magic port forwarding fix may not be working.

wiki/en/Software-Manual.md
```diff
@@ -84,7 +84,7 @@ The Connection Setup window shows a list of available servers together with the
 
 You can filter the list by server name or location. To list only occupied servers, enter a "#" character.
 
-If you know the IP address or URL of a private server, you can connect to it using the Server Name/Address
+If you know the IP address or URL of unregistered server, you can connect to it using the Server Name/Address
```
pljones: > If you know the IP address or URL

image

It's _not_ a URL -- it doesn't have the right format.  _All_ URLs have a protocol prefix.  We _do not_ support URLs.  Do not say that we do in the documentation...

Also the label is just "Server Address" (currently - to change to just "Address" at some point, as this screen is for connecting to a server).

So
"If you know the _address_ of a server," (doesn't _have to be_ unregistered, so no point saying that at all) "you can enter it in (Server) Address and connect directly".

wiki/en/Software-Manual.md
```diff
@@ -84,7 +84,7 @@ The Connection Setup window shows a list of available servers together with the
...
 field. An optional port number can be added after the address using a colon as a separator, e.g,
```
pljones: > An optional port number can be added after the address using a colon as a separator

The port number is _part of_ the Server address.  It's optional only because it has a default value.  So this needs rephrasing.

"The address will look like `jamulus.example.io:22584` (when a specific port is needed) or `192.168.3.23` (where the default port number will be used).  When the client is running with IPv6 enabled, you can use a fully specified IPv6 host address, such as `[fe80::7285:c2ff:fe80:f4ae]:22345`."

wiki/en/Software-Manual.md
```diff
@@ -84,7 +84,7 @@ The Connection Setup window shows a list of available servers together with the
...
 `jamulus.example.com:22124` (Note that IPv6 addresses must be entirely enclosed in square brackets). The field will also show a list of the most recently used server addresses.
```
pljones: > The field will also show a list of the most recently used server addresses.

It's not actually clear currently from context what "The field" refers to.

wiki/en/Software-Manual.md
pljones: Line 83, above this
> The Connection Setup window shows

...which it isn't.  It's the "Connect" window.  The section probably needs splitting into the separate fields more clearly.

wiki/en/Privacy-Statement.md
pljones: Maybe add a note after the definition saying "There may be some places in the documentation where 'server' has been used to refer to the system hosting the Jamulus Server, rather than the Server itself.  Please report these and we'll tidy them up!" :)

wiki/en/Running-a-Server.md
gilgongo: Do we not intend to make it so that the default with the GUI is unregistered?

wiki/en/Running-a-Server.md
gilgongo: Is there a reason why it doesn't do that automatically?

wiki/en/Software-Manual.md
gilgongo: It'll be self-evident anyway so I'll remove that reference.

wiki/en/Software-Manual.md
gilgongo: Not sure I understand. The window title says "Connection Setup".

pljones: On Windows, in Jamulus 3.8.1, I see:
image

The window name is "Connect" there.  (The button to make it appear is also labelled "Connect".)

wiki/en/Running-a-Server.md
pljones: Yep, but my PR as it stands hasn't changed legacy behaviour.  I'd also need to review the messing about when the settings are loaded to make sure everything still worked as intended.

I can do it in https://github.com/jamulussoftware/jamulus/pull/2199 ... probably won't be today.

wiki/en/Running-a-Server.md
pljones: The _vast_ majority of people registering with a Custom Directory are doing so as Servers.

They're _not_ setting up a Directory.

Otherwise you'd end up with more Directories than Servers... which makes no sense at all.

wiki/en/Running-a-Server.md
```diff
@@ -150,21 +141,17 @@ Note also that mean ADSL2 transfer rate is 10 Mbit/s for downstream and 1 Mbit/s
...
+**None**: By default, you will not be connected to a directory and will be running in unregistered mode. [Read these instructions](#running-an-unregistered-server) to have other people connect to your server in this mode.
 
+**Genre**: To allow other people to see your server on one of the built-in public directories, select your desired genre directory. You should see a confirmation message saying whether your server has registered successfully. If not, and you leave your server running, it will keep trying to register until a free slot becomes available.  
 
-### Genre
-
-Because there is a limit of 150 servers per Directory, you need to select which Directory Server you want to advertise your server on. Select a genre if you intend to limit players to that style (players can then see your server in their genre list). This will make your server appear in the relevant server list for clients.
-
+**Custom**: This sets your server to be a directory on which other servers can be listed. Note that when setting this mode, you may want also to specify the location of server list file (See "Options" below) unless you want to disconnect all listed servers when restarting Jamulus.
```
pljones: > This sets your server to be a directory on which other servers can be listed.

As noted else where, this is wrong.

Custom Directories are where Servers register so they can be seen by Clients viewing that Custom Directory in the Connect window.

Setting the select box to "Custom" and setting a Custom Directory server address is how to do this.

For the tiny number of operators running Directories, putting "localhost" into the Custom Directory server address means you're running as a Custom Directory yourself, rather than registering your server with a Custom Directory.

wiki/en/Running-a-Server.md
```diff
@@ -150,21 +141,17 @@ Note also that mean ADSL2 transfer rate is 10 Mbit/s for downstream and 1 Mbit/s
...
 
 ### My Server Info
 
-Type in the name, city and country so that other users can easily identify you.
+When running as a registered server this displays the server's name, city and country so that other users can easily identify it in the directory listing.
```
pljones: > this displays

Not really the right way to say it.  This is the server set up screen.  It's not "displaying" it - it's allowing it to be set by the server operator.

"Display" says "look but don't touch".

wiki/en/Running-a-Server.md
```diff
@@ -181,28 +168,31 @@ This sets the path to where server's recording will be stored. With this path se
...
-### Delay panning
+### Custom Directory Server Address
 
-This option uses small differences in sound arrival time between the two ears. It produces a stereo effect similar to natural human hearing when compared to normal "volume" panning. Note that to achieve the effect, all clients should be set to 'mono in / stereo out' (or to 'stereo', but only if transmitting a stereo signal).
+Leave this field empty unless you need to list your server on a [Custom Directory](#3-directory)
 
-### Custom Directory Server Address
+### Server List Filename
+
+If you are running your server as a custom directory and need to restart it for any reason (for example when rebooting the host), servers connected to it will be disconnected until they re-register. This does not mean that clients connected to those servers will be disconnected, but does mean that new clients will be not able to see servers listed by your directory until those servers reconnect. To enable the list of registered servers to persist between restarts, specify the location and name of a file that the directory can read and write to here.
```
pljones: > as a custom directory

Just "Directory".  Doesn't need to be a Custom one.

wiki/en/Running-a-Server.md
```diff
@@ -181,28 +168,31 @@ This sets the path to where server's recording will be stored. With this path se
...
 
-### Custom Directory Server Address
+### Server List Filename
+
+If you are running your server as a custom directory and need to restart it for any reason (for example when rebooting the host), servers connected to it will be disconnected until they re-register. This does not mean that clients connected to those servers will be disconnected, but does mean that new clients will be not able to see servers listed by your directory until those servers reconnect. To enable the list of registered servers to persist between restarts, specify the location and name of a file that the directory can read and write to here.  
 
-Leave this field empty unless you need to list your server on a [private Directory](#3-directory)
+### Delay panning
 
+This option uses small differences in sound arrival time between the two ears. It produces a stereo effect similar to natural human hearing when compared to normal "volume" panning. Note that to achieve the effect, all clients should be set to 'mono in / stereo out' (or to 'stereo', but only if transmitting a stereo signal).
```
pljones: > Note that to achieve the effect, all clients should be set to 'mono in / stereo out' (or to 'stereo', but only if transmitting a stereo signal).

I think that's redundant.  To hear a stereo effect, you need stereo out.  Goes without saying?  (Yes, cut-n-paste from above, I can see that.)

It also implies that the effect doesn't work for any of the clients if one of the clients has mono out.  I don't think that's even true as the mixer runs independently for each client output.

wiki/en/Server-Linux.md
```diff
@@ -29,9 +29,9 @@ You can control Jamulus with the `systemctl` command. For example, to stop the s
 
 `sudo systemctl stop jamulus-headless`
 
-### Running in public mode
+### Running in Registered mode
 
-The following minimum setup is required to [run a public server](Running-a-Server#server-types):
+The following minimum setup is required to [run a Registered Server](Running-a-Server#server-types):
 
 ~~~
```
pljones, line 36: I know it's unchanged... can we add a reference after line 39 (well, after line 55, I guess) to where the country IDs are specified?  It should also say clearly "Replace `genreServer:port` with one of the following options", rather than 44.

wiki/en/Server-Linux.md
```diff
@@ -29,9 +29,9 @@ You can control Jamulus with the `systemctl` command. For example, to stop the s
 
 `sudo systemctl stop jamulus-headless`
 
-### Running in public mode
+### Running in Registered mode
 
-The following minimum setup is required to [run a public server](Running-a-Server#server-types):
+The following minimum setup is required to [run a Registered Server](Running-a-Server#server-types):
```
pljones, line 34: > run a Registered Server

The example in lines 35 to 55 is to "run a Server registered with a Public Directory" - it makes no mention of registering with a Custom Directory.

wiki/en/Server-Linux.md
```diff
@@ -54,6 +54,11 @@ Set your Directory (genre) server using the `--directoryserver` (`-e`) option as
 |**Genre Classical/Folk** |`classical.jamulus.io:22524`|
 |**Genre Choral/Barbershop** |`choral.jamulus.io:22724`|
```
pljones, line 56: Add another example on how to register with a Custom Directory - showing how it's pretty much the same.

wiki/en/Server-Linux.md
```diff
@@ -54,6 +54,11 @@ Set your Directory (genre) server using the `--directoryserver` (`-e`) option as
...
 
+
+### Running a Custom Directory
+
+Set the server to use itself (in most cases this will be `localhost`) as the Directory with the `-e` or `--directoryserver` option. Note also that you may want to set the `--directoryfile` option. See [this explanation](Running-a-Server#server-list-filename).
```
pljones: > Set the server to use itself (in most cases this will be `localhost`) as the Directory...

Better to say
"Set the server to run as a Directory by using 'localhost' or '127.0.0.1' (exactly) as the `--directoryserver` (`-e`) option."

There's no "in most cases" about it -- those are literal strings that are tested for.

wiki/en/Server-Linux.md
```diff
@@ -54,6 +54,11 @@ Set your Directory (genre) server using the `--directoryserver` (`-e`) option as
 |**Genre Classical/Folk** |`classical.jamulus.io:22524`|
 |**Genre Choral/Barbershop** |`choral.jamulus.io:22724`|
 
+
+### Running a Custom Directory
```
pljones: > Running a Custom Directory

"Running as a Directory"

wiki/en/Server-Troubleshooting.md
```diff
@@ -43,22 +43,14 @@ This should be your **public** IP address (find that with [Google](https://www.g
 
 You can set your server to listen on a custom port with the `--port` option. If you do this, you will need to tell people which port to connect on. They will need to append the port number to the address of your server in the format `[serverAddress]:[portNumber]`.
 
-### Is a private server a direct server without contact to the Jamulus Directory Server?
+### **_Headless Linux_**: I have created a server but it is joining a directory and I don't want it to.
```
pljones: Not Linux specific.  (Yes, c-n-p...)

wiki/en/Software-Manual.md
```diff
@@ -80,13 +80,13 @@ pressing this button will end the session.
...
+Use the List dropdown to select a genre, click on the server you want to join and press the Connect button to connect to it. Alternatively, double-click on the server name. Permanent servers (those that have been listed for longer than 24 hours) are shown in bold.
 
-If you know the IP address or URL of a private server, you can connect to it using the Server Name/Address
-field. An optional port number can be added after the address using a colon as a separator, e.g,
-`jamulus.example.com:22124` (Note that IPv6 addresses must be entirely enclosed in square brackets). The field will also show a list of the most recently used server addresses.
+You can filter the list by server name or location using the Filter field. To list only occupied servers, enter a "#" character.
+
+If you know the address of an unregistered server, you can connect to it using the Server Address field. A port number may also be needed if the server is not running on the default port 22124 e.g,`jamulus.example.com:22124` (Note that IPv6 addresses must be entirely enclosed in square brackets, e.g. `[fe80::7285:c2ff:fe80:f4ae]:22345`).
```
pljones: > address of an unregistered server

It doesn't _need_ to be unregistered.  _Any_ server has an address and it can be used here.

> A port number may also be needed if the server is not running on the default port

No.  It _is_ needed in that case.

wiki/en/Privacy-Statement.md
gilgongo: I'm going to do a pass at the end for that. We've been using capitalised words to denote "specific" meanings like "Server" to denote the software, and lowercase to denote generic terms like the system it runs on. Same for Directory, for example.

wiki/en/Running-a-Server.md
gilgongo: Sorry, I meant when they choose the "Directory" option in the Directory drop-down. Or have I misunderstood what that means? So right now if I want to set my server up as a Directory, I have to declare that in two places in the UI?

wiki/en/Running-a-Server.md
gilgongo: See https://github.com/jamulussoftware/jamuluswebsite/pull/677#discussion_r780814854 This seems a rather odd UX.

wiki/en/Running-a-Server.md
gilgongo: The fact they can be edited in the UI here is self-evident (it's a form) and is consistent with how we talk about My Profile profile too.

wiki/en/Running-a-Server.md
gilgongo: I recall somebody was very keen to add this line when we originally did it (can't remember who) but if it's not accurate we should correct.

wiki/en/Server-Linux.md
gilgongo: I think I'm beginning to become confused about what you mean by "Custom Directory" vs "Directory". Can you explain?

wiki/en/FAQ.md
```diff
@@ -28,9 +28,9 @@ Bear in mind also that all participants should follow [The Golden Rule](/wiki/Cl
 
 ### How do I know if I can join a server? Are there rules?
 
-In general, if somebody sets up a public server, they accept that anyone can play on it. Jamulus has no password protection or other authentication mechanisms. However, some servers may state their policies in the welcome message you will see in the chat window.
+In general, if somebody lists a Server on one of the Public Directories provided by Jamulus by default, they accept that anyone can play on it. Jamulus has no password protection or other authentication mechanisms built in. However, some servers may state their policies in the welcome message you will see in the chat window.
 
-You can also set up a private server on Jamulus and give others your address to connect to that. Have a look at [this page for more information](/wiki/Running-a-Server).
+Note also that servers do not have to be registered on a Directory in order for Jamulus clients to connect to them. Server operators can simply give out the address of their  servers to those they want to play with, and they can then connect directly. Have a look at [this overview for more information](/wiki/Running-a-Server#server-types).
```
ann0see:
```suggestion
Note also that servers do not have to be registered on a (public) Directory in order for Jamulus clients to connect to them. Server operators can simply give out the address of their servers to those they want to play with, and they can then connect directly. Have a look at [this overview for more information](/wiki/Running-a-Server#server-types).
```

wiki/en/Running-a-Server.md
```diff
@@ -50,55 +49,47 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
...
-You can run your server in one of three "modes" (either at home or on a 3rd party host):
+You can run your server in different ways (either at home or on a 3rd party host):
 
-### 1. Public
+### 1. Registered 
 
-Use this mode when you want anyone to join your server, whoever they are.   
-
-Your server will be listed in the Directory that clients use by default. Musicians can then discover and connect to your server.
+Your server will be listed in a Directory. By default, Jamulus has a list of Public Directories that clients can connect to. If you register with one of these, anyone can then discover and connect to your server.
```
ann0see:
```suggestion
Your server will be listed in a Directory. By default, Jamulus has a list of Public Directories that clients can connect to. If you register with one of these, your server is believed to be public and anyone can then discover and connect to your server. 
```

Not happy with my change, but we should stress the "public" part here.

wiki/en/Running-a-Server.md
```diff
@@ -50,55 +49,47 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
...
 </figure>
 
+### 2. Unregistered 
 
-
-### 2. Private
-
-This is the default when starting a server for the first time. Private Servers are not listed by Directories, so only musicians who know your server's address to will be able to connect to it. This is useful because Jamulus does not (in any server mode) let you control who can connect to a server.
-
+This is the default when starting a server for the first time. Unregistered Servers are not listed by Directories, so only musicians who know your server's address to will be able to connect to it. This is useful because Jamulus does not let you control who can connect to a server.
```
ann0see:
```suggestion
This is the default when starting a server for the first time. Unregistered Servers are not listed by Directories, so only musicians who know your server's address to will be able to connect to it. This is useful if you don't want your server to be public since Jamulus does not let you control who can connect to a server.
```

... or published

wiki/en/Running-a-Server.md
```diff
@@ -50,55 +49,47 @@ Unless you plan on hosting more than about 5 players on a slower-speed home conn
...
-
-To view servers listed by a custom Directory Server, musicians must enter the address in their client’s "Custom Directory Server" settings field.  They will then see a Connection Setup list which is generated by your Directory.
+To view servers listed by a Custom Directory, musicians must enter its address in their client’s "Custom Directory Server" settings field.  They will then see a Connection Setup list which is generated by your Directory. Custom Directories otherwise work for clients in the same way as Public Directories.
 
 Up to 150 servers can then register with your Directory by setting your Directory address in their `--directoryserver` option at startup.
 
-To run a server as a Directory, it should be configured with `--directoryserver localhost` (that is, making itself the Directory to query for servers).
+To run a server as a Directory, it should be configured with `--directoryserver` (that is, making itself the Directory to query for servers). When running a server with the GUI, select "Custom" from the Directory drop-down list.
```
ann0see: @pljones is this true? Don't we need --directoryserver localhost or -e localhost anymore?

wiki/en/Running-a-Server.md
```diff
@@ -150,21 +141,17 @@ Note also that mean ADSL2 transfer rate is 10 Mbit/s for downstream and 1 Mbit/s
 
 <figure><img src="{% include img/en-screenshots/server-window-setup.inc %}" style="width:80%; border:5px solid grey;" loading="lazy" alt="Image of the Jamulus server setup window"></figure>
 
-### Make My Server Public
+### The Directory list 
 
-By default, you will be running a Private Server and need to [read these instructions](#running-a-private-server) to have others connect to you in this mode.
+**None**: By default, you will not be connected to a directory and will be running in unregistered mode. [Read these instructions](#running-an-unregistered-server) to have other people connect to your server in this mode.
```
ann0see:
```suggestion
**None**: By default, you will not be connected to a directory and will be running in unregistered mode. This means your server will not be publically visible. [Read these instructions](#running-an-unregistered-server) to have other people connect to your server in this mode.
```

wiki/en/Server-Troubleshooting.md
gilgongo: I think this is a pretty low-quality entry TBH and was tempted to delete it. It comes from a long list of similarly slightly odd things that were added to the wiki a while ago.

wiki/en/Running-a-Server.md
```diff
@@ -181,28 +168,31 @@ This sets the path to where server's recording will be stored. With this path se
...
 
 **Windows users** - If you want the server to start automatically on system start, enable the corresponding check box.
 
 
 
-# Running a Private Server
+# Running an Unregistered Server
 
-It is highly recommended to test your server in **public mode first** so as to narrow down any subsequent problems in private mode.
+It is highly recommended to test your server on a **Public Directory** first so as to narrow down any subsequent problems in uregistered mode.
```
ann0see:
```suggestion
It is highly recommended to test your server on a **Public Directory** first so as to narrow down any subsequent problems in unregistered mode.
```

wiki/en/Running-a-Server.md
pljones: On Linux master, I get
image

I'm not sure where the (local) hostname is coming from.

wiki/en/Server-Troubleshooting.md
```diff
@@ -43,22 +43,14 @@ This should be your **public** IP address (find that with [Google](https://www.g
...
 
-Yes. Please see [Server Types](Running-a-Server#server-types)
-
-### Can I convert my private server to being a public one?
-
-Yes. Simply check the "Make my Server Public" box in the server window and fill out the details. If you register your server, do not use port forwarding as it may prevent others using your server.
```
ann0see: Should be re-phrased and not deleted. Is it still possible to switch from registered to unregistered?

ann0see (commented):
Not very happy about a radical change like that,...
Unregistered is still very unfamiliar...

wiki/en/Server-Troubleshooting.md
ann0see: Probably we can delete it since the default is now unregistered.

wiki/en/Running-a-Server.md
pljones: Jamulus 3.8.1 GUI server on Windows registers by default with Any Genre 1.

And... another new profile...
image 
image

**However**... now I check, Jamulus _master_ (ff281ca1 2022-01-08) on Linux doesn't...

So I take this all back, I must have done it in one of the "OCD re-order" changes as a "freebie add-on".  Or accident.

ann0see:
Just opened https://github.com/jamulussoftware/jamulus/discussions/2228 to get a concensus for a new term except for "private" servers. As I said, I'm not happy about "unregistered"

wiki/en/Running-a-Server.md
pljones: There is no "Directory" option in the Server "Directory" drop down.  You pick "Custom" and enter one of the two literal strings in the Custom Directory server address that you'd pass to the `-e` option on the command line.

Basically, "Custom" works the same way as the command line option.  ("None" works like not giving the `-e` on the command line; the genre values aren't available from the command line.)

wiki/en/FAQ.md
gilgongo: > Not very happy about a radical change like that,...

Yeah I know what you mean and I think it's unlikely we'll stop people from using the terms "private" and "public" for servers. But it is what it is. A Directory can be public or private, so adding "(public") isn't correct here I don't think.

wiki/en/Running-a-Server.md
gilgongo: This might sort of clash with the FAQ entry on this, which says that some servers on the public Directories are sort of private.

wiki/en/Running-a-Server.md
pljones: It might be simpler to strip _all_ mention of running as a Directory from the documentation except for the definition list and from there have a link to a single page that explains how to set one up both from the command line (the normal thing you'd do) and from the GUI.

Almost no one wants to know.

wiki/en/Running-a-Server.md
pljones: I wuz wrong.

wiki/en/Running-a-Server.md
gilgongo: Again, the use of the term "public" there might be misleading because you might be connecting your server to a private directory (eg on a LAN). This kind of "correctness" is the point of this PR.

wiki/en/Running-a-Server.md
ann0see: > In general, if somebody sets up a public server, they accept that anyone can play on it. Jamulus has no password protection or other authentication mechanisms. However, some servers may state their policies in the welcome message you will see in the chat window.

Implies that mostly all servers registered at a public directory are believed to be freely available

wiki/en/Running-a-Server.md
gilgongo: I think attempts to resurrect the ideas of public and private may not be right. But let's see. You can run an unregistered server and make it public by putting its address out in some other way, for instance.

wiki/en/Running-a-Server.md
pljones: Why?

First off, most _servers_ are run headless, as far as I know.

No one wants to run as a Directory apart from a handful of people.

1. Most people running a server register with a Public Directory (Jamulus built in address) - they pick a Genre.

2. Those not wanting to register at all pick None.

3. The next most common group is those who are registering with a Custom Directory - they select Custom here (and put the Custom Directory server address in the "Custom Directory server address" field).

4. The least common group is those who are running Directories (both the Public and Custom ones -- same deal).  Of these, a tiny fraction will want to do it using the GUI (so a tiny fraction of a tiny fraction).
Just like from the command line, they use a Custom Directory server address of "localhost" or "127.0.0.1" -- because that's how we identify running as a Directory -- and, just like any other GUI server operator they'll select Custom for Directory and enter the Custom Directory server address.

There's _no point_ making running as a GUI Directory a special case as it's so rare.

wiki/en/Server-Troubleshooting.md
gilgongo: I removed it because I really can't see anyone being confused about this. And it's going to be in the GUI now too.

wiki/en/Running-a-Server.md
ann0see: Yes. I think so too. Probably we'd also stress the unregistered server part in the docs.
Setting up a registered (public) server is probably no longer that useful. We have a bunch of free ones all over the world (which are probably all running via CLI by more experienced users).

wiki/en/Running-a-Server.md
pljones: But it's still wrong.  Is it okay to use bad English just because you use it all the time?

wiki/en/Running-a-Server.md
ann0see: You can, but then you know that you're publishing it. I think this discussion should go to https://github.com/jamulussoftware/jamulus/discussions/2228

wiki/en/Server-Linux.md
pljones: "Directory" -- A service from which a Client can request a server list or with which a Server can register to be listed.

"Public Directory" -- Any Genre 1, Any Genre 2, Any Genre 3, Genre Rock, Genre Jazz, Genre Classical/Folk, Genre Choral/Barbershop -- the Directories built into Jamulus.

"Custom Directory" -- another address (than the Public Directory addresses) entered into "Custom Directories:" in Advanced Settings on the Client or "Custom Directory server address" on the Server Options tab.
For the Client, the user can then select that Custom Directory as a Directory in the Connect window (shown as the address entered in Settings).
For the Server, they can then choose to register with that Directory by selecting Custom.
The main difference is the Client supports multiple Custom Directories and the Server only supports a single one (as it's unlikely to change where it registers).

To run as a Directory, Jamulus is "tricked" into registering with itself using the either string "localhost" or the string "127.0.0.1" (literals, no alternatives).  It then starts allowing other Servers to register and supplying a server list to those Clients requesting one.  The method is the same whether run headless or with the GUI.

wiki/en/Server-Troubleshooting.md
ann0see: >  I really can't see anyone being confused about this

True... Probably this is on the wrong page then. Didn't fully test the new UI yet, so it might be evident with the new UI.

wiki/en/FAQ.md
pljones: `(public)` adds nothing and probably confuses things.  It's got nothing to do with what's being discussed in the paragraph.

wiki/en/Running-a-Server.md
pljones: > your server is believed to be public and

You can set up a firewall rule that lets you register your Server with Any Genre 1 but (unless `--showallservers` is used) won't be listed in the Connect window.  So adding any extras like this only confuses matters.

I agree _in principle_ that anyone using the Public Directories _should_ be registering their server for public use -- but we don't require it.

wiki/en/Running-a-Server.md
pljones: > if you don't want your server to be public since

I think the simpler version ("because") is clearer and less confusing (without any change).

wiki/en/Running-a-Server.md
pljones: No.  I've raised it in many places.

wiki/en/Running-a-Server.md
pljones: >  This means your server will not be publically visible.

I wouldn't bother with the addition.  Simply registering doesn't mean others will see you.  You'd get into the discussion of different types of Directory here, which isn't wanted.

wiki/en/Server-Troubleshooting.md
pljones: Is it not covered elsewhere now?  Hard to tell with the changeset view...

pljones:
> Just opened [jamulussoftware/jamulus#2228](https://github.com/jamulussoftware/jamulus/discussions/2228) to get a concensus for a new term except for "private" servers. As I said, I'm not happy about "unregistered"

I'm happy enough with "not registered".  "Unregistered" is a bit _active_ sense - if you _register_ you become _registered_.  If you _unregister_ you become _unregistered_?  "Not registered" works there, too.  If you do nothing... well... you're just not registered...  There are, then, a few places where some rewording is needed to move from a slightly active voice to a more passive voice.

wiki/en/Server-Troubleshooting.md
gilgongo: In so far as the UI for this will now be self-evident (and I think probably always was for this really) I don't think we need to document it because we don't document how to use the UI, only what it means to use it.

wiki/en/Running-a-Server.md
gilgongo: Sorry, does your comment imply a change to the wording here or not?

wiki/en/Running-a-Server.md
gilgongo: Good idea. It's too confusing trying to keep it all together as different "states" of the server.

wiki/en/Running-a-Server.md
gilgongo: My point was simply one of basic usability, but if we take the view that it's rare enough not to matter (and we're putting the Directory docs into another place now anyway) then that's OK.

wiki/en/Running-a-Server.md
gilgongo: It's standard practice to explain the _effect_ of using a UI, not _how_ you use it (unless your UI is unusable). So if you say "You can set the server's name, city and country... " then that begs the question why and how, which is the more important thing to explain.

gilgongo:
OK so pending the outcome of https://github.com/jamulussoftware/jamulus/discussions/2228 vocabulary we go for, I've done all the changes save for the screenshots now.

wiki/en/Running-a-Server.md
pljones: Yes.  The text is wrong as it stands.

gilgongo: OK I've re-written this in the new Custom-Directories.md file.

ann0see:
Not sure what to do with this PR.

gilgongo:
I can try fixing the conflicts. Otherwise, I think it's OK. I think we agree that we can't prevent people from using the term "private", but if we try to be clear with our terms in the docs then that's all we can do.

gilgongo:
BTW  I'm going to do another PR to turn the structure of the server docs inside out so that they lead (after suitable intro) with running a headless Linux server. So I'd like to get this merged :-)

gilgongo:
> I can try fixing the conflicts. Otherwise, I think it's OK. I think we agree that we can't prevent people from using the term "private", but if we try to be clear with our terms in the docs then that's all we can do.

Actually, no I can't as I don't know what all those `_translator-files` and non-en things are doing.

ann0see:
Hmm. I think fixing the conflicts is a tricky one... @ignotus666 I think it makes sense to do it the "dirty" way i.e. just copy the changed files manually to a new branch

ignotus666:
Yeah, it seems that lots of the changes in files are stuff that has nothing to do with "server UI changes" (image links edited to use includes, etc.). There might still be conflicts in some files because small things may have got edited since this PR was created, but they'll be much easier to fix than what's here now. The .po and .inc file conflicts should go away anyway.

gilgongo:
OK so if I close this. What's the best way of working out what I changed? Maybe I can just remember what those files were. I think it was only a few.

ignotus666:
The list of changed files in this PR should tell you (just ignore all the. po and .inc files).

PR #679, opened 2022-01-12
Remove symlinks and update scripts

### LINKED ISSUE #678: [Bug] Wiki overview page shows link to pt data files
**URL**
https://jamulus.io/es/wiki/
https://jamulus.io/de/wiki/

etc. all have two bullet points linking to /es/wiki/pt/navigation.yml or similar. I haven't found the cause yet.

**Describe the bug**
Misleading "empty" bullet points on the overview page lead to broken links. Probably something is wrong in the compilation step.

ignotus666:
# Does this need translation?

- [ ] Yes 
- [x] No 

# Changes

Fixes #678

Removed symlinks to translated files in ./_data and ./_includes (symlinks to the EN version stay - they're necessary and don't cause trouble). When generating the target translated files the po4a script now creates these files directly in their separate locations instead of lumping them all together and relying on symlinks. I did leave the symlinks for 1-Index.html though as it's easier than having to insert the language code in each one.


_po4a-tools/po4a-create-all-targets.sh
```diff
@@ -71,6 +83,17 @@ process_with_po4a () {
 
         # Get source doc names and set target file names and dirs
         filename=$(basename "$doc" .$ext)
+
+        if [[ $filename == 'general' || $filename == 'navigation' ]] ; then
```
ann0see, line 87: Probably not perfect to hardcode the file names but we‘ll have to live with it.

ann0see (commented):
The zip seems ok. Thanks!

_po4a-tools/po4a-create-all-targets.sh
ignotus666: I know, I had to do that with the threshold exclusions too... but I can't think of a better way that keeps it simple.

ann0see (approved):

PR #684, opened 2022-01-23
Layout: Center small-width images properly

### LINKED ISSUE #672: Text alignment below images not matching
**URL**
https://jamulus.io/wiki/Getting-Started
https://jamulus.io/wiki/Software-Manual

**Describe the bug**
All images in the Getting Started page, include a figure caption. 
The first image in the Software-Manual page contains a figure caption.

These captions are not aligned with the image. This can be best seen at the images which are a narrow.
Here an example:

website-settings

It would be nicer to have the text to align with the image.

hoffie:
# Changes
Currently, small-width images with captions are left-aligned, while their figcaptions are centered. This looks strange, at least one wider screens (#672).
This PR changes `img`s to `inline-block` and adjusts `figure`s to `text-align` their (inline) content.
Tested on Firefox and Chromium.

- [ ] Yes 
- [x] No 

I have therefore opened this PR against the release branch.

## Related issues

Fixes #672

### DIFF
--- a/assets/css/wiki.css
+++ b/assets/css/wiki.css
@@ -42,8 +42,12 @@
 }
 
 
+figure {
+    text-align: center;
+}
+
 figure img {
-  display: block;
+  display: inline-block;
   border-radius: 5px;
 }
 


henkdegroot:
@hoffie @gilgongo , the text which was below the images is no longer shown. Not all images had text....now the images are in center but without the text below that would not be needed. Was that intentional?

_EDIT: Sorry was looking at the wrong page.....the setup page image look similar to the user manual_ Thanks for the quick action!

hoffie:
@henkdegroot Can you clarify your edit, is there still something broken?

henkdegroot:
@hoffie, sorry I when I was checking the update I was looking at the user manual page. This page only has a "caption" below the first image. The remaining images on this page do not have that. However some images are the same as the one on the getting started page....where I should have checked.

So all fixed! sorry....

PR #700, opened 2022-02-07
Document Mute Myself

ignotus666:
# Changes

Document the new Mute Myself MIDI command

- [x] Yes 
- [ ] No 

## Related issues
Documents #2334

Ideally should be translated before the release. @jamulussoftware/translators


wiki/en/Tips-Tricks-More.md
```diff
@@ -82,11 +82,11 @@ Here is the script:
...
+The volume fader, pan control and mute and solo buttons in the client's mixer window strips can be controlled using a MIDI controller by using the `--ctrlmidich` parameter (note: only available for use with macOS and Linux using Jamulus version 3.7.0 or higher, and on Windows using the Jamulus version with JACK support). To enable this feature, Jamulus must be launched with `--ctrlmidich`. There is one global MIDI channel parameter (1-16) and two parameters you can set for each item controlled: `offset` and `consecutive CC numbers`. Set the first parameter to the channel you want Jamulus to listen on (0 for all channels) and then specify the items you want to control (f = volume fader; p = pan; m = mute; s = solo; o = mute myself) with the offset (CC number to start from) and number of consecutive CC numbers. There is one exception that does not require establishing an offset which is the "Mute Myself" command - it only requires a single CC number as it is only applied to one's own audio stream. Take the following example:
```
pljones: > not require establishing an offset

That's the number that's given, so that's wrong.  The bit that's "not supplied" is the "consecutive CC numbers" (`*8`).

wiki/en/Include-Client-Commands.md
```diff
@@ -2,5 +2,5 @@
 - `--mutemyown`  Mute me in my personal mix ("headless" mode only)                                                      
 -  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`  
 -  `-j` or `--nojackconnect`  Disable auto JACK connections  
--  `--ctrlmidich`  MIDI controller channel to listen on, control number offset and consecutive CC numbers (channels). Format: `channel[;f*][;p*][;s*][;m*]` See [Tips & Tricks](Tips-Tricks-More#using-ctrlmidich-for-midi-controllers) 
+-  `--ctrlmidich`  MIDI controller channel to listen on, control number offset and consecutive CC numbers (channels) and Mute Myself CC number. Format: `channel[;f*][;p*][;s*][;m*][;o]` See [Tips & Tricks](Tips-Tricks-More#using-ctrlmidich-for-midi-controllers)
```
ignotus666, line 5: Not sure if the 'o' parameter should be followed by an asterisk. Can't remember if it's a wildcard or that the parameter is literally followed by an asterisk.

wiki/en/Tips-Tricks-More.md
```diff
@@ -82,11 +82,11 @@ Here is the script:
...
 
-`--ctrlmidich "1;f0*8;p16*8;s32*8;m48*8"`
+`--ctrlmidich "1;f0*8;p16*8;s32*8;m48*8;o64"`
 
-Here, Jamulus listens on MIDI channel 1. Volume fader CC numbers start at 0 and there are 8 of them (so end at CC number 7). Pan controls start at CC number 16 and end at 23; Solo 32 to 39 and Mute 48 to 55.
+Here, Jamulus listens on MIDI channel 1. Volume fader CC numbers start at 0 and there are 8 of them (so end at CC number 7). Pan controls start at CC number 16 and end at 23; Solo 32 to 39 and Mute 48 to 55. Mute Myself is enabled/disabled by CC number 64. Please note that for the functions controlled by buttons to work properly, your MIDI controller needs the buttons to be set to "toggle" mode.
```
pljones: It might be worth separating
> Please note that for the functions controlled by buttons to work properly, your MIDI controller needs the buttons to be set to "toggle" mode.

into its own paragraph explaining that the controller button should send one value when pressed to "turn on" on the control -- (`>= 64`) -- and a separate value to "turn off" the control -- (`< 64`) -- i.e. "toggle mode".  ("Read your controller's manual to find out how to set this.")

wiki/en/Include-Client-Commands.md
pljones: No, the `*` means "times" (as in multiplication).  So `*8` means the control is used 8 times.

So the format is `channel[;f*N][;p*M][;s*P][;m*Q][;o]`, where `N`/`M`/`P`/`Q` are the number of times that control is to be used.

wiki/en/Tips-Tricks-More.md
ignotus666: Gah! you're right. Will correct right now.

wiki/en/Include-Client-Commands.md
pljones: There's also an "old style" format that I don't remember...  So long as the documentation doesn't _clash_ with that, it's okay not to mention it...

ignotus666: Ok, thanks.

wiki/en/Tips-Tricks-More.md
ignotus666: Yes, probably. I'll do that.

wiki/en/Include-Client-Commands.md
pljones: And, just as a note, I think the order of the `f`, `p`, `s`, `m` and `o` doesn't matter... I've not checked that, though.

ignotus666: No, it doesn't. Should we mention that?

ignotus666: > There's also an "old style" format that I don't remember... So long as the documentation doesn't _clash_ with that, it's okay not to mention it...

I know I originally got that format from _somewhere_ - but just can't remember any more.

henkdegroot: It is documented in the soundbase.cpp file:

   // parse the server info string according to definition: there is
    // the legacy definition with just one or two numbers that only
    // provides a definition for the controller offset of the level
    // controllers (default 70 for the sake of Behringer X-Touch)
    // **[MIDI channel];[offset for level]**
 
Just in case you want to add it, though I would not include it.

ignotus666: I would leave the client-commands doc as it is - no need to clutter it when there is a link to a more detailed explanation. 

My initial doubt was about _how_ the command is presented in client-commands.md. I know what the command should look like when actually used, but there's obviously a reason why it's shown as e.g. `[;f*]` and not `[;f(offset)*(cons.ch.)]` or whatever. I just couldn't remember whether the `*` represented the multiplier or was a wildcard for everything after 'f' - in which case the new parameter, 'o', should have been followed by an `*`.

ignotus666:
If no one has any objections it would be nice to merge this so we can do the translations.

PR #701, opened 2022-02-08
Add iOS ipa installation guide

ann0see:
# Changes

Updates guide as soon as https://github.com/jamulussoftware/jamulus/pull/2355 is merged

- [x] Yes 
- [ ] No 

## Related issues
[See PR for ipa creation](https://github.com/jamulussoftware/jamulus/pull/2355)

Please comment on alternatives/wording/improvements.

@emlynmac could you please review (this is just for experienced users)


wiki/en/Installation-for-iOS.md
```diff
@@ -16,14 +16,26 @@ Make sure you've already read the [Getting Started](Getting-Started) page.
...
 
 
 ## Install for iOS (for experienced users only)
+**Note**: We provide an unsigned .ipa file which you need to sign and then install on your device. If you are not Jailbroken (and have AppSync installed), you might need additional software to install Jamulus. There are multiple ways to do that, but this guide only explains the most standard approach. Please note that we assume you already have a (free) Apple developer account. If you don't have one, create an Apple ID and [Sign up for a developer account](https://developer.apple.com/membercenter). We recommend to create one without two factor authentification and unrelated to your potential personal account. You will need to enter the password in non-Apple software.
+
+1. [Download Jamulus (.ipa) on a PC/Mac]({{ site.download_root_link }}{{ site.download_file_names.ios }}){:.button}
+2. Install a program to sideload .ipa files ([Sideloadly](https://sideloadly.io/), [AltStore](https://altstore.io/), [Cydia Impactor](http://www.cydiaimpactor.com/) are possible options). This guide will focus on Sideloadly, but other software should work similar (Search for "Sideload .ipa with [name]" if you use different software).
```
emlynmac: @ann0see I'm really against encouraging side-loading on iOS. While this scenario is possible, I'm not sure we should be 'recommending' it

ann0see: I see it as the only "unlimited" way to provide Jamulus (especially for users without a Mac). It *is* not perfect and inconvenient (7 days re-signing necessity). However we are independent from Apple and an unlimited number of users can install it.

TestFlight is an option but as far as I know limited? Also I think it’s more work on our side?
What about adding a disclaimer, shortening the instructions and in future work on publishing the App to TestFlight?

Why do you dislike sideloading? Is it because of a negative connotation (piracy)?

emlynmac: https://developer.apple.com/testflight/
10,000 users and up to 90 days before the build expires. Should be fine until we get proper releases through the App Store, I would have thought?

ann0see: I installed sonobus via TestFlight and they hit the maximum user count? Maybe provide both options but stress TestFlight?

Proper builds for the AppStore would be great – but we first need to get the UI fixed. Or do you think Apple would approve it as is?

ann0see: Ok. Removed specific recommendations and made the process a bit more general. I hope that’s ok for you.

PR #719, opened 2022-03-10
Do not suggest editing systemd file directly

ann0see:
# Changes

The systemd file might (or will) be overwritten by an update. Therefore, `systemctl edit jamulus-headless` is suggested.
@hoffie @softins are experts in this field, so they should review it.

- [x] Yes 
- [ ] No

### DIFF
--- a/wiki/en/Running-a-Server.md
+++ b/wiki/en/Running-a-Server.md
@@ -96,7 +96,7 @@ If you want to run a Server on a Raspberry Pi, you will need to [compile from so
 1. Update apt to make sure you have a current list of standard packages: `sudo apt update`
 1. Install the Jamulus package: `sudo apt install ./{{ site.download_file_names.deb-headless }}`
 1. Enable the headless Server process via systemd: `sudo systemctl enable jamulus-headless`
-1. Add your desired [command line options](Running-a-Server#command-line-options) to the `ExecStart` line in the systemd service file by running `systemctl edit jamulus-headless` (By default you will be running an Unregistered Server).
+1. Add your desired [command line options](Running-a-Server#command-line-options) to the `ExecStart` line in the systemd service file by running `sudo systemctl edit --full jamulus-headless` (By default you will be running an Unregistered Server).
 1. Reload the systemd files `sudo systemctl daemon-reload` and restart the headless Server: `sudo systemctl restart jamulus-headless`
 1. Check all is well with `systemctl status jamulus-headless` (hit `q` to get back to the command prompt).
 
--- a/wiki/en/Running-a-Server.md
+++ b/wiki/en/Running-a-Server.md
@@ -92,15 +92,13 @@ If you want to run a Server on a Raspberry Pi, you will need to [compile from so
 
 ### Installation
 
-1. Download the [latest headless .deb file]({{ site.download_root_link }}{{ site.download_file_names.deb-headless }})
+1. Download the latest headless .deb file: `wget {{ site.download_root_link }}{{ site.download_file_names.deb-headless }}`
 1. Update apt to make sure you have a current list of standard packages: `sudo apt update`
 1. Install the Jamulus package: `sudo apt install ./{{ site.download_file_names.deb-headless }}`
 1. Enable the headless Server process via systemd: `sudo systemctl enable jamulus-headless`
-1. Add your desired [command line options](Running-a-Server#command-line-options) to the `ExecStart` line in the systemd service file in `/lib/systemd/system/jamulus-headless.service` (By default you will be running an Unregistered Server).
+1. Add your desired [command line options](Running-a-Server#command-line-options) to the `ExecStart` line in the systemd service file by running `systemctl edit jamulus-headless` (By default you will be running an Unregistered Server).
 1. Reload the systemd files `sudo systemctl daemon-reload` and restart the headless Server: `sudo systemctl restart jamulus-headless`
 1. Check all is well with `systemctl status jamulus-headless` (hit `q` to get back to the command prompt).
-
-You may also be interested in downloading [this set of useful tools](https://github.com/jamulussoftware/jamulus/tree/master/tools) from the Jamulus repository (clone the Git repo and also call `git submodule update --init`).
 
 ### Configuration
 
--- a/wiki/en/Running-a-Server.md
+++ b/wiki/en/Running-a-Server.md
@@ -92,7 +92,7 @@ If you want to run a Server on a Raspberry Pi, you will need to [compile from so
 
 ### Installation
 
-1. Download the latest headless .deb file: `wget {{ site.download_root_link }}{{ site.download_file_names.deb-headless }}`
+1. Download the [latest headless .deb file]({{ site.download_root_link }}{{ site.download_file_names.deb-headless }})
 1. Update apt to make sure you have a current list of standard packages: `sudo apt update`
 1. Install the Jamulus package: `sudo apt install ./{{ site.download_file_names.deb-headless }}`
 1. Enable the headless Server process via systemd: `sudo systemctl enable jamulus-headless`


gilgongo:
It's not very obvious how you actually edit what you get in `systemctl edit` though. Do you put the `ExecStart` line between the comments? 

```
### Editing /etc/systemd/system/jamulus-headless.service.d/override.conf
### Anything between here and the comment below will become the new contents of the file



### Lines below this comment will be discarded

### /lib/systemd/system/jamulus-headless.service
# [Unit]
# Description=Jamulus headless server
# After=network.target
# StartLimitIntervalSec=0
# 
# [Service]
# Type=simple
# User=jamulus
# Group=nogroup
# NoNewPrivileges=true
# ProtectSystem=true
# ProtectHome=true
# Nice=-20
# IOSchedulingClass=realtime
# IOSchedulingPriority=0
# 
# #### Change this to publish this server, set genre, location and other parameters.
# #### See https://jamulus.io/wiki/Command-Line-Options ####
# ExecStart=/bin/sh -c 'exec /usr/bin/jamulus-headless -s -n'
# 
# 
# Restart=on-failure 
...
```
if I try that, I get:

`systemd[1]: /etc/systemd/system/jamulus-headless.service.d/override.conf:1: Assignment outside of section. Ignoring.`

So I assume I'd need to uncomment and paste all that stuff in between the comments (and edit the ExecStart line). Bit annoying.

hoffie:
@ann0see Thanks for bringing this up, this is rather important to get right.

> So I assume I'd need to uncomment and paste all that stuff in between the comments?

No. The override file has to have the same structure as a unit file, but only the parts it wants to change. More specifically, it needs the proper section header:
https://github.com/jamulussoftware/jamulus/blob/7c1c1e992af959ba56d28083307222cd2542277e/distributions/debian/jamulus-headless.service#L6-L19

The override file should therefore be:
```
[Service]
ExecStart=/bin/sh -c 'exec /usr/bin/jamulus-headless -s -n --whatever-the-user-wants-to-add'
```

We should probably outline the whole structure of the override in our docs (or link a more concrete example).

gilgongo:
If I add the `[Service]` bit, I get:

`systemd[1]: jamulus-headless.service: Service has more than one ExecStart= setting, which is only allowed for Type=oneshot services. Refusing.`

hoffie:
Umm, interesting, I wasn't aware of that. Seems like one needs to clean it first. Even less intuitive...

```
[Service]
ExecStart=
ExecStart=/bin/sh -c 'exec /usr/bin/jamulus-headless -s -n --whatever-the-user-wants-to-add'
```

gilgongo:
Now I get `systemd[1]: jamulus-headless.service: Service has no ExecStart=, ExecStop=, or SuccessAction=. Refusing.`

Ah, Linux...

hoffie:
:(
Based on this [post](https://stackoverflow.com/questions/68817332/why-use-execstart-with-no-value-before-another-execstart-new-value-in-a-syst/68818218#68818218) I would have expected it to work. I currently don't have access to a Debian machine and can't test in a container which doesn't run systemd.

ann0see:
The systemd file should probably read from a config file to avoid this.

gilgongo:
Is there a reason why we don't do the `apt` thing where if there's a config file difference it brings up the "keep yours/keep theirs/edit" menu?

Either way, I think @ann0see suggestion of a separate config seems a good idea to me. You'd not need to run `systemctl reload` either (or whatever that command is when you edit a unit file or its dropin)

softins:
It's probably better to suggest instead doing `sudo systemctl edit --full jamulus-headless`.

The `--full` takes a complete copy of the installed service file from `/lib/systemd/system/jamulus-headless.service` or `/usr/lib/systemd/system/jamulus-headless.service` (depending on the distro), gives it to you for editing, and saves it into `/etc/systemd/system/jamulus-headless.service`.

The latter file will them be used by `systemd` instead of the former, default one.

If the one in `/etc` already exists, then `systemctl edit --full` will load it for further editing, instead of looking in `/lib` or `/usr/lib`.

Installing a new `.deb` will only overwrite the service file in `/lib` or `/usr/lib`, and won't clobber the user's own copy in `/etc`.

softins:
> Either way, I think @ann0see suggestion of a separate config seems a good idea to me. You'd not need to run `systemctl reload` either (or whatever that command is when you edit a unit file or its dropin)

You would, because the way it would work is that the config file would set environment variables, which are read when `systemd` reads the service file and finds a line like `EnvironmentFile=-/etc/sysconfig/jamulus-headless` in it. The config file is read, setting environment variables, which are then referred to in the `ExecStart` line. That is in fact what I do on **jamminthejazz**. Something like this:

```
# Do not edit this command. Edit /etc/sysconfig/jamulus-headless instead.
ExecStart=/bin/sh -c 'exec /usr/local/bin/jamulus.sh --server --nogui --fastupdate $${OPTS} \
        --log "$${LOGFILE:-/var/log/jamulus/jamulus.log}" \
        $${MULTITHREADING:+--multithreading} \
        $${PORT:+--port "$$PORT"} \
        $${RECORDING:+--recording "$$RECORDING"} \
        $${CENTRALSERVER:+--centralserver "$$CENTRALSERVER"} \
        $${SERVERINFO:+--serverinfo "$$SERVERINFO"} \
        $${WELCOME:+--welcomemessage "$$WELCOME"} \
        --numchannels $${CHANNELS:-10}'
```

But if I edit the config file, I still need to do `systemctl restart jamulus-headless` to make it re-read the file and restart with the new values.

But this is out of the scope of this particular PR, which I think should just promote and explain `sudo systemctl edit --full jamulus-headless`

hoffie:
> ```
> # Do not edit this command. Edit /etc/sysconfig/jamulus-headless instead.
> ExecStart=/bin/sh -c 'exec /usr/local/bin/jamulus.sh --server --nogui --fastupdate $${OPTS} \
> [...]
> ```
> 
> But if I edit the config file, I still need to do `systemctl restart jamulus-headless` to make it re-read the file and restart with the new values.

If you use a shell like in this example, we could also source a shell environment file. Changes to such a file should indeed become active by a simple `restart` and should not require a reload. E.g. something like this:
`ExecStart=/bin/sh -c 'source /etc/default/jamulus-headless; /usr/bin/jamulus-headless ... ${JAMULUS_EXTRA_ARGS:-}'`

wiki/en/Running-a-Server.md
```diff
@@ -92,16 +92,14 @@ If you want to run a Server on a Raspberry Pi, you will need to [compile from so
 
 ### Installation
 
-1. Download the [latest headless .deb file]({{ site.download_root_link }}{{ site.download_file_names.deb-headless }})
+1. Download the latest headless .deb file: `wget {{ site.download_root_link }}{{ site.download_file_names.deb-headless }}`
 1. Update apt to make sure you have a current list of standard packages: `sudo apt update`
 1. Install the Jamulus package: `sudo apt install ./{{ site.download_file_names.deb-headless }}`
 1. Enable the headless Server process via systemd: `sudo systemctl enable jamulus-headless`
-1. Add your desired [command line options](Running-a-Server#command-line-options) to the `ExecStart` line in the systemd service file in `/lib/systemd/system/jamulus-headless.service` (By default you will be running an Unregistered Server).
+1. Add your desired [command line options](Running-a-Server#command-line-options) to the `ExecStart` line in the systemd service file by running `systemctl edit jamulus-headless` (By default you will be running an Unregistered Server).
```
ann0see:
```suggestion
1. Add your desired [command line options](Running-a-Server#command-line-options) to the `ExecStart` line in the systemd service file by running `sudo systemctl edit --full jamulus-headless` (By default you will be running an Unregistered Server).
```

ann0see (commented):
@softins see your suggestion below.

wiki/en/Running-a-Server.md
gilgongo: That works for me. Don't need to `sudo systemctl daemon-reload` either after you do the edit either as far as I can tell. Just edit, `systemctl restart jamulus-headless` and it's fine.

wiki/en/Running-a-Server.md
```diff
@@ -92,16 +92,14 @@ If you want to run a Server on a Raspberry Pi, you will need to [compile from so
...
 1. Reload the systemd files `sudo systemctl daemon-reload` and restart the headless Server: `sudo systemctl restart jamulus-headless`
```
gilgongo, line 100:
```suggestion
1. Restart the headless Server: `sudo systemctl restart jamulus-headless`
```

At least, I didn't need to `daemon-reload` it.

wiki/en/Running-a-Server.md
```diff
@@ -92,16 +92,14 @@ If you want to run a Server on a Raspberry Pi, you will need to [compile from so
 
 ### Installation
 
-1. Download the [latest headless .deb file]({{ site.download_root_link }}{{ site.download_file_names.deb-headless }})
+1. Download the latest headless .deb file: `wget {{ site.download_root_link }}{{ site.download_file_names.deb-headless }}`
```
gilgongo: Is wget always installed? I thought some distros have curl instead (which is seen as more modern?)

ann0see: Yes. I googled it and it seems as if Ubuntu has wget by default while others have curl

gilgongo: But Ubuntu has curl as well, so maybe we should specify using that. BTW the reason I originally had it as just "download" was to avoid getting sucked into tools and method arguments like this, but I agree it's hard to know where to draw the line.

ann0see: Can we assume a user knows about how to download the deb? I think that’s the main reason to specify or not to specify a command here.

gilgongo: Well, we don't say on the Windows or iOS client pages "Use your web browser to download Jamulus...", so I assume that somebody interested in installing a headless server will know what they're doing to "acquire" the binary for the machine they wish to run it on.  After all, there's no mention of wget here: https://httpd.apache.org/download.cgi :-) 

My instinct here is to steer clear of anything that isn't strictly needed to be known in terms of using Jamulus though.

ann0see: Ok. So revert it:

```suggestion
1. Download the [latest headless .deb file]({{ site.download_root_link }}{{ site.download_file_names.deb-headless }})
```

wiki/en/Running-a-Server.md
ann0see: I think it’s cleaner to do it this way

gilgongo: Yes, so no separate "config" needed as systemd effectively creates a local config with `edit`. And if the user upgrades, we don't have to worry about overwriting their stuff.

ann0see (commented):
Great.

ann0see:
@gilgongo can we merge this?

gilgongo:
Yes - will do.

PR #737, opened 2022-04-16
Home: Highlight download links with a small border

hoffie:

This adjusts the styling of the header's download links to
- add a light bottom border for non-hovered links to make it obvious that they are clickable
- remove the text-underlining on hover as it would lead to two lines
- highlight the border on hover

It's hard to see that the the platform names are clickable. I'm open to other suggestions, but this sounded like the most-straightfoward solution for now.

Ready.

Reviews.

No.

## Checklist

- [x] I'm sure that this Pull Request goes to the correct branch: pretty much... :)

### DIFF
--- a/assets/css/home.css
+++ b/assets/css/home.css
@@ -32,6 +32,20 @@ header {
   color: #ececec;
 }
 
+#quick_dl_container {
+  margin: .5em 0 0 0;
+}
+
+#quick_dl_container a {
+  color: #ffffff;
+  text-decoration: none;
+  border-bottom: solid 1px #83cae5;
+}
+
+#quick_dl_container a:hover {
+  border-bottom-color: #fff;
+}
+
 @media (prefers-color-scheme: dark) {
   #get_started_btn_top {
     color: #dbdbdb;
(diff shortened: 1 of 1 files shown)


ann0see:
Initially they were kind of hidden due to the reason that we wanted people to first read the specific install guides (Windows needs ASIO etc). I think they're ok, but still dark mode looks a bit strange (contrast?)

grafik

hoffie:
> Initially they were kind of hidden due to the reason that we wanted people to first read the specific install guides (Windows needs ASIO etc).

Understood, but we also have to consider existing users which are only looking for the latest version for updating.

> I think they're ok, but still dark mode looks a bit strange (contrast?)

I've added a rule for dark mode to match the regular link color. Not sure if I like it, but this way it's consistent at least and I don't want to introduce yet another color. :)

I've also rebased.

assets/css/home.css
```diff
@@ -32,28 +32,32 @@ header {
...
+    color: #dbdbdb;
+    border-color: #dbdbdb;
+  }
+
+  .mainsite header h1, .mainsite, header a {
+    color: #dbdbdb;
+  }
+  #quick_dl_container a {
+    border-bottom-color: #3daed2;
+  }
```
hoffie, line 60: Had to move this block around in order for it to match.

ann0see (approved):
Ok. Tested and I think it's ok like that.

PR #741, opened 2022-04-23
Add JACK version for Windows

### LINKED ISSUE #739: Link JACK build on Windows page
I think we should link the JACK download on the windows page too (but less prominent without a button) on the Windows install page. 
Any disadvantages?

ann0see:
Adds JACK version to Windows page

Fixes: #739 

Ready for review

Needs review

YES

## Checklist

### DIFF
--- a/wiki/en/Installation-for-Windows.md
+++ b/wiki/en/Installation-for-Windows.md
@@ -22,9 +22,13 @@ Upgrading? You may want to [back up your configuration](Software-Manual#backing-
 
 _Please note that you are not permitted to redistribute this binary without acquiring a [licence agreement from Steinberg](https://www.steinberg.net/developers/)._
 
+If you don't want to use ASIO or prefer [JACK on Windows](https://jackaudio.org/faq/jack_on_windows.html), you can also use the [JACK version of Jamulus]({{ site.download_root_link }}{{ site.download_file_names.windows-jack }}).
+
 ***
 
-## Jamulus uses [ASIO](https://en.wikipedia.org/wiki/Audio_Stream_Input/Output) 
+## ASIO
+
+Jamulus can use [ASIO](https://en.wikipedia.org/wiki/Audio_Stream_Input/Output).
 
 * If you have an external sound card/audio interface, use its official ASIO driver if you can (they usually provide the best quality).
 * If you don't have an external sound card, you will probably not have an ASIO driver so will need to install a generic one like ASIO4ALL:


wiki/en/Installation-for-Windows.md
```diff
@@ -22,9 +22,13 @@ Upgrading? You may want to [back up your configuration](Software-Manual#backing-
 
 _Please note that you are not permitted to redistribute this binary without acquiring a [licence agreement from Steinberg](https://www.steinberg.net/developers/)._
 
+If you don't want to use ASIO, you can also use the (JACK version described below)(#JACK).
+
 ***
 
-## Jamulus uses [ASIO](https://en.wikipedia.org/wiki/Audio_Stream_Input/Output) 
+## ASIO
```
ann0see, line 29: @gilgongo since we link ASIO in the app, we need to have an anchor with the id ASIO on this page.

wiki/en/Installation-for-Windows.md
```diff
@@ -73,6 +77,15 @@ Have a look at [this video](https://youtu.be/_GzOsitVgLI) by [trombonepizza](htt
...
 
+## JACK
+
+Similar to the Linux version, Jamulus can also use [JACK](https://jackaudio.org/faq/jack_on_windows.html) on Windows. To get started with the JACK version:
+
+1. **Download and install JACK for Windows**. Get it [from the jackaudio page](https://jackaudio.org/downloads/).
+1. [Download Jamulus (JACK version)]({{ site.download_root_link }}{{ site.download_file_names.windows-jack }}){: .button}
+1. Now install Jamulus similar to the Windows ASIO version as mentioned above.
+1. Afterwards, to set up JACK, follow the [Linux setup guide](Installation-for-Linux).
+
```
ann0see: We could also move this somewhere else too, but I think it's better to put it at the bottom.

wiki/en/Installation-for-Windows.md
gilgongo: Sorry what do you mean? We have the ASIO rights statement in the site footer now if that's what you mean?

ann0see: No. The Windows installer links to the headline ASIO via a HTML ID (anchor). So that means we must not name the headline anything other than ASIO.

gilgongo: Oh OK sure.

wiki/en/Installation-for-Windows.md
gilgongo: Is there anything we can say about either the JACK or ASIO versions to help someone who might not know which to choose? Could we say ASIO is the preferred version but that JACK may allow more advanced use in some cases? I'm not too sure about the differences myself.

pljones: If the user's soundcard ASIO driver is up to scratch and the only ASIO app they use is Jamulus, then there's no reason to use anything but the ASIO version of Jamulus.

If they use other ASIO applications, they have a number of choices: Jack for Windows supports I/O using ASIO, which allows it to be used with other ASIO application without JACK support.  So you could choose to use the ASIO version of Jamulus with JACK anyway.  Or you could use the JACK-native version of Jamulus, which, if you're going to use JACK at all, probably makes more sense.

If the user's soundcard ASIO driver _isn't_ up to scratch, they might find JACK a better option than ASIO4ALL.  There's probably still more help available on the Web regarding ASIO4ALL than JACK for Windows, of course...  So, unless the user is reasonably happy working out how to set up something like ASIO4ALL or JACK for Windows, ASIO4ALL is likely the better choice (their problems will find more help in fixing them).  Where they're happy trying anything, JACK might open up opportunities they'd not thought to try, though.

In simple terms: if you're likely to want to link multiple applications together, JACK _might_ be a good alternative.

gilgongo: > if you're likely to want to link multiple applications together, JACK might be a good alternative.

So can we assume 99% of all Windows users have never heard of JACK and don't need to know about it? And the remaining 1% probably know about as part of some specialised thing they're doing, independent of their use of Jamulus. 

In that case, I'd just have a straight download link to the app and cut out the JACK download steps and stuff because that's going to be redundant and draws us into supporting JACK issues. 

Instead, I'd just have a line under the upgrading statement like this:

"If you use [JACK](https://jackaudio.org/) on Windows, we have a version that [supports JACK](link)"

and then the rest of the page is about ASIO. Otherwise it just clogs up the page with stuff that virtually nobody is going to want to know.

ann0see: Yes. Probably. However I know some sites proposing that JACK works better than ASIO4ALL due to Port audio support. But I can't verify that JACK works better.

gilgongo: Be that as it may, the JACK version is only really relevant if you know about JACK already. Inviting people to "get started" with JACK will get us into trouble if people think its some easy alternative to all the ASIO stuff they need to read though on that page. I'd just put it in as a link and be done with it.

ann0see: @gilgongo changed.

pljones (approved):

PR #772, opened 2022-07-10
Screenshots and updates

gilgongo:
Fixes:

https://github.com/jamulussoftware/jamulus/pull/2640 - sound alert 
https://github.com/jamulussoftware/jamulus/pull/2199 - screenshot (also 2424)

(I looked at https://github.com/jamulussoftware/jamulus/pull/2673 but the website seems fine)

### DIFF
--- a/wiki/en/Software-Manual.md
+++ b/wiki/en/Software-Manual.md
@@ -146,6 +146,10 @@ This changes the visual style of the audio meters, independently of the chosen s
 
 This sets the number of rows displayed in the server audio mixer, for use with larger ensembles.
 
+### Audio Alerts
+
+This turns on a sound alert for when someone joins or when receiving a new chat message.
+
 ## Audio/Network Setup
 
 <figure><img src="{% include img/en-screenshots/settings-network.inc %}" style="border: 5px solid grey;" loading="lazy" alt="Image of audio/network settings window"></figure>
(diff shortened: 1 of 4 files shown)


ann0see (approved):

wiki/en/Software-Manual.md
```diff
@@ -146,6 +146,10 @@ This changes the visual style of the audio meters, independently of the chosen s
 
 This sets the number of rows displayed in the server audio mixer, for use with larger ensembles.
 
+### Audio Alerts
+
+This turns on a sound alert for when someone joins a chat, or when receiving a new chat message.
```
ann0see: It also gives a message if someone joins and doesn't send a chat

gilgongo: I think that's implied, no?

ignotus666: I think he means that there's an alert when someone joins a server, not a chat. Then there's another alert for new chat messages.

gilgongo: Oh - I didn't realise the audio alert was for server joining too. I better amend that wording then.

gilgongo: So the audio alert plays when EITHER somebody joins a Server, OR when somebody sends a new chat message, is that right?

ignotus666: As far as I know, yes, with a different sound for each event.

ann0see (approved):

PR #815, opened 2022-09-07
Use configured Directory Public IP for Client when hole-punching

### LINKED ISSUE #812: Directory: Use configured Directory Public IP for Client when hole-punching
Following up on https://github.com/jamulussoftware/jamulus/pull/2819#issuecomment-1232115690 so it doesn't get lost.

https://jamulus.io/wiki/Custom-Directories (to be renamed, IIRC) should have an additional note added under "Points to note about Directories":

* When using a client on the same LAN as a directory, the directory itself needs to be run with the correct `--serverpublicip` value, so that any (non-LAN) servers registering with the directory that require the "hole punch" can be accessed by the LAN client(s).  This is because otherwise the directory would provide the LAN address of the client to the server and the "hole punch" would fail.

Does that sound right?

pljones:

Adds the hint for "hole punch" for anyone running a directory and clients on the same LAN, along with public servers.

Fixes: #812 

Also adjusted a couple of the other hints and tried to use the appropriate wording here (i.e. not exactly what I'd said in 812).

Review: 
Old: https://jamulus.io/wiki/Custom-Directories
New: http://jamuluswebsite.drealm.info/wiki/Directories

Yes.

## Checklist
- [x] I'm sure that this Pull Request goes to the correct branch (*this* time...)


hoffie (approved):
LGTM.

This describes some real edge case and may be hard to understand for the average user, but I don't think it can be described in any simpler ways as the scenario is... special. :)

PR #829, opened 2022-09-24
Therefore I Declared War On Simply

mcfnord:
The word simply can be unhelpful, and is often unneeded. Who are you to say what's simple for me? I've attempted to maintain the .po files, but it's a little hard yet to be sure this is right and complete. I'm trying to determine if there's too much friction around small changes to bother.

### DIFF
--- a/_posts/2020-03-28-Server-Rpi.md
+++ b/_posts/2020-03-28-Server-Rpi.md
@@ -11,7 +11,7 @@ This guide shows you how to install Jamulus on a Raspberry Pi.
 
 ## Server Installation - Raspberry Pi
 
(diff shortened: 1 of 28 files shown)


ignotus666:
@mcfnord you don't need to edit any .po files. Changes to EN are done to the .md files in `wiki/en` and those changes are propagated to the .po files (and Weblate) automatically when the PR gets merged.

I see at least one change that I think is also in another of your PRs (the "or" separating downloads). I'm not sure if it's better to do a separate PR for each doc or a separate commit for each within a single PR, but things are getting a bit scattered over several PRs with overlaps and lots of docs involved, making it difficult to review it all. Maybe start with a PR for what are obvious mistakes (those shouldn't have a problem with being accepted and merged), and then move on to more stylistic changes that might need some discussion.

PR #831, opened 2022-09-25
re-ordering for clarity

mcfnord:
re-orders to assure content matches title, conflates duplication for GRP into one section

### DIFF
--- a/wiki/en/Software-Manual.md
+++ b/wiki/en/Software-Manual.md
@@ -101,7 +101,7 @@ If you have set your Audio Channel to Stereo or Stereo Out in your Settings, you
 
 If you see a "mute" icon above a user, it means that person cannot hear you. Either they have muted you, soloed one or more users not including you, or have set your fader in their mix to zero.
 
-Users are listed left to right in the order that they connect. You can change the sort order using the Edit option in the application menu.
+Users usually appear left-to-right in the order that they connect. You can sort instead by name, instrument, group, or city using the View menu.
 
 ### Grp button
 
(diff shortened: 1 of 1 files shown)


wiki/en/Software-Manual.md
```diff
@@ -101,23 +101,25 @@ If you have set your Audio Channel to Stereo or Stereo Out in your Settings, you
 
 If you see a "mute" icon above a user, it means that person cannot hear you. Either they have muted you, soloed one or more users not including you, or have set your fader in their mix to zero.
 
+Users are listed left to right in the order that they connect. You can change the sort order using the Edit option in the application menu.
```
ann0see: I don't think that's necessarily true all the times (just most of the times) since old server slots might get filled by newcomers. Although it's mostly ok, I'd rather not remove that.

ann0see (commented):
Thanks!

pljones (approved):

wiki/en/Software-Manual.md
mcfnord: I could add "Users are _usually_..." if you think that's accurate. Hmmm, what did I remove? I noticed a duplication, about GRP, but removed no claims. What got removed?

ann0see: Ah sorry. I edited the comment in a wrong way. It should say that I'd rather remove this paragraph.

mcfnord: It does reveal the re-ordering feature. But it says it's on the Edit menu, when in fact it appears to be on the View menu. 

"Users usually appear left-to-right in the order that they connect. You can sort instead by name, instrument, group, or city using the View menu."

This kind of page is typically rather comprehensive.

ann0see:
@gilgongo could you please review this PR?

Issue #817, opened 2022-09-09
Update po4a for better markdown support


ignotus666:
Upstream po4a has received several updates recently that improve markdown support. Our scripts currently process .md files as asciidoc and work well enough, but this can be improved. I've been doing some tests on my repo and would like to implement these improvements in our scripts (and update the po4a .deb package in the assets repo). 

These improvements are mainly about stripping out markdown formatting like # (translators don't have to worry about how many there are), ``` (identifying code blocks and treating them as such) and '-' symbols from .po files and some other things. After a .po file update there will be a lot of translated strings that will be marked as fuzzies though, but I can fix those myself.

So to do this update I would like to ask that after merging the translation updates currently on Weblate, we lock it temporarily to avoid conflicts while it gets done. I hope to have time next Monday 12 September during the morning - it should only take a couple of hours. This update should in future reduce the chances of errors made with markdown formatting and save translators from having to worry about it.

pljones:
Can this wait until 3.10.0?

ignotus666:
I don't see why not. The thing is, with weblate translations are coming in constantly, not like before when there were defined pre-release periods. Is there any specific reason you'd prefer to wait?

pljones:
Stability during the release cycle, is really the only thing.  3.9.1 is intended to be out "ASAP" as we don't want anything risking that, like suddenly finding the tools don't work.  If it's something that can be done, tested and proved working without impacting what we have now, then I guess that should be fine.

ignotus666:
Fair enough, it's not urgent so no problem to wait until after 3.9.1. I've been doing testing on my repo but there's always the chance I missed something.

ann0see:
@ignotus666 Any updates here?

ignotus666:
Looking into it again. As it involves changes to po4a scripts, workflows, an updated version of po4a and .po files, it has to be done in a series of PRs and in a certain order.

ann0see:
Closed as the respective PR was merged.

PR #840, opened 2022-10-07
Remove unneeded trailing slash of void elements

ann0see:

HTML void elements should not use trailing slashes:  https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#trailing-slashes

This removes (hopefully) all of them
Makes w3c validator happy

Ready for review.

Review

No.

## Checklist

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -2,19 +2,19 @@
   <div class="fx-col-100-xs fx-col-33-s fx-col-center-xs">
     <a href="https://github.com/jamulussoftware/jamulus" target="_blank" rel="noreferrer">
       <!-- https://icongr.am/fontawesome/github.svg?size=128&color=ffffff -->
-      <img src="{{ '/assets/img/github.svg'| relative_url }}" alt="{{ site.data.general.footer.alt.github }}" loading="lazy" class="fx-center" />
+      <img src="{{ '/assets/img/github.svg'| relative_url }}" alt="{{ site.data.general.footer.alt.github }}" loading="lazy" class="fx-center">
     </a>
   </div>
   <div class="fx-col-100-xs fx-col-33-s fx-col-center-xs">
     <!-- https://icongr.am/fontawesome/facebook-official.svg?size=128&color=ffffff -->
     <a href="https://www.facebook.com/groups/619274602254947/" target="_blank" rel="noreferrer">
-      <img src="{{ '/assets/img/facebook.svg' | relative_url }}" alt="{{ site.data.general.footer.alt.facebook }}" id="fb_icon" loading="lazy" class="fx-center" />
+      <img src="{{ '/assets/img/facebook.svg' | relative_url }}" alt="{{ site.data.general.footer.alt.facebook }}" id="fb_icon" loading="lazy" class="fx-center">
     </a>
   </div>
   <div class="fx-col-100-xs fx-col-33-s fx-col-center-xs">
     <!-- https://icongr.am/fontawesome/comments.svg?size=128&color=ffffff -->
     <a href="https://github.com/jamulussoftware/jamulus/discussions" target="_blank" rel="noreferrer">
-      <img src="{{ '/assets/img/forum.svg' | relative_url }}" alt="{{ site.data.general.footer.alt.help }}" loading="lazy" class="fx-center" />
+      <img src="{{ '/assets/img/forum.svg' | relative_url }}" alt="{{ site.data.general.footer.alt.help }}" loading="lazy" class="fx-center">
     </a>
   </div>
   <div id="copyright" class="fx-col-100-xs fx-row-center-xs">
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -1,6 +1,6 @@
 <header class="fx-row fx-row-between-xs fx-text-center">
   <div class="fx-col fx-text-center"><a href="/{% if site.active_lang != site.default_lang %}{{ site.active_lang }}/{% endif %}">
-      <img alt="{{ site.data.general.nav.altJamulusIcon }}" width="40" height="40" src="{{ 'assets/img/jamulus-icon-2020.svg'| relative_url }}" /></a>
+      <img alt="{{ site.data.general.nav.altJamulusIcon }}" width="40" height="40" src="{{ 'assets/img/jamulus-icon-2020.svg'| relative_url }}"></a>
   </div>
   <div class="fx-col-center-xs fx-txt-right hide-on-dt">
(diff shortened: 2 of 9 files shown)


ann0see:
Po4a doesn't seem to like removing "/"? Checks fail.

pljones:
See https://www.thoughtco.com/why-different-html-versions-3471349

ann0see:
I thought we're using HTML5?

pljones:
> I thought we're using HTML5?

At a guess, do files declare that with a doctype?  If we're leaving parsers to guess, it's easy to guess wrongly.

ann0see:
We do have a doctype html tag, so it should be interpreted as html5.

pljones:
In `Update .po files` I can see:
```
../wiki/en/misc/1-index.html:48: (po4a::xml)
Unexpected closing tag </p> found. The main document may be wrong.
```
when updating the language files.  Does that mean po4a thinks it's using `xml` style tags?  That's also reported in `Create translated docs and stats`.  I'd check for an _extra_ `</p>` where one isn't needed based on the error report.

----

It looks like that then has a knock on effect as, later, in `Build the site in the jekyll/builder container`:
```
Installing jekyll-minifier 0.1.10
Bundle complete! 4 Gemfile dependencies, 37 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [x86_64-linux-musl]
Configuration file: /srv/jekyll/_config.yml
            Source: /srv/jekyll
       Destination: /srv/jekyll/_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
/usr/gem/gems/jekyll-4.2.2/lib/jekyll/utils.rb:141:in `initialize': No such file or directory @ rb_sysopen - /srv/jekyll/1-nl-index.html (Errno::ENOENT/usr/gem/gems/jekyll-4.2.2/lib/jekyll/utils.rb:141:in `initialize': No such file or directory @ rb_sysopen - /srv/jekyll/1-nl-index.html (Errno::ENOENT)
	from /usr/gem/gems/jekyll-4.2.2/lib/jekyll/utils.rb:141:in `open'Installing jekyll-minifier 0.1.10
Bundle complete! 4 Gemfile dependencies, 37 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
...
```
and then fails.

Then, `Zip Website` reports
```
  zip -r /home/runner/work/jamuluswebsite/jamuluswebsite/website.zip _site/*
  shell: /usr/bin/bash -e {0}
	zip warning: name not matched: _site/*

zip error: Nothing to do! (try: zip -r /home/runner/work/jamuluswebsite/jamuluswebsite/website.zip . -i _site/*)
Error: Process completed with exit code 12.
```

ann0see:
> Does that mean po4a thinks it's using xml style tags? 

Yes, that's what I suspect.

pljones:
Mm.  Nothing obvious like unbalanced tags...  The [document itelf](https://github.com/jamulussoftware/jamuluswebsite/blob/release/wiki/en/misc/1-index.html) doesn't have a `<doctype>` as it's a fragment.  Does po4a have a config that's fed in?

ann0see:
I searched for it but didn't find anything. 

@ignotus666 ?

ignotus666:
I can't really look into it properly yet until Monday but yes, html docs are processed as xml by po4a. I think the checks for tag "errors" can be disabled or left at just a warning (without aborting the file processing), as it seems to default to aborting. There's info about this in the official po4a documentation. It should be a matter of adding a config parameter to the scripts.

pljones:
I'd really like to leave the config until we identify what's changed to cause it to start breaking.

ignotus666:
[Here's](https://po4a.org/man/man3/Locale::Po4a::Xml.3pm.php) the relevant page in the po4a docs - scroll down to "ontagerror". It seems that po4a interprets that `<br>` tag as an opening tag and doesn't like that the next closing tag doesn't match it. Turns out the update I was preparing for #817 would "fix" this by just printing a warning (and not exiting). I don't know if that's an acceptable solution though.

On closer inspection, I was using "ontagerror", but maybe "optionalclosingtag" would be a better choice?

pljones:
Does HTML5 allow
```html
<p>paragraph one
<p>paragraph two without a previous closing &lt;/p&gt;</p>
```
?  Do we want to allow that?  It sounds like "ontagerror" would cause the error - but also errors on `<br>` rather than `<br/>` and "optionalclosingtag" would allow the "missing" `</p>` whilst allowing `<br>`.

ignotus666:
Another option is to use the "Xhtml" instead of "xml" module in po4a. It correctly deals with those tags without adding options but segments the paragraphs in a completely different way, meaning translators would have to re-translate "1-Index.html", though it seems a re-write is in the works so maybe not such a big deal.

[Here's](https://github.com/mquinson/po4a/issues/179) an issue raised about the same problem that arose with the Xhtml module, but which seems to have been solved since.

_includes/general/navbtn.html
```diff
@@ -1,8 +1,8 @@
 <svg role="img" width="20" height="20" aria-labelledby="ntitle" id="navbtn" class="hide-on-dt">
   <title id="ntitle">{{ site.data.general.nav.btnOpenNavigation }}</title>
   <g>
-    <rect fill="currentcolor" id="r1" width="100%" height="3" x="0" y="0" />
-    <rect fill="currentcolor" id="r2" width="100%" height="3" x="0" y="8" />
-    <rect fill="currentcolor" id="r3" width="100%" height="3" x="0" y="16" />
+    <rect fill="currentcolor" id="r1" width="100%" height="3" x="0" y="0">
```
ann0see: Firefox mobile dev mode doesn't seem to like this change. The last two hamburger rectangles disappear.  SVG's need self closing tags.

ann0see: I think I fixed it now.

ann0see (commented):
I think all is ok now after the update by ignotus.

ignotus666:
There's something wrong - all the file updates with po4a fail, but the script doesn't exit and the check looks like it passed when it didn't. Apparently`libs-parse-keyword-perl` needs to be installed. I'm going to add it to the cache script, and change the key back to "po4a" to force a new cache to be created. Then re-run this action to see if it's fixed.

ignotus666:
Ok, fixed now.

ignotus666 (approved):

ignotus666:
Aargh, this has caused Weblate to lock due to a whole load of conflicts...

PR #870, opened 2022-11-03
Translations update from Hosted Weblate

weblate:
Translations update from [Hosted Weblate](https://hosted.weblate.org) for [Jamulus/Startpage](https://hosted.weblate.org/projects/jamulus/1-index/).

It also includes the following components:

* [Jamulus/Installation-for-Android](https://hosted.weblate.org/projects/jamulus/installation-for-android/)

* [Jamulus/Getting-Started](https://hosted.weblate.org/projects/jamulus/getting-started/)

* [Jamulus/Contribution](https://hosted.weblate.org/projects/jamulus/contribution/)

* [Jamulus/Directories](https://hosted.weblate.org/projects/jamulus/directories/)

* [Jamulus/Installation-for-Macintosh](https://hosted.weblate.org/projects/jamulus/installation-for-macintosh/)

* [Jamulus/Client-Troubleshooting](https://hosted.weblate.org/projects/jamulus/client-troubleshooting/)

* [Jamulus/Installation-for-Windows](https://hosted.weblate.org/projects/jamulus/installation-for-windows/)

* [Jamulus/Installation-for-Linux](https://hosted.weblate.org/projects/jamulus/installation-for-linux/)

* [Jamulus/Include-Server-Commands](https://hosted.weblate.org/projects/jamulus/include-server-commands/)

* [Jamulus/Command-Line-Options](https://hosted.weblate.org/projects/jamulus/command-line-options/)

* [Jamulus/General strings](https://hosted.weblate.org/projects/jamulus/general/)

* [Jamulus/QOS-Windows](https://hosted.weblate.org/projects/jamulus/qos-windows/)

* [Jamulus/Server-Bandwidth](https://hosted.weblate.org/projects/jamulus/server-bandwidth/)

* [Jamulus/Installation-for-iOS](https://hosted.weblate.org/projects/jamulus/installation-for-ios/)

* [Jamulus/Privacy-Statement](https://hosted.weblate.org/projects/jamulus/privacy-statement/)

* [Jamulus/Include-Shared-Commands](https://hosted.weblate.org/projects/jamulus/include-shared-commands/)

* [Jamulus/Include-Backing-Up](https://hosted.weblate.org/projects/jamulus/include-backing-up/)

* [Jamulus/Include-Client-Commands](https://hosted.weblate.org/projects/jamulus/include-client-commands/)

* [Jamulus/Navigation](https://hosted.weblate.org/projects/jamulus/navigation/)

[...]

PR #873, opened 2022-11-16
Update po4a scripts for better markdown support 

### LINKED ISSUE #817: Update po4a for better markdown support

ignotus666:
Update po4a and workflow scripts to better support markdown, and display errors (but not fail) when there are tag inconsistencies in html files.
The po4a cache has been renamed to force the creation of a new one with the updated version of po4a in the assets repo. The updated po4a scripts will not work with the old version as new options have been introduced, so this PR and [this one](https://github.com/jamulussoftware/assets/pull/5) should be merged at the same time. In any event, I think #870 should be merged before this PR to avoid possible conflicts (e.g. it would happen if this PR gets merged and afterwards the EN files get updated).

The stats script has been removed as with Weblate it's become redundant.

The next time the .po files are updated, a number of strings in each language will become fuzzies - it will mostly just be a matter of removing characters (`#, ```` and maybe others) in the fuzzy strings.

Fixes #817

Should be ready.

No.

## Checklist


ignotus666:
Note: the check fails because it looks for po4a version 0.68, which isn't yet in the assets repo.

_po4a-tools/po4a-create-all-targets.sh
```diff
@@ -116,6 +119,8 @@ process_with_po4a () {
             --po "$PO_DIR/$lang/${filename}.po" \
             --localized "$targ_doc" \
             --localized-charset "UTF-8" \
+            --no-deprecation \
```
ignotus666, line 122: This is to prevent a stream of warning messages - we use po4a in a way that is considered deprecated, but which actually works better for our workflow - we want separation between .po file updates and their processing into translated files. The 'updated' way of using po4a combines both steps, which we don't want.

.github/workflows/add-lang.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         id: cache-po4a
         with:
           path: "~/po4a"
-          key: ${{ runner.os }}-po4a
+          key: ${{ runner.os }}-po4a-0.68
```
ann0see, line 51: You could use a hash like in the main repo to make the key unique

_po4a-tools/po4a-cache.sh
```diff
@@ -6,8 +6,8 @@
 if [[ "$CACHE_HIT" == 'true' ]] ; then
     sudo cp --force --recursive ~/po4a/* /
 else
-    sudo apt install -yq gettext libsgmls-perl libyaml-tiny-perl opensp
-    wget -O po4a.deb https://github.com/jamulussoftware/assets/raw/main/po4a/po4a_0.66.deb
+    sudo apt install -yq gettext libsgmls-perl libyaml-tiny-perl opensp libsyntax-keyword-try-perl
+    wget -O po4a.deb https://github.com/jamulussoftware/assets/raw/main/po4a/po4a_0.68.deb
```
ann0see, line 10: Could you please upload the deb?

_po4a-tools/po4a-create-all-targets.sh
ann0see: Hmm. But will we need to change the behaviour any time soon?

ignotus666: Not that I know of. In reality the "updated" way to use po4a is just an automated combined use of the scripts that we use separately.

.github/workflows/add-lang.yml
ignotus666: Uh, I had a look but don't have a clue how that works...

_po4a-tools/po4a-cache.sh
ignotus666: It's in a PR in the assets repo. As I said in the description, I think #870 should be merged first, then the PR in assets, then this one. If I upload it now, the workflows won't find it.

.github/workflows/add-lang.yml
ignotus666: The cache hasn't been used for a week, which means that any workflow run should now trigger the creation of a new one. So I could leave the key here as "po4a" and merge the new po4a version in the assets repo now. Does that sound ok @ann0see?

ann0see:
Could you please check the validity of this again? Locally I don't see a translation for any language.

ann0see:
Ok. Got it. Maybe the threshold is too high?

ignotus666:
The new markdown processing in this PR causes a lot of strings to become fuzzies, which is why the threshold isn't reached for so many files. They're easy to update, as it's just a matter of removing a few characters, so I'd leave the threshold as it is. It will take some effort to do though, as there are quite a few languages now. 
But in the long run we'll have "cleaner" strings for translators.

ignotus666:
Also, I think it would be better to get this in before all the changes to the EN files that are happening, so they get processed in the "new" way.

ann0see:
Yes, but this will make the production site largely untranslated

ignotus666:
Ahh, true... So maybe we can make this PR against next-release then?

ann0see:
We could. It's not perfect. Especially since this will make the build fail on release – unless we re-upload the old po4a deb and use different links.

ann0see:
Let’s merge this to next release.

.github/workflows/add-lang.yml
ann0see: Use ${{ hashFiles('file1, 'file2') }}... Will open an issue for that soon.

ann0see (approved):

PR #889, opened 2022-12-28
ask nicely

mcfnord:
This can be ignored if it's too controversial or not consequential enough. I don't claim English is a universal language.


pljones (approved):
Looks fair.

PR #920, opened 2023-04-18
Fix language tags in 1-index.html if incorrect

### LINKED ISSUE #883: Some Norwegian slipped into the main jamulus.io English website
**URL**
https://jamulus.io/

**Describe the bug**
Some Norwegian translations slipped into the main jamulus.io English website

grafik

### LINKED ISSUE #918: Fail CI build if Index file has wrong or no language tag set
The translation gets messed up if an error occurs on the main website's translation e.g if the language is not set correctly. I think the bash scripts should check that and fail on error. 

_related to https://github.com/jamulussoftware/jamuluswebsite/issues/883#issuecomment-1367169162_

ignotus666:
Updated po4a script to check if the language tag is correct in '1-index.html' and correct it if required. Avoids the problem where the website won't build properly if the language is mistyped or left as "en".

Fixes: #918

Should be ready.

Tested and works both if the language has been mistyped or left as "en". Does nothing if correct. Might be a more elegant way of doing it though.

NO.

## Checklist

### DIFF
--- a/_po4a-tools/po4a-create-all-targets.sh
+++ b/_po4a-tools/po4a-create-all-targets.sh
@@ -128,6 +128,14 @@ process_with_po4a () {
             echo "$filename.$ext" translated into "$lang"
         fi
 
+        # Check if language is set correctly in '1-$lang-index.html'
+        if [ $filename == '1-index' ] ; then
+            if ! grep -Fxq 'lang: "'$lang'"' "$WIKI_DIR/$lang/1-index.html" ; then
+                echo replacing incorrect language tag in 1-"$lang"-index.html;
+                sed -i 's/lang: "[^"]*"/lang: "'$lang'"/' "$WIKI_DIR/$lang/1-index.html"
+            fi
+        fi
+
     done <   <(find -L "$SRC_DIR" -name "*.*"  -print0)
 }
 
--- a/_po4a-tools/po4a-create-all-targets.sh
+++ b/_po4a-tools/po4a-create-all-targets.sh
@@ -132,7 +132,7 @@ process_with_po4a () {
         if [ $filename == '1-index' ] ; then
             if ! grep -Fxq 'lang: "'$lang'"' "$WIKI_DIR/$lang/1-index.html" ; then
                 echo replacing incorrect language tag in 1-"$lang"-index.html;
-                sed -i 's/lang: "[^"]*"/lang: "'$lang'"/' "$WIKI_DIR/$lang/1-index.html"
+                sed -i '0,/lang: "[^"]*"/s/lang: "[^"]*"/lang: "'$lang'"/' "$WIKI_DIR/$lang/1-index.html"
             fi
         fi
 


_po4a-tools/po4a-create-all-targets.sh
```diff
@@ -128,6 +128,14 @@ process_with_po4a () {
             echo "$filename.$ext" translated into "$lang"
         fi
 
+        # Check if language is set correctly in '1-$lang-index.html'
+        if [ $filename == '1-index' ] ; then
+            if ! grep -Fxq 'lang: "'$lang'"' "$WIKI_DIR/$lang/1-index.html" ; then
+                echo replacing incorrect language tag in 1-"$lang"-index.html;
```
ann0see, line 134: I'd make it more stupid probably and just fail the test with an error message. Otherwise we must ensure that we actually inferred the correct language and nobody fiddled around with file names etc.

ignotus666: Hmm, the only way to change a file name would be via a PR and would be soon spotted. The language tag it inserts is taken from `$lang` (in turn originally created by another script previously), so not much room for error.

The scenario where this keeps happening is when a new language is added and the translator either hasn't got round to translating `1-index.html` or doesn't spot the language tag because of how it's embedded within a large paragraph (not the case with all the other docs - it's a segment of its own). This just makes sure that the language tag will always be correct in `1-index.html` before deployment. It'll still be wrong in the .po file, but that won't have consequences.

pljones: This also implies that the only occurrence of the characters `lang: "` followed by the _correct_ value and `"` are in the right place.  If it's going to grep, it should be more anchored, I'd say.

ignotus666: Sorry, not sure I follow you - is it too vague/too specific/limiting/ …? I arrive at my `grep`, `sed` and other commands through an excruciating process of trial and error - I barely know what I'm doing, but enjoy the challenge of making it "work" ;-)

pljones: I mean it will find, for example:
```
The language tag for English, `lang: "en"` should be replaced with your language, for example `lang: "es"` or `lang: "ru"`.
```
or any other occurrence and replace the first match on the line.  Should someone ever mention `lang: "XX"` in the 1-index.html page, that is.  You should take more context to reduce the likelihood of a mismatch.  (I'm not suggesting this is _currently_ an issue - it's just something that'll be lurking to bite in the future, though.)

ignotus666: I see. I reckon the chances of the `lang: "XX"` pattern occuring again in that file are pretty remote, but no harm in protecting against that I suppose. I'll have a look into it when I get a chance.

ignotus666: Had a look at this. The grep command only returns the first matched pattern of `lang: "XX"` because of the `-x `option (only return a complete matching line). `-m 1` (stop at first match) could also be used with the same result. Any other occurrence of the pattern is ignored. So I think the `grep` command is fine as it is (maybe remove the `-q` option?)

What's replacing the first matched pattern in _all_ given lines is the `sed` command. This:
`sed -i '0,/lang: "[^"]*"/s/lang: "[^"]*"/lang: "'$lang'"/' "$WIKI_DIR/$lang/1-index.html"`
seems to solve it.
@pljones what do you think?

pljones: OK, that's safer, indeed.  Give it a go.  At least it won't be _as_ potentially disasterous ;).  (Reducing the possibility from 0.001% to 0.00001% chance...)

pljones (approved):

ann0see:
Merged due to approval.

PR #927, opened 2023-05-28
welcome path gotcha tip

mcfnord:
Help user locate path without providing specific guidance.

### DIFF
--- a/wiki/en/Running-a-Server.md
+++ b/wiki/en/Running-a-Server.md
@@ -172,6 +172,7 @@ Show an agreement window before users can connect.  The text of the agreement to
 
 ##### `-w or --welcomemessage`
 A "welcome message" to display in the Client chat window on connect. Can be given as a string or filename, and can contain HTML.
+When a path is used, the full path must be accessible by the `jamulus` Linux user. If not accessible, the literal path (rather than its contents) will appear.
 
 ##### `--serverpublicip`
 The public IP address of the Server if connecting to a Directory behind the same NAT. See [the Directories guide](Directories) for further information.
--- a/wiki/en/Running-a-Server.md
+++ b/wiki/en/Running-a-Server.md
@@ -172,6 +172,7 @@ Show an agreement window before users can connect.  The text of the agreement to
 
 ##### `-w or --welcomemessage`
 A "welcome message" to display in the Client chat window on connect. Can be given as a string or filename, and can contain HTML.
+When a path is used, the file must be accessible by the user account running Jamulus. (On most Linux installations, user `jamulus` is used by default.) If not accessible, the literal path (rather than its contents) will appear. 
 
 ##### `--serverpublicip`
 The public IP address of the Server if connecting to a Directory behind the same NAT. See [the Directories guide](Directories) for further information.
--- a/wiki/en/Running-a-Server.md
+++ b/wiki/en/Running-a-Server.md
@@ -172,7 +172,7 @@ Show an agreement window before users can connect.  The text of the agreement to
 
 ##### `-w or --welcomemessage`
 A "welcome message" to display in the Client chat window on connect. Can be given as a string or filename, and can contain HTML.
-When a path is used, the full path must be accessible by the `jamulus` Linux user. If not accessible, the literal path (rather than its contents) will appear.
+When a path is used, the file must be accessible by the user account running Jamulus. (On Linux and Mac, user `jamulus` is used by default.) If not accessible, the literal path (rather than its contents) will appear. 
 
 ##### `--serverpublicip`
 The public IP address of the Server if connecting to a Directory behind the same NAT. See [the Directories guide](Directories) for further information.
--- a/wiki/en/Running-a-Server.md
+++ b/wiki/en/Running-a-Server.md
@@ -172,7 +172,7 @@ Show an agreement window before users can connect.  The text of the agreement to
 
 ##### `-w or --welcomemessage`
 A "welcome message" to display in the Client chat window on connect. Can be given as a string or filename, and can contain HTML.
-When a path is used, the file must be accessible by the user account running Jamulus. (On Linux and Mac, user `jamulus` is used by default.) If not accessible, the literal path (rather than its contents) will appear. 
+When a path is used, the file must be accessible by the user account running Jamulus. (On most Linux installations, user `jamulus` is used by default.) If not accessible, the literal path (rather than its contents) will appear. 
 
 ##### `--serverpublicip`
 The public IP address of the Server if connecting to a Directory behind the same NAT. See [the Directories guide](Directories) for further information.
--- a/robots.txt
+++ b/robots.txt
@@ -2,3 +2,4 @@ User-agent: *
 Allow: /
 
 Sitemap: https://jamulus.io/sitemap.xml
+
--- a/robots.txt
+++ b/robots.txt
@@ -2,4 +2,3 @@ User-agent: *
 Allow: /
 
 Sitemap: https://jamulus.io/sitemap.xml
-


mcfnord:
re: https://github.com/jamulussoftware/jamulus/issues/3072

mcfnord:
This isn't good enough, because the default security prohibits access to /home/*. Even if /home/jamulus exists?

wiki/en/Running-a-Server.md
```diff
@@ -172,6 +172,7 @@ Show an agreement window before users can connect.  The text of the agreement to
 
 ##### `-w or --welcomemessage`
 A "welcome message" to display in the Client chat window on connect. Can be given as a string or filename, and can contain HTML.
+When a path is used, the full path must be accessible by the `jamulus` Linux user. If not accessible, the literal path (rather than its contents) will appear.
```
pljones: > the `jamulus` Linux user

This should refer to "the user running Jamulus", because it applies regardless of operating system and who runs it.

I'd have to check the source but the file should only need to be readable (i.e. if as the user running Jamulus, you can read the file, then Jamulus running as that user can read the file and will use it).

mcfnord: I know there can be a "we rely on operators" approach around here, but I prefer assuming the user is completely ignorant about the nuances of platform access controls. Comprehending which user is running Jamulus is a reasonably heavy mental lift to me. The `jamulus` user appears in the `systemctl --edit` configuration file. I suspect this user gets created by systemd if it does not exist already, and it probably does not. Which means by default, there is no `/home/jamulus`, and even if there was, a security feature we use would prevent this process from reading from /home/* anyway. So specifying this user is misleading, because they don't have that interactive user file path, and we couldn't use it if they did.

Creating `/etc/jamulus` was suggested, which probably entails a **mkdir** and then a **chown**. If we provide a little three-line pasteable bash script for that, maybe that's not a bad solution.

> On Linux or Mac, enter:
> 
> ```
> mkdir /etc/jamulus
> chown jamulus /etc/jamulus
> echo Welcome > /etc/jamulus/welc.html
> ```
> 
> Example: ... -w /etc/jamulus/welc.html

ann0see: What about "the user running the Server (e.g. `jamulus` on the default Linux install)"?

ann0see:
Before this gets merged, please squash the commits or squash-merge them.

wiki/en/Running-a-Server.md
mcfnord: Do you feel that's better than the draft I wrote in the previous comment?

Problem 1: /home/jamulus doesn't exist. So... what user? On Day 1, Linus told us some user accounts don't have interactive /home/username space. But I slept through Day 1.
Problem 2: Even if the user just creates an interactive /home/jamulus, they can't use it. Hope they don't try, because
Problem 3: There are no error explanations.

So I like recommending /etc/jamulus/welc.html on Linux and Mac, if that would work. It's not really clear where such a file can and should go.

mcfnord:
```suggestion
When a path is used, the file must be accessible by the user account running Jamulus. (On Linux and Mac, user `jamulus` is used by default.) If not accessible, the literal path (rather than its contents) will appear. 
```

ann0see: I mean there's probably no "jamulus" user on macOS or differently set up linux systems. 
Yes, the /etc/jamulus/ path would be ok if you think of it as a config file. (I think I store it under /var/jamulus/ for my server - but I don't remember).

ann0see: If you install the deb file a user called jamulus gets created. See the postinst file. (At the moment it's here: https://github.com/jamulussoftware/jamulus/blob/main/linux/debian/jamulus-headless.postinst)

pljones: You don't need to run on a Debian-based Linux OS system to run Jamulus as a server.  Any documentation needs to cover all cases or be generic enough to cover all cases.

mcfnord: Covering all cases isn't needed. Generic plus major platforms is enough. Generic without major platforms results in friction for most readers. With 20 years writing docs, I have 20 years of this debate with lead engineers. Be skeptical of omissions and support explicit guidance for the biggest platforms.

wiki/en/Running-a-Server.md
```diff
@@ -172,6 +172,7 @@ Show an agreement window before users can connect.  The text of the agreement to
 
 ##### `-w or --welcomemessage`
 A "welcome message" to display in the Client chat window on connect. Can be given as a string or filename, and can contain HTML.
+When a path is used, the file must be accessible by the user account running Jamulus. (On Linux and Mac, user `jamulus` is used by default.) If not accessible, the literal path (rather than its contents) will appear.
```
mcfnord:
```suggestion
When a path is used, the file must be accessible by the user account running Jamulus. (On most Linux installations, user `jamulus` is used by default.) If not accessible, the literal path (rather than its contents) will appear. 
```

ann0see (approved):
I'd be happy with it. Still it should be squashed and reviewed by @pljones or @gilgongo

gilgongo:
Would the addition of this note apply not only to `-w` but also to `--jsonrpcsecretfile`, `--directoryfile` and other commands that read files?

pljones:
> Would the addition of this note apply not to only to `-w` but also to `--jsonrpcsecretfile`, `--directoryfile` and other commands that read files?

Yes, it's fundamental to supplying a file that a program needs to read (and potentially write to).  So it applies not only to Jamulus but to any program run from the command line with filename arguments.

gilgongo:
> > Would the addition of this note apply not to only to `-w` but also to `--jsonrpcsecretfile`, `--directoryfile` and other commands that read files?
> 
> Yes, it's fundamental to supplying a file that a program needs to read (and potentially write to). So it applies not only to Jamulus but to any program run from the command line with filename arguments.

OK so this another case of deciding whether *nix users can use *nix. Either we add this wording to all those (and any future) command docs, or we assume a certain level of knowledge.

ann0see:
I mean jsonrpcsecretfile (?) would give an error if the file is not readable - at least that's what I remember.

mcfnord:
Squash has defeated me. Can I, a git user, use git? No.

pljones:
> Squash has defeated me. Can I, a git user, use git? No.

Where `main` is set to the appropriate value (`main` for `jamulussoftware/jamulus` and `next-release` for `jamulussoftware/jamuluswebsite`):
```sh
git fetch upstream ${main} ;# pull upstream project
git push origin ${main} ;# make sure we're up to date ... if this doesn't work, manual fix ups may be needed...
git status ;# manually check you're on the right branch and have commited
git rebase -i ${main} ;# interactive rebase - easiest way to squash cleanly
```
Then, in your chosen editor, you'll see your branch commits labelled `pick`.  Leave the first one, change the others to `s`. (`!}s/^pick/s` in `vi`).

ann0see:
@gilgongo @mcfnord would you be ok with me squashing it? I'd still prefer if the original author did it.

Git is confusing - I agree. https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#submitting-code-and-getting-started is a link to a short paragraph for contributors helping to get more information about git. There are some links which might be good? I believe something like https://medium.com/swlh/squash-and-rebase-git-basics-5cb1be1e0dac for the concept and  https://medium.com/@slamflipstrom/a-beginners-guide-to-squashing-commits-with-git-rebase-8185cf6e62ec would help me.

mcfnord:
I tried squashing but failed. I don't care enough about this PR to try again, but do recommend publishing it.

gilgongo:
> @gilgongo @mcfnord would you be ok with me squashing it? I'd still prefer if the original author did it.

There is still the question of whether, if we explain the issue of permissions in one place, we need to also explain it in all other places where someone might also not understand it. https://github.com/jamulussoftware/jamuluswebsite/pull/927#issuecomment-1570973777 

And if we explain that when a program needs read/write access to a file it also needs to be readable/writable by that program's user, then what else do we need to explain?

mcfnord:
Philosophically, improving one thing doesn't _require_ parallel improvements elsewhere. And philosophically, while we might say more about something, saying half often resolves 90% of confusion, leaving 10% for the other half. 

I look for failure patterns. If I haven't witnessed evidence of failure, it's not a priority to me to improve guidance. Whereas with the welcome file, I experienced a cascade of failures without error guidance due to the access restrictions and the unique text-or-path architecture of -w.

ann0see:
I think I'll leave the decision, further actions and merge up to @gilgongo as I feel I'd block here. 

We squash merge next-release to release so I believe we can merge this without a squash.

ann0see:
This means: I'll unsubscribe from this PR. If you need me, please ping me.

PR #933, opened 2023-07-06
Welcome path file access clues

mcfnord:
This PR is a replacement for https://github.com/jamulussoftware/jamuluswebsite/pull/927/

Prior discussion visible at previous PR (now closed).

I believe @ann0see defers to @gilgongo for this PR.

@gilgongo and @pljones articulated concern that all instances of this UX should receive the same guidance, but I don't believe there are other instances of this particular UX, as the others serve error messages, while this one just indicates failure by treating the inaccessible path as a literal string.

### DIFF
--- a/wiki/en/Running-a-Server.md
+++ b/wiki/en/Running-a-Server.md
@@ -172,6 +172,8 @@ Show an agreement window before users can connect.  The text of the agreement to
 
 ##### `-w or --welcomemessage`
 A "welcome message" to display in the Client chat window on connect. Can be given as a string or filename, and can contain HTML.
+When a path is used, the file must be accessible by the user account running Jamulus. (On most Linux installations, user `jamulus` is used by default.)
+If not accessible, the literal path (rather than its contents) will appear. 
 
 ##### `--serverpublicip`
 The public IP address of the Server if connecting to a Directory behind the same NAT. See [the Directories guide](Directories) for further information.


ann0see (approved):

gilgongo:
The words, "_the file must be accessible by the user account running Jamulus_" would apply to `directoryfile`  `recording` `inifile` and some other things too I think?

~~The other problem is that if a reader needs to know that a process has to have read access to something it needs to read, then that implies they'd need to know about other things at that "level", which might be hard to predict - for example, in the Client docs we talk about ports needing to be open and that IP addresses need to be static, etc. But we don't in the Server docs.~~

EDIT: I see you explained this in the closed PR. In essence, because you have experienced a problem with not making something readable by the process, this amend to the docs is warranted in case somebody else experiences the same. That's another way of looking at it I suppose. In which case, and if we ignore the above fact that other things would also qualify for this caveat, then yes.

gilgongo (approved):

mcfnord:
I'm glad this got merged! 

I'm convinced an empty /etc/jamulus/welcome.html file specified in the default systemd configuration is the best solution, and @ann0see has said it's not hard to do. I haven't determined if an empty welcome.html file has the same behavior as no welcome file.

I apply a "failure-driven" approach to prioritizing doc improvements. I had a very frustrating time figuring out where to put welcome.html. 

I'd act on evidence that people fail because the Client docs don't specify that the port must be open, or IP address must be static. How can I trigger these failures? Are there helpful error messages?

ann0see:
> I'd act on evidence that people fail because the Client docs don't specify that the port must be open, or IP address must be static. How can I trigger these failures? Are there helpful error messages?

That's in the docs and the client can't detect that reliably. Also registered servers neither need port forwarding nor a static IP. There is a private (and purposely undocumented) directory people can register on. 
But that should be discussed in another issue.

PR #936, opened 2023-07-24
Document macOS ini file sandbox requirements

ann0see:
Adds a short description for macOS users where the ini files should be saved to due to the sandbox.

Related to: https://github.com/jamulussoftware/jamulus/issues/2131

Ready for review. Path was verified.

Ready for review

YES

## Checklist

### DIFF
--- a/wiki/en/Include-Shared-Commands.md
+++ b/wiki/en/Include-Shared-Commands.md
@@ -2,7 +2,7 @@
 [comment]: # (This is an include file for use in multiple documents)
 
 - `-h` or `--help` Display help text
-- `-i` or `--inifile` Set location of initialization file (overrides default)
+- `-i` or `--inifile` Set location of initialization file (overrides default. On macOS simply provide a filename only, since config files can only be read from `/Users/<username>/Library/Containers/io.jamulus.Jamulus/Data/`. For the server replace `io.jamulus.Jamulus` with `io.jamulus.JamulusServer`. Turn on "Show Library folder" in "Show view options" in Finder to see this folder.)
 - `-n` or `--nogui` Disable GUI (for use in headless mode)
 - `-p` or `--port` Sets the local UDP port number. Default is 22124
 - `--jsonrpcport` Enables JSON-RPC API server to control the app, set TCP port number (EXPERIMENTAL, APIs might change; only accessible from localhost). Please see [the JSON-RPC API Documentation file](https://github.com/jamulussoftware/jamulus/blob/main/docs/JSON-RPC.md).


ann0see (commented):
I believe it could also be in [...]/.config/name.ini

wiki/en/Include-Shared-Commands.md
```diff
@@ -2,7 +2,7 @@
 [comment]: # (This is an include file for use in multiple documents)
 
 - `-h` or `--help` Display help text
-- `-i` or `--inifile` Set location of initialization file (overrides default)
+- `-i` or `--inifile` Set location of initialization file (overrides default. On macOS you should only give a filename, not a path, as the App is sandboxed. This will save the files to `~/Library/Containers/io.jamulus.Jamulus/Data/`)
```
ann0see:
```suggestion
- `-i` or `--inifile` Set location of initialization file (overrides default. On macOS you should only give a filename, not a path, as the App is sandboxed. This will save the files to `~/Library/Containers/io.jamulus.Jamulus/Data/` replace io.jamulus.Jamulus with io.jamulus.JamulusServer for the server)
```

ann0see: Ok. Checked. The folder is correct.

ann0see:
@gilgongo could you please have a look at this?

ann0see:
Quoting you here:

> "On macOS, the path to the file must be relative to /Users/<username>/Library/Containers/io.jamulus.Jamulus/Data/.config/Jamulus/ (turn on "Show Library folder" in "Show view options" in the Finder to see this). So a file name with no path is recommended. Note that the legacy version of Jamulus does not have this restriction."

So we might want to add that to the documentation. But it still makes the text quite long.

gilgongo:
Or if we assume that people on macOS who are interested in .ini files will know how to navigate to things, we could shorten it a bit to:

"On macOS, the path to the file must be relative to `/Users/<username>/Library/Containers/io.jamulus.Jamulus/Data/.config/Jamulus/`. So a file name with no path is recommended. Note that the legacy version of Jamulus can read the file from any location."

ann0see:
I think that's more confusing. Why give the path if you don't even need to give it? Therefore I'd prefer to say that it gets saved there.

gilgongo:
> I think that's more confusing. Why give the path if you don't even need to give it? Therefore I'd prefer to say that it gets saved there.

How about:

"On macOS, simply provide a file name only, since all config files can only be read from `/Users/<username>/Library/Containers/io.jamulus.Jamulus/Data/.config/Jamulus/`. Note that the legacy version of Jamulus for macOS can read from any location."

Or if we think people don't need the "no path" thing explaining:

"On macOS all config files must be saved in `/Users/<username>/Library/Containers/io.jamulus.Jamulus/Data/.config/Jamulus/`. Note that the legacy version of Jamulus for macOS can read from any location."

ann0see:
I'd prefer the first one

> "On macOS, simply provide a file name only, since all config files can only be read from ~/Library/Containers/io.jamulus.Jamulus/Data/.config/Jamulus/. Note that the legacy version of Jamulus for macOS can read from any location."

we can use `~/` instead of the username stuff

ann0see:
I'll drop mentioning the legacy should be considered as deprecated.

gilgongo:
> we can use `~/` instead of the username stuff

Although we did wonder if macOS users would know what `$HOME` meant, so let's hope they know what `~/` means as that won't let them get to the file path in the Finder if they wanted to do that.

ann0see:
I believe finder would support that. It's the normal user directory.

gilgongo:
> I believe finder would support that. It's the normal user directory.

As I noted earlier though, you have to explicity show Library or you can't see it in the Finder.

ann0see:
True. Probably worth documenting that. Could you please add a suggestion?

gilgongo:
I think we'd have to have the full path in there for that though, because if you didn't know what `~/` meant you would not know to go to home in the Finder and then turn on the Library, because that option is only available from the Finder's home location.

"On macOS, simply provide a file name only, since all config files can only be read from `/Users/<username>/Library/Containers/io.jamulus.Jamulus/Data/.config/Jamulus/` (turn on "Show Library folder" in "Show view options" to see this in the Finder). Note that the legacy version of Jamulus for macOS can read from any location."

ann0see:
1. I'll not mention the legacy version as it is deprecated. 
2. See the changes for a slightly adapted version including the server. 
3. There's no need to mention the `.config/Jamulus` folder, as that's ONLY for the default config file. If someone adds a new one, it won't be saved there.

gilgongo:
Oh OK (what do you mean in your point 2?), so more like:

"On macOS, simply provide a file name only, since all config files must be in `/Users/<username>/Library/Containers/io.jamulus.Jamulus/Data/` (turn on "Show Library folder" in "Show view options" to see this in the Finder)."

ann0see:
For 2. See the current state of this PR (= the change in my commit)

ann0see:
@gilgongo Please merge if this is finished. I think the state is ok.

PR #1001, opened 2024-07-27
Apply rexml upgrade again

ann0see:
Got another PR from dependabot which should be merged to next-release


pljones:
Closing as duplicate of #1000

ann0see:
The next-release branch needs to be updated.

pljones:
Oh I see.  Is there anything blocking this happening?

ann0see:
Yes. I think automation doesn’t work since we haven’t white listed some outdated dependencies. And there are conflicts here.

ann0see:
Probably we‘ll just need to update the lock file locally on a PC. 
The site is ok despite the build error.

ann0see:
Ok. Updated the lock file on PC. Will also need some update of the ruby version - but this should happen on next-release.

PR #1026, opened 2024-08-25
Identify canonical and alternate URLs

gilgongo:
_(The Google docs on this are confusing because they also address things that don't apply to us, so I hope I have the following right!)_

We regard the English version of a page as always being the canonical version. So no other languages should be marked as canonical URLs to content that also exits in English. Alternate URLs are other language versions of the English content.

**Context**

This should fix an issue which (I think!) Google has when one or more URLs have similar content. For example, if https://jamulus/[lang]/wiki/Installation-for-Linux is partially translated, we need to tell Google that it should regard https://jamulus/wiki/Installation-for-Linux as the canonical version.  **NOTE**: It's possible that this problem was a symptom of us not having correct `hreflang` codes, but using these `rel` tags seems to be best practice in our case. 

Removed a "rel=canonical" tag from the index pages because we should not put that outside of the HTML head. 

Also removed the `x-default-hreflang`, as we [don't have a language-switching](https://developers.google.com/search/blog/2013/04/x-default-hreflang-for-international-pages?authuser=0) page.

May need to update the language index pages? Otherwise no?

## Checklist

### DIFF
--- a/_includes/headtags.html
+++ b/_includes/headtags.html
@@ -1,7 +1,7 @@
 <meta charset="utf-8">
 <!-- Canonical and Alternate Links -->
 {% comment %}
-English URLs don't have a lang subdirectory (exact reason unkown):
+English URLs don't have a lang subdirectory since it's the default language:
 {% endcomment %}
     <link rel="canonical" hreflang="en" href="{{ site.url }}{{ site.baseurl }}{{ page.url }}">
 {% for lng in site.languages | sort %}
--- a/_includes/headtags.html
+++ b/_includes/headtags.html
@@ -1,6 +1,9 @@
 <meta charset="utf-8">
 <!-- Canonical and Alternate Links -->
-<link rel="canonical" hreflang="en" href="{{ site.url }}{{ page.url }}">
+{% comment %}
+English URLs don't have a lang subdirectory (exact reason unkown):
+{% endcomment %}
+   <link rel="canonical" hreflang="en" href="{{ site.url }}{{ page.url }}">
 {% for lng in site.languages | sort %}
   {% if lng != "en" %}
     <link rel="alternate" hreflang="{{ lng }}" href="{{ site.url }}{{ site.baseurl }}/{{ lng }}{{ page.url }}">
--- a/_includes/headtags.html
+++ b/_includes/headtags.html
@@ -36,4 +36,3 @@
 <meta name="color-scheme" content="dark light">
 <link rel="author" href="/humans.txt">
 <script src="/assets/js/nav.js"></script>
-gilgongo@desk:~$
--- a/_includes/headtags.html
+++ b/_includes/headtags.html
@@ -1,14 +1,13 @@
 <meta charset="utf-8">
 <!-- Canonical and Alternate Links -->
+<link rel="canonical" hreflang="en" href="https://jamulus.io{{ page.url }}">
 {% assign sorted_languages = site.languages | sort %}
 {% for lng in sorted_languages %}
-  {% if lng == site.default_lang %}
-    <link rel="canonical" href="{{ site.url }}{{ site.baseurl }}{{ page.url }}">
+  {% if lng == site.active_lang %}
   {% else %}
     <link rel="alternate" hreflang="{{ lng }}" href="{{ site.url }}{{ site.baseurl }}/{{ lng }}{{ page.url }}">
   {% endif %}
 {% endfor %}
-<link rel="alternate" hreflang="x-default" href="{{ site.url }}/{{ page.url }}">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 <link rel="stylesheet" href="/assets/css/fox.min.css">
 <link rel="stylesheet" href="/assets/css/fw.css">
@@ -37,3 +36,4 @@
 <meta name="color-scheme" content="dark light">
 <link rel="author" href="/humans.txt">
 <script src="/assets/js/nav.js"></script>
+gilgongo@desk:~$
--- a/_includes/headtags.html
+++ b/_includes/headtags.html
@@ -3,7 +3,7 @@
 {% comment %}
 English URLs don't have a lang subdirectory (exact reason unkown):
 {% endcomment %}
-   <link rel="canonical" hreflang="en" href="{{ site.url }}{{ page.url }}">
+    <link rel="canonical" hreflang="en" href="{{ site.url }}{{ site.baseurl }}{{ page.url }}">
 {% for lng in site.languages | sort %}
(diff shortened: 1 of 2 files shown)


_includes/headtags.html
```diff
@@ -1,14 +1,13 @@
 <meta charset="utf-8">
 <!-- Canonical and Alternate Links -->
+<link rel="canonical" hreflang="en" href="https://jamulus.io{{ page.url }}">
```
pljones: So this is absolute, even for the `en` version?

_includes/headtags.html
```diff
@@ -1,14 +1,13 @@
...
 {% assign sorted_languages = site.languages | sort %}
 {% for lng in sorted_languages %}
-  {% if lng == site.default_lang %}
-    <link rel="canonical" href="{{ site.url }}{{ site.baseurl }}{{ page.url }}">
+  {% if lng == site.active_lang %}
```
pljones: Change the empty `if` block using `==` to use `!=` (so making the `else` content the `if` content and removing the `else`).

_includes/headtags.html
gilgongo: I couldn't get the variable to expand properly, but can try again.

_includes/headtags.html
```diff
@@ -1,14 +1,12 @@
 <meta charset="utf-8">
 <!-- Canonical and Alternate Links -->
+<link rel="canonical" hreflang="en" href="{{ site.url }}{{ page.url }}">
 {% assign sorted_languages = site.languages | sort %}
 {% for lng in sorted_languages %}
```
pljones: Can these be done in the one line?
```suggestion
{% for lng in site.languages | sort %}
```

gilgongo: I think so - will check.

EDIT: yes that works.

_includes/headtags.html
```diff
@@ -1,14 +1,11 @@
 <meta charset="utf-8">
 <!-- Canonical and Alternate Links -->
-{% assign sorted_languages = site.languages | sort %}
-{% for lng in sorted_languages %}
-  {% if lng == site.default_lang %}
-    <link rel="canonical" href="{{ site.url }}{{ site.baseurl }}{{ page.url }}">
-  {% else %}
+<link rel="canonical" hreflang="en" href="{{ site.url }}{{ page.url }}">
```
pljones: By the way, why isn't `{{ page.url }}` here `{{ site.baseurl }}/en/{{ page.url }}`?  It might be worth adding a `{# #}` comment.

gilgongo: I wondered that too but not sure what I'd put as a comment since I don't know why it is.

pljones: Maybe "This works... but keep in view we do not understand why"?

ann0see: There is no /en/ subfolder as far as I know. Also the baseurl would probably just to make it an absolute url

pljones: `href="{{ site.url }}..."` should be the absolute part.

ignotus666 (approved):
I don't really know enough so I'll defer to those who do.

gilgongo:
@ignotus666 Just checking: nothing in the po4a scripts is affected by this, I take it?

ignotus666:
No, I shouldn't think so - if you can build and view the website offline and it works, then the po4a stuff should be fine.

pljones (approved):

_includes/headtags.html
```diff
@@ -1,14 +1,14 @@
 <meta charset="utf-8">
 <!-- Canonical and Alternate Links -->
-{% assign sorted_languages = site.languages | sort %}
-{% for lng in sorted_languages %}
-  {% if lng == site.default_lang %}
-    <link rel="canonical" href="{{ site.url }}{{ site.baseurl }}{{ page.url }}">
-  {% else %}
+{% comment %}
+English URLs don't have a lang subdirectory (exact reason unkown):
```
ann0see:
```suggestion
English URLs don't have a lang subdirectory since it's the default language:
```

https://github.com/jamulussoftware/jamuluswebsite/blob/628943e60ad9c73d615ac913a101b40744052ce2/_config.yml#L6

ann0see (changes requested):

_includes/headtags.html
pljones: That doesn't explain why `en` doesn't appear in the URL, though -- it just says what the default value is.

ann0see: But the default is just no language folder I guess? 


We use https://github.com/untra/polyglot

ann0see (approved):
@pljones Do you think we need to add anything else concerning the plug in?

_includes/headtags.html
pljones: Well, that probably needs to be in the explanatory comment, in case things change in the tooling.

ann0see: @gilgongo Could you please add the link?

ann0see:
Thanks

gilgongo:
Are we OK to merge?

pljones (approved):

pljones:
I don't think it's right...

https://github.com/jamulussoftware/jamuluswebsite/actions/runs/10696055730
http://jamuluswebsite.drealm.info/wiki/Installation-for-Linux
```
<!DOCTYPE html>
<!--[if IE 9]>
<html class="ie9">
<![endif]-->
<html lang="en">
<head>
<title>Installation for Linux – Jamulus Website</title>
<meta charset="utf-8">
<link rel="canonical" hreflang="en" href="/wiki/Installation-for-Linux">
<link rel="alternate" hreflang="sv-SE" href="/sv-SE/wiki/Installation-for-Linux">
...
```
http://jamuluswebsite.drealm.info/sv-SE/wiki/Installation-for-Linux
```
<!DOCTYPE html>
<!--[if IE 9]>
<html class="ie9">
<![endif]-->
<html lang="sv-SE">
<head>
<title>Installation för Linux – Jamulus websida</title>
<meta charset="utf-8">
<link rel="canonical" hreflang="en" href="/sv-SE/wiki/Installation-for-Linux">
<link rel="alternate" hreflang="sv-SE" href="/sv-SE/wiki/Installation-for-Linux">
...
```

1. If `ie9`, we get _two_ `html` tags, which cannot be right.  Does IE 9 still exist?  Is the XML comment actually parsed?
2. The http://jamuluswebsite.drealm.info/sv-SE/wiki/Installation-for-Linux page says `canonical` is `en`, okay but gives the wrong `href` to get there.


I should have spotted that earlier, sorry!

ann0see:
I think we can get rid of IE compatibility. Nobody should be using it anymore now.

PR #1039, opened 2024-09-08
Fix for jekyll canonical URL construction

gilgongo:
This is to test whether the .zip version of the site renders the same as that of a local build, since we spotted a possible issue with that in with https://github.com/jamulussoftware/jamuluswebsite/pull/1026

EDIT: Turned out that it was a quirk of the Jekyll Polyglot plugin that was only apparent when running `jekyll build`.

### DIFF
--- a/_includes/headtags.html
+++ b/_includes/headtags.html
@@ -4,7 +4,7 @@
 English URLs don't have a lang subdirectory since it's the default language.
 See https://github.com/untra/polyglot
 {% endcomment %}
-    <link rel="canonical" hreflang="en" href="{{ site.url }}{{ site.baseurl }}/{{page.name}}">
+    <link rel="canonical" hreflang="en" href="{{ site.url }}/{{page.name}}">
 {% for lng in site.languages | sort %}
   {% if lng != "en" %}
     <link rel="alternate" hreflang="{{ lng }}" href="{{ site.url }}{{ site.baseurl }}/{{ lng }}{{ page.url }}">
--- a/_includes/headtags.html
+++ b/_includes/headtags.html
@@ -4,7 +4,7 @@
 English URLs don't have a lang subdirectory since it's the default language.
 See https://github.com/untra/polyglot
 {% endcomment %}
-    <link rel="canonical" hreflang="en" href="{{ site.url }}{{ page.url }}">
+    <link rel="canonical" hreflang="en" href="{{ site.url }}{{ site.baseurl }}/{{page.name}}">
 {% for lng in site.languages | sort %}
   {% if lng != "en" %}
     <link rel="alternate" hreflang="{{ lng }}" href="{{ site.url }}{{ site.baseurl }}/{{ lng }}{{ page.url }}">
--- a/_includes/headtags.html
+++ b/_includes/headtags.html
@@ -4,7 +4,7 @@
 English URLs don't have a lang subdirectory since it's the default language.
 See https://github.com/untra/polyglot
 {% endcomment %}
-    <link rel="canonical" hreflang="en" href="{{ site.url }}{{ site.baseurl }}{{ page.url }}">
+    <link rel="canonical" hreflang="en" href="{{ site.url }}{{ page.url }}">
 {% for lng in site.languages | sort %}
   {% if lng != "en" %}
     <link rel="alternate" hreflang="{{ lng }}" href="{{ site.url }}{{ site.baseurl }}/{{ lng }}{{ page.url }}">
--- a/_includes/headtags.html
+++ b/_includes/headtags.html
@@ -4,15 +4,7 @@
 English URLs don't have a lang subdirectory since it's the default language.
 See https://github.com/untra/polyglot
 {% endcomment %}
-    <link rel="canonical" hreflang="en" href="{{ site.url }}{{page.url}}">
-<!--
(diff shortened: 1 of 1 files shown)


gilgongo:
``
'base' input parameter is ignored when action is triggered by pull request event``

Wonder what that means?

pljones:
Yep, still gets
```
<link rel="canonical" hreflang="en" href="/sv-SE/wiki/Installation-for-Linux">
```
unfortunately.  The `jekyll serve` shows "correctly" now for me.

I notice this in the build log under "Check for po4a cache":
> Cache restored from key: Linux-po4a-0.68

Later, under "Build the site in the jekyll/builder container":
> Bundler 2.3.25 is running, but your lockfile was generated with 2.5.18. Installing Bundler 2.5.18 and restarting using that version.

So it's picked up @softins' change.  Based on the above, I guess we need to try deleting the cached `po4a` stuff and running `./_po4a-tools/po4a-create-all-targets.sh` again with a fresh install.

gilgongo:
> I guess we need to try deleting the cached po4a stuff and running ./_po4a-tools/po4a-create-all-targets.sh again with a fresh install.

Isn't that what GH does when it deploys the site anyway? I'm very confused about all this, but if when we release the site we start telling Google that whatever is the current language is the canonical version, I think that's almost the opposite of what they want us to do (although mind you, today we're saying there is no canonical and all langs are alternate...!)

So how come locally built versions (I built mine using the latest Gemfile.lock from @softins) render correctly while the GH ones don't?

gilgongo:
OK I'm out of ideas, but this seems related if GH is using jekyll build: https://github.com/untra/polyglot/issues/206 

**Edit**: Yes. If I locally do `rm -r _site` and then `jekyll -b`, sure enough I see:

```
<html lang="fr">

<head>
  <title>Manuel du logiciel – Site internet de Jamulus</title>
  <meta charset="utf-8">
<!-- Canonical and Alternate Links -->

<link rel="canonical" hreflang="en" href="/fr/wiki/Software-Manual">
```

So - bug?

softins:
If GitHub builds `release`, I think it would still be using the old `Gemfile.lock`, as my PR was only merged into `next-release`. We could cherry pick that commit into `release`?

gilgongo:
> If GitHub builds `release`, I think it would still be using the old `Gemfile.lock`, as my PR was only merged into `next-release`. We could cherry pick that commit into `release`?

I think the issue is that `jekyll build` isn't interpreting the Jekyll stuff correctly (although `jekyll serve` is which is why I didn't spot this issue before the merge). At least, my branch is against `next-release`, and I can now reproduce the problem locally with that. 

What I'm seeing is something is adding the current page's language code path to the `{{ page.url }}` variable in some way. Even odder, it doesn't happen when I echo the variables in a comment:

So using `<link rel="canonical" hreflang="en" href="{{ site.url }}{{ page.url }}">` in `_includes/headtags.html`, I get this:

```

<link rel="canonical" hreflang="en" href="/de/wiki/Installation-for-Windows">
<!--
        page.url /wiki/Installation-for-Windows
-->
```

So where is the `/de` coming from? It also means I can't hack out a correct URL because when I try to use `{{ page.url }}` in any Liquid statements (to chop off the leading dir), it gets expanded correctly without the language dir. I've also tried `{{ page.permalink }}` which we use in front matter but that gets the same treatment!

**EDIT**: Out of interest, I tried removing the `href=` from `href="{{ site.url }}{{ page.url }}"` and the path comes out OK. So is there something in Polyglot that looks for that and appends the language dir onto it?

gilgongo:
Ah, found it. Needed to use the `static_href` variable. Should be working now.

Will let somebody verify that independently but otherwise good to merge.

pljones:
```
<link rel="canonical" hreflang="en" href="/wiki/Installation-for-Linux">
<link rel="alternate" hreflang="sv-SE" href="/sv-SE/wiki/Installation-for-Linux">
```
Looks good!

pljones (approved):

gilgongo:
Oh.

```
  Conversion error: Jekyll::Converters::Markdown encountered an error while converting '_posts/2021-01-05-Jamulus-Sound-Devices.md/#excerpt':
                    No such file or directory @ rb_sysopen - /srv/jekyll/.jekyll-cache/Jekyll/Cache/Jekyll--Converters--Markdown/86/4d0f894f12a515b443a594ebf17e166e9e29842271eddd6981d54bd3a84081
  Liquid Exception: No such file or directory @ rb_sysopen - /srv/jekyll/.jekyll-cache/Jekyll/Cache/Jekyll--Converters--Markdown/86/4d0f894f12a515b443a594ebf17e166e9e29842271eddd6981d54bd3a84081 in /srv/jekyll/_layouts/kblist.html

```

softins:
Maybe clear the cache and try again?

gilgongo:
How do I do that?

softins:
I would guess just remove the `.jekyll-cache` directory. I think it should then get recreated. Maybe there is some mismatch between a new and an old cached file?

gilgongo:
Ah, OK. I re-ran the workflow and it worked on the second try.

softins:
> Ah, OK. I re-ran the workflow and it worked on the second try.

Ah, I didn't realise you were talking about a Github workflow, rather than your local system.

PR #1053, opened 2024-10-05
--ctrlmidich improved documentation

pljones:

https://github.com/jamulussoftware/jamulus/pull/3394 adds a new feature to the `--ctrlmidich` option to allow binding a MIDI CC number explicitly to the participant's "own" channel on the server, regardless of the assigned channel number.

This required some changes to the documentation to explain how it worked.  I took the opportunity to clarify (hopefully) the detailed information in Tips & Tricks.

See https://github.com/jamulussoftware/jamulus/pull/3394

Looks okay locally.

Needs reviewing and shouldn't be merged until after https://github.com/jamulussoftware/jamulus/pull/3394

YES

## Checklist

### DIFF
--- a/wiki/en/Include-Client-Commands.md
+++ b/wiki/en/Include-Client-Commands.md
@@ -2,5 +2,5 @@
 - `--mutemyown` Prevent me from hearing what I play in the server mix (headless only)                                                      
 -  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`  
 -  `-j` or `--nojackconnect`  Disable auto JACK connections  
--  `--ctrlmidich`  MIDI controller channel to listen on, control number offset and consecutive CC numbers (channels) and Mute Myself CC number. Format: `channel[;f*][;p*][;s*][;m*][;o]` See [Tips & Tricks](Tips-Tricks-More#using-ctrlmidich-for-midi-controllers)
+-  `--ctrlmidich`  MIDI controller channel to listen on, MIDI control number and count of consecutive CC numbers (Jamulus channels), Mute Myself CC number and "My Channel" option. Format: `channel[;fn[*n]][;pn[*n]][;sn[*n]][;mn[*n]][;on][;z]` See [Tips & Tricks](Tips-Tricks-More#using---ctrlmidich-for-midi-controllers).
 - `--clientname`  Window title and JACK client name
--- a/wiki/en/Include-Client-Commands.md
+++ b/wiki/en/Include-Client-Commands.md
@@ -2,5 +2,5 @@
 - `--mutemyown` Prevent me from hearing what I play in the server mix (headless only)                                                      
 -  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`  
 -  `-j` or `--nojackconnect`  Disable auto JACK connections  
--  `--ctrlmidich`  MIDI controller channel to listen on, MIDI control number and count of consecutive CC numbers (Jamulus channels), Mute Myself CC number and "My Channel" option. Format: `channel[;fn[*n]][;pn[*n]][;sn[*n]][;mn[*n]][;on][;z]` See [Tips & Tricks](Tips-Tricks-More#using---ctrlmidich-for-midi-controllers).
+-  `--ctrlmidich`  MIDI controller channel to listen on, MIDI control number and count of consecutive CC numbers (Jamulus channels), Mute Myself CC number and device selection option. Format: `channel[;fn[*n]][;pn[*n]][;sn[*n]][;mn[*n]][;on][;dDeviceName]` See [Tips & Tricks](Tips-Tricks-More#using---ctrlmidich-for-midi-controllers).
 - `--clientname`  Window title and JACK client name
--- a/wiki/en/Include-Client-Commands.md
+++ b/wiki/en/Include-Client-Commands.md
@@ -2,5 +2,5 @@
 - `--mutemyown` Prevent me from hearing what I play in the server mix (headless only)                                                      
 -  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`  
 -  `-j` or `--nojackconnect`  Disable auto JACK connections  
--  `--ctrlmidich`  MIDI controller channel to listen on, MIDI control number and count of consecutive CC numbers (Jamulus channels), Mute Myself CC number and device selection option. Format: `channel[;fn[*n]][;pn[*n]][;sn[*n]][;mn[*n]][;on][;dDeviceName]` See [Tips & Tricks](Tips-Tricks-More#using---ctrlmidich-for-midi-controllers).
+-  `--ctrlmidich`  MIDI channel to listen on, Jamulus control + MIDI control number and count of consecutive CC numbers (or Jamulus channels), device selection option. Format: `channel[;fn[*n]][;pn[*n]][;sn[*n]][;mn[*n]][;on][;dDeviceName]` See [Tips & Tricks](Tips-Tricks-More#using---ctrlmidich-for-midi-controllers).
 - `--clientname`  Window title and JACK client name
--- a/wiki/en/Tips-Tricks-More.md
+++ b/wiki/en/Tips-Tricks-More.md
@@ -119,7 +119,7 @@ When using JACK or macOS, make sure you connect your MIDI device's output port t
 2. The long form is a sequence of offsets and counts for various controllers:
 
    ```
-   [MIDI channel];[control letter][offset](*[count](;...))
+   [MIDI channel];[control letter][offset](*[count])(;...)
    ```
 
    * `MIDI channel` is required or else the parameter argument is ignored and the feature is not active.  `0` means "any channel", `1`-`16` listen only to MIDI messages on the specified MIDI channel.
--- a/wiki/en/Tips-Tricks-More.md
+++ b/wiki/en/Tips-Tricks-More.md
@@ -82,11 +82,13 @@ Here is the script:
 
 ### Using `--ctrlmidich` for MIDI controllers
 
-When this option is used on the command line, Jamulus will display a channel number assigned to each Client, which can be used to control the channel using MIDI CC numbers.  (Note: This is only available for use with Windows, macOS and Linux using Jamulus version 3.7.0 or higher.)  In Jamulus version 3.12.0 onwards, when connected to a server of at least version 3.5.5, your own fader will always be given channel 0, and so will appear first, whether or not "Own Fader First" is enabled.
+The volume fader, pan control and mute and solo buttons in the Client's mixer window strips can be controlled using a connected MIDI controller. This feature is available from version 3.7.0 on macOS, Linux, and the JACK version of Jamulus for Windows. From Jamulus 3.12.0 onwards, it is also available for the non-JACK (ASIO) Windows version.  To enable this feature, Jamulus must be launched with the `--ctrlmidich` command-line option.
+
+When this option is used on the command line, Jamulus will prepend a channel number to each Client name, which can be used to control the channel using MIDI CC numbers. In Jamulus version 3.12.0 onwards, when connected to a server of at least version 3.5.5, your own fader will always be given channel 0, and so will appear first when unsorted or sorted by channel, whether or not "Own Fader First" is enabled.
 
 *Tip*: With default settings, when some users leave and others join, their left-right arrangement in the GUI may cease to follow a numerical order, making it more difficult to know who each physical fader/knob on your MIDI controller corresponds to. To keep the fader strips following a numerical order, go to "View" on the top menu bar and switch to "Sort by Channel" (or type `Ctrl+E`).
 
-Make sure you connect your MIDI device's output port to the Jamulus MIDI in port (QjackCtl (Linux/Windows), Audio/MIDI Setup (macOS) or whatever you use for managing connections). In Linux you will need to install and launch a2jmidid so your device shows up in the MIDI tab in Qjackctl.  See below for Windows.
+When using JACK or macOS, make sure you connect your MIDI device's output port to the Jamulus MIDI in port (QjackCtl (Linux/Windows), Audio/MIDI Setup (macOS) or whatever you use for managing connections). In Linux you may need to install and launch `a2jmidid` so your device shows up in the MIDI tab in Qjackctl.  For non-JACK Windows, Jamulus will find the MIDI device(s) automatically, but see the `d` option below if more than one MIDI device is connected.
 
 `--ctrlmidich` takes a single argument.  If you omit it, the parameter is ignored.  There are two formats for the argument:
 
@@ -96,7 +98,7 @@ Make sure you connect your MIDI device's output port to the Jamulus MIDI in port
    [MIDI channel];[offset for first fader]
    ```
 
-   * `MIDI channel` is required or the the parameter argument is ignored and the feature is not active.  `0` means "any channel", `1`-`16` listen only to MIDI messages on the specified MIDI channel.
+   * `MIDI channel` is required or else the parameter argument is ignored and the feature is not active.  `0` means "any channel", `1`-`16` listen only to MIDI messages on the specified MIDI channel.
 
    * `offset for first fader` is the first MIDI CC to use to control a Jamulus Channel fader (default 70, which matches the Behringer X-Touch defaults), with all MIDI CCs after that being used; must be a number or else the long form is used.
 
@@ -120,7 +122,7 @@ Make sure you connect your MIDI device's output port to the Jamulus MIDI in port
    [MIDI channel];[control letter][offset](*[count](;...))
    ```
 
-   * `MIDI channel` is required or the the parameter argument is ignored and the feature is not active.  `0` means "any channel", `1`-`16` listen only to MIDI messages on the specified MIDI channel.
+   * `MIDI channel` is required or else the parameter argument is ignored and the feature is not active.  `0` means "any channel", `1`-`16` listen only to MIDI messages on the specified MIDI channel.
 
    * `control letter` defines which Jamulus Control the MIDI controller number is assigned to:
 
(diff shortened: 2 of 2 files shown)


pljones:
Website available for preview here:
http://jamuluswebsite.drealm.info/wiki/Software-Manual#command-line-options
http://jamuluswebsite.drealm.info/wiki/Tips-Tricks-More#using---ctrlmidich-for-midi-controllers

wiki/en/Tips-Tricks-More.md
```diff
@@ -80,21 +80,92 @@ Here is the script:
...
+When this option is used on the command line, Jamulus will display a server channel number assigned to each client, which can be used to control the channel using MIDI CC numbers.  (Note: This is only available for use with macOS and Linux using Jamulus version 3.7.0 or higher, and on Windows using the Jamulus version with JACK support.  It also relies on the Jamulus UI being displayed.)
```
gilgongo: Do "server" and "client" need to be capitalised here?

wiki/en/Tips-Tricks-More.md
```diff
@@ -80,21 +80,92 @@ Here is the script:
...
+   ```
+   [MIDI channel];[offset for first fader]
+   ```
 
-Fader strips in the mixer window are controlled in ascending order from left to right. Continuing with the above example, in strip number 1 (farthest left), the volume fader would be controlled by CC number 0; pan by 16; solo by 32 and mute by 48. As we have specified 8 consecutive controllers for each parameter, this would give us MIDI control over 8 strips (volume, pan, solo and mute in each one) in the mixer window. The next strip would be controlled by 1, 17, 33 and 49, and so forth.
+   * `MIDI channel` defaults to `0`, meaning "any channel"; value may be omitted if ";..." is present.
 
-Make sure you connect your MIDI device's output port to the Jamulus MIDI in port (QjackCtl (Linux/Windows), MIDI Studio (macOS) or whatever you use for managing connections). In Linux you will need to install and launch a2jmidid so your device shows up in the MIDI tab in Qjackctl.
+   * `offset for first fader` is the first MIDI CC to use to control a Jamulus Channel fader (default 70 for the sake of Behringer X-Touch), with all MIDI CCs after that being used; must be a number to use the legacy format.
```
gilgongo: Not sure what's meant by "for the sake of Behringer X-Touch") here.

wiki/en/Tips-Tricks-More.md
```diff
@@ -80,21 +80,92 @@ Here is the script:
...
+   ```
+
+   Here, Jamulus listens on MIDI channel 1 and uses nine MIDI controller numbers for each Jamulus control: one for "My Channel" plus eight for Jamulus channels 0 to 7.  MIDI CC numbers 0 (for "My Channel") to 8 control faders.  Pan controls start at CC number 16 ("my") and end at 24; Solo 32 ("my") to 40 and Mute 48 ("my") to 56; Mute Myself is enabled/disabled by CC number 64.
+
+Please note that buttons on your MIDI controller need to be set to "toggle" mode. This means that when pressed to 'turn on' a control, it must send a MIDI CC number with a value >=64, and to 'turn off' the control it must send the same CC number with a value <64. You can read your controller's manual to find out how to set this.
+
+*Note*: Jamulus does not provide feedback on the on/off state of buttons, meaning that your controller must keep track and toggle LEDs (if any) to 'on' or 'off' itself.
+
+Jamulus channel numbers are assigned by the server and displayed in the mixer when you use `--ctrlmidich`.  When you join a server, you can get a different Jamulus channel depending on who else is already there (hence using the "my channel" option, `z`).
```
gilgongo: The use of the term "Jamulus channel" - is that to differentiate it from another type of channel? If so, I'm not sure if that's clear.

wiki/en/Tips-Tricks-More.md
```diff
@@ -80,21 +80,92 @@ Here is the script:
...
 
 *Tip*: When you enable MIDI control in Jamulus, each user's name is prepended with a number, with the leftmost user starting at 0, then 1, etc. With default settings, when some users leave and others join, their left-right arrangement in the GUI may cease to follow a numerical order, making it more difficult to know who each physical fader/knob on your MIDI controller corresponds to. In order to keep the fader strips following a numerical order, go to "View" on the top menu bar and switch between "No User Sorting" and another option and then back again (e.g. type `Ctrl+N`, `Ctrl+O`).
```
gilgongo: I think the style guide implies we should use the word "fader" here (para starting line 170) and not "user", because it says a user could have more than one fader/channel. Although saying "when some users leave and others join" is OK, I assume.

wiki/en/Tips-Tricks-More.md
pljones: Original author's prerogative, I think.

wiki/en/Tips-Tricks-More.md
pljones: MIDI has channels and Jamulus has channels but they mean different (although somewhat similar) things - check the picture under "MIDI Channels" here:
https://cecm.indiana.edu/361/midi.html

wiki/en/Tips-Tricks-More.md
pljones: Could you state more clearly which instance of the word "fader" you're referring to.

I'll just point out that we need to distinguish between the UI widget that adjusts the volume level (a "slider" or "fader") and the hardware component that emits a MIDI message (also a "slider" or "fader" but could be a "knob" or "dial") and the Jamulus channel number assigned by the server to a participant that has volume, pan and mute widgets in the UI.

wiki/en/Tips-Tricks-More.md
pljones: Ah yes.

pljones: Done

wiki/en/Tips-Tricks-More.md
pljones: (Also, I don't know if the para is strictly correct.  It would probably be useful to have a "sort by channel" option available when `--ctrlmidich` was in use, too.)

wiki/en/Tips-Tricks-More.md
gilgongo: Does it mean "in the case of the Behringer X-Touch"? If so I'd change it to that.

wiki/en/Tips-Tricks-More.md
gilgongo: Oh I see, in which case it's normal to refer to the software's channel in this way.

wiki/en/Tips-Tricks-More.md
gilgongo: On second thoughts, I'm not sure. The style guide clarifies the differences between "channel", "fader" and "person", so I was wondering if here it should be, "When you enable MIDI control in Jamulus, each user's fader is prepended with a number,..."

wiki/en/Tips-Tricks-More.md
pljones: No, it doesn't.  It means for the sake of the Behringer X-Touch, which is what the original author used.  "In the case of" would mean there was some knowledge of what was connected and there isn't.

wiki/en/Tips-Tricks-More.md
pljones: That's still nonsense.  The fader isn't.  It's in front of the name field in the label below the mixer channel.

wiki/en/Tips-Tricks-More.md
```diff
@@ -80,21 +80,92 @@ Here is the script:
...
+When this option is used on the command line, Jamulus will display a Server channel number assigned to each Client, which can be used to control the channel using MIDI CC numbers.  (Note: This is only available for use with macOS and Linux using Jamulus version 3.7.0 or higher, and on Windows using the Jamulus version with JACK support.  It also relies on the Jamulus UI being displayed.)
 
-`--ctrlmidich "1;f0*8;p16*8;s32*8;m48*8;o64"`
+Make sure you connect your MIDI device's output port to the Jamulus MIDI in port (QjackCtl (Linux/Windows), MIDI Studio (macOS) or whatever you use for managing connections). In Linux you will need to install and launch a2jmidid so your device shows up in the MIDI tab in Qjackctl.
```
ann0see: Is it really called MIDI studio on macOS?
I thought it's Audio/MIDI setup. @gilgongo

ignotus666: I think I wrote this section back in the day. No idea where MIDI Studio came from. A quick check confirms that it's "Audio MIDI Setup". Also, I see there's a closing bracket missing in `(QjackCtl (Linux/Windows)`, or it needs to be rephrased.

pljones: `(QjackCtl (Linux/Windows), MIDI Studio (macOS) or whatever you use for managing connections)` looks balanced.

wiki/en/Tips-Tricks-More.md
gilgongo: OK, but what does "for the sake of Behringer X-Touch" mean exactly because I can't parse that phrase here. Is it "to be compatible with Behringer X-Touch"?

wiki/en/Tips-Tricks-More.md
gilgongo: > Is it really called MIDI studio on macOS? I thought it's Audio/MIDI setup. @gilgongo

I'm afraid I no longer have access to a Mac, but Apple talk about [Audio/MIDI setup here](https://support.apple.com/en-gb/guide/audio-midi-setup/ams59f301fda/mac ).

ignotus666: > (QjackCtl (Linux/Windows), MIDI Studio (macOS) or whatever you use for managing connections) looks balanced.

Oops, wasn't looking at the bigger picture!

wiki/en/Tips-Tricks-More.md
pljones: It's just "because ... Behringer X-Touch", I think.  As in "Well, that's what I used, why wouldn't I?"

(I'm happy to remove it.)

What I wish is I could remember the couple of other bits of info I saw _somewhere_ in the source I wanted to put here...

wiki/en/Tips-Tricks-More.md
pljones: But yes, I'll fix the macOS audio/midi thing.  (I thought it was always "Audio/MIDI setup"? At least it was about 14 years ago when I last saw a mac.)

pljones: Done.

wiki/en/Tips-Tricks-More.md
pljones: Updated.  Plus also corrected something I'd not understood in the parsing.

gilgongo (approved):

ann0see (approved):
On the basis that @ignotus666 also gives his ok to this...

ignotus666 (approved):
Looks good to me

pljones:
Spotted another couple of places I'd said "0 is default" (when it's not -- feature disabled is default).

ann0see:
If this is mergable, please do so.

pljones:
> shouldn't be merged until after https://github.com/jamulussoftware/jamulus/pull/3394

pljones:
Converted to draft as will need reconsidering after https://github.com/jamulussoftware/jamulus/pull/3426

pljones:
@softins do you want to take this branch and replace the `z` with the `d` option?

pljones:
Rebased and updated.  @softins you'll probably need to fix what I've just broken...

softins:
@pljones I have made my updates to the docs; please review and comment if desired.

I found that the CI was failing from today. It seems like the `ubuntu-latest` runner has now switched from 22.04 to 24.04, and this causes the Jekyll build to fail. I pushed a commit here to pin the CI runner to `ubuntu-22.04` and it then succeeded.

We ought to update the CI runner separately from this PR - I'll raise one shortly, and then revert that commit from this PR. ***DONE***

wiki/en/Include-Client-Commands.md
```diff
@@ -2,5 +2,5 @@
 - `--mutemyown` Prevent me from hearing what I play in the server mix (headless only)                                                      
 -  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`  
 -  `-j` or `--nojackconnect`  Disable auto JACK connections  
--  `--ctrlmidich`  MIDI controller channel to listen on, control number offset and consecutive CC numbers (channels) and Mute Myself CC number. Format: `channel[;f*][;p*][;s*][;m*][;o]` See [Tips & Tricks](Tips-Tricks-More#using-ctrlmidich-for-midi-controllers)
+-  `--ctrlmidich`  MIDI controller channel to listen on, MIDI control number and count of consecutive CC numbers (Jamulus channels), Mute Myself CC number and device selection option. Format: `channel[;fn[*n]][;pn[*n]][;sn[*n]][;mn[*n]][;on][;dDeviceName]` See [Tips & Tricks](Tips-Tricks-More#using---ctrlmidich-for-midi-controllers).
```
pljones:
```suggestion
-  `--ctrlmidich`  MIDI channel to listen on, Jamulus control + MIDI control number and count of consecutive CC numbers (or Jamulus channels), device selection option. Format: `channel[;fn[*n]][;pn[*n]][;sn[*n]][;mn[*n]][;on][;dDeviceName]` See [Tips & Tricks](Tips-Tricks-More#using---ctrlmidich-for-midi-controllers).
```
The English text was a bit inconsistent with the "Format" bit.  I'm not sure this is explanatory enough but it gives the link to the docs anyway.

wiki/en/Tips-Tricks-More.md
```diff
@@ -80,23 +80,110 @@ Here is the script:
...
+The volume fader, pan control and mute and solo buttons in the Client's mixer window strips can be controlled using a connected MIDI controller. This feature is available from version 3.7.0 on macOS, Linux, and the JACK version of Jamulus for Windows. From Jamulus 3.12.0 onwards, it is also available for the non-JACK (ASIO) Windows version.  To enable this feature, Jamulus must be launched with the `--ctrlmidich` command-line option.
 
-`--ctrlmidich "1;f0*8;p16*8;s32*8;m48*8;o64"`
+When this option is used on the command line, Jamulus will prepend a channel number to each Client name, which can be used to control the channel using MIDI CC numbers. In Jamulus version 3.12.0 onwards, when connected to a server of at least version 3.5.5, your own fader will always be given channel 0, and so will appear first when unsorted or sorted by channel, whether or not "Own Fader First" is enabled.
```
pljones:
```suggestion
When this option is used on the command line, Jamulus will prepend a channel number to each Client name, which can be used to control the channel using MIDI CC numbers. In Jamulus version 3.12.0 onwards, when connected to a server of at least version 3.5.5, your own fader will always be given channel 0, and so will appear first when sorted by channel or when "Own Fader First" is enabled.
```
In "unsorted", if you join an occupied server, you'll go to the position in the list you arrive back at the client in.  Are you guaranteed to be the first in that list?

wiki/en/Tips-Tricks-More.md
```diff
@@ -80,23 +80,110 @@ Here is the script:
...
+     ```
+     --ctrlmidich "2;50"
+     ```
+
+     This would listen on MIDI channel 2 and use MIDI controller 50 to control Jamulus channel 0 fader and so on.
+
+2. The long form is a sequence of offsets and counts for various controllers:
+
+   ```
+   [MIDI channel];[control letter][offset](*[count](;...))
```
pljones:
```suggestion
   [MIDI channel];[control letter][offset](*[count])(;...)
```

pljones:
Just a few bits and pieces.

wiki/en/Tips-Tricks-More.md
softins: I tested that. In 3.12.0, if you join an occupied server of 3.5.5 or later, you get channel 0 _and_ appear first in the unsorted list.

softins: When I tested the new version by connecting to an occupied server of a recent version, my own channel always showed up first in the unsorted view. Performing the same test against a server older than 3.5.5, my own channel appeared to the right of the the prior occupants.

I haven't worked through the code to examine whether this is guaranteed, but it appeared consistent in practice.

Also my wording of the final sentence "whether or not" was intended to make clear that when unsorted or sorted by channel, the _Own fader first_ option made no difference.

pljones: > I tested that. In 3.12.0, if you join an occupied server of 3.5.5 or later, you get channel 0 _and_ appear first in the unsorted list.

Does that mean the server holds multiple lists of clients, one per client, so it can consistently return the list to the client in that order?  Or does it just filter the response and set to zero and move to first the client for each client it's sending the list to?  Sounds like more work than needed...  (Especially given how much work the client then does with the list when it arrives...)

> Also my wording of the final sentence "whether or not" was intended to make clear that when unsorted or sorted by channel, the Own fader first option made no difference.

"Own Fader First" should be taken as the one that disregards sort order.  To say a sort order then disregards "Own Fader First" is somewhat circular.  It's either unnecessary or confusing.  "Unsorted" _shouldn't_ be expressed to mean "your own fader will be first (under some circumstances)" because that's not really that useful (even if it may be true).

softins: > > I tested that. In 3.12.0, if you join an occupied server of 3.5.5 or later, you get channel 0 _and_ appear first in the unsorted list.
> 
> Does that mean the server holds multiple lists of clients, one per client, so it can consistently return the list to the client in that order? Or does it just filter the response and set to zero and move to first the client for each client it's sending the list to? Sounds like more work than needed... (Especially given how much work the client then does with the list when it arrives...)

No, the server always sends its own channel numbers to every client. And from 3.5.5 it also sends the client its own server channel ID when it connects. Since the recently-merged update, each client maintains its own set of client-side channel numbers and maps between the two. It maps its own server channel number to client channel 0, and each new server-side channel to the lowest free client channel ID.

When the server sends a connected clients list to the client, the client rewrites the channel numbers in that list according to its own mapping, before passing it up to the mixer board. The mixer board doesn't need to know the client layer mapped them. But I think it processes new channels in order of channel number, rather than order in the list, which means that when processing the first list received after a new connection, it will put channel 0 first. I need to verify this by examining the code.

> > Also my wording of the final sentence "whether or not" was intended to make clear that when unsorted or sorted by channel, the Own fader first option made no difference.
> 
> "Own Fader First" should be taken as the one that disregards sort order. To say a sort order then disregards "Own Fader First" is somewhat circular. It's either unnecessary or confusing. "Unsorted" _shouldn't_ be expressed to mean "your own fader will be first (under some circumstances)" because that's not really that useful (even if it may be true).

OK, fair enough. I'll commit your wording.

ignotus666:
@softins any outstanding Weblate PRs have to be merged before a documentation (website) PR goes in... Weblate's locked up now.

softins:
> @softins any outstanding Weblate PRs have to be merged before a documentation (website) PR goes in... Weblate's locked up now.

Oh, right, sorry. Didn't know that! Would you like me to resolve the conflict? I'm happy to, but don't want to duplicate effort.

ignotus666:
Don't worry, it's happened to me a few times! The way I did it was to just revert the offending merged PR, merge the Weblate PR and then this one - but feel free to apply whatever fix you think is best.

softins:
That sounds good to me.

softins:
@ignotus666 ok, I've successfully reverted this PR, and the weblate PR no longer complains of a conflict. Shall I just go ahead and merge it, or does it need checking first?

ignotus666:
Yes, I think you can merge it, I checked the Spanish docs and they're ok, can't speak for the other one though.

PR #1083, opened 2025-01-05
Clarify use of capital letters

gilgongo:
This is just for the style guide, but I've created a PR just in case a direct commit to main causes a problem.


contribute/en/Style-and-Tone.md
```diff
@@ -67,7 +67,7 @@ Try not to sound like a robot. Write conversationally, as if you were talking to
 
 Headings use sentence case “This is a heading” unless delineated (eg “Look - This is a heading”).
 
-Nouns use lower case unless they have a formal definition in this style guide (eg “Directory”, “Fader”). The only exception to this is the word “person” or “people” which can remain in lower case.
+Nouns use lower case unless they have a formal definition in this style guide (eg “Directory”, “Fader”) or refer to specific Jamulus features (eg "Audio Alerts", "Jitter Buffer"). The only exception to this is the word “person” or “people” which can remain in lower case.
 
 Refer to UI labels in inverted commas (eg 'click on the “mute” button')
```
pljones: So, um, would that be `click on the "Mute" button` to stick with the new direction..? :)

contribute/en/Style-and-Tone.md
```diff
@@ -67,9 +67,9 @@ Try not to sound like a robot. Write conversationally, as if you were talking to
 
 Headings use sentence case “This is a heading” unless delineated (eg “Look - This is a heading”).
 
-Nouns use lower case unless they have a formal definition in this style guide (eg “Directory”, “Fader”). The only exception to this is the word “person” or “people” which can remain in lower case.
+Nouns use lower case unless they have a formal definition in this style guide (eg “Directory”, “Fader”) or refer to specific Jamulus features (eg "Audio Alerts", "Jitter Buffer"). The only exception to this is the word “person” or “people” which can remain in lower case.
```
pljones: OK, the remaining problem here is the sentence:
> The only exception to this is the word “person” or “people” which can remain in lower case.

Does that mean it _must_ or _should_?  Or can you just mix it up?  It's a bit non-directive...

gilgongo: Maybe we can just remove that sentence, since “person” or “people” don't have a formal definition for us really.

pljones: It's under "Standard terms":
> “Person” A human connected to a server (may be on multiple channels). We might say "People on the server", or "People who have muted themselves", rather than Musicians or Channels.

Note "rather than Musicians or Channels", with capitalisation, implying that "Person" and "People" would be.  Of course, we don't...  So it's probably worth adding an explicit mention _there_ about capitalisation, too, by reference (and consistently: we don't).

pljones: Separately, under Units:
> We use the following abbreviations:
> Kbit/s; Mbit/s; KByte/s; MByte/s.

I think the lack of clarity people have (for bits and bytes stuff) is whether "K", "M", etc are 1024-based or 1000-based.  It's probably not worth worrying about -- and probably worth mentioning that we don't worry about it.

contribute/en/Style-and-Tone.md
```diff
@@ -109,3 +109,5 @@ Kbit/s; Mbit/s; KByte/s; MByte/s.
 
 Values followed by units such as `ms`, `s`, `m`, `km`, `GHz` etc. should be separated by a space; e.g. "20 ms" and "1 GHz" not "20ms" and "1GHz".
 
+Note that we do not distinguish between 1024-based and 1000-based values for these.
```
pljones: Only applies to the bits and bytes.  Should be part of the same paragraph.  Does not apply to time.

pljones (approved):
Apart from the two extra blank lines, it looks fine now.

contribute/en/Style-and-Tone.md
gilgongo: BTW I take it we're still OK to merge this direct to release? I was just wary of the recent issues with Weblate.

contribute/en/Style-and-Tone.md
mcfnord:
```suggestion
Nouns use lower case unless they have a formal definition in this style guide (e.g. “Directory”, “Fader”) or refer to specific Jamulus features (e.g. “Audio Alerts”, “Jitter Buffer”). The only exception to this is the word “person” or “people” which can remain in lower case.
```

contribute/en/Style-and-Tone.md
```diff
@@ -67,9 +67,9 @@ Try not to sound like a robot. Write conversationally, as if you were talking to
 
 Headings use sentence case “This is a heading” unless delineated (eg “Look - This is a heading”).
 
-Nouns use lower case unless they have a formal definition in this style guide (eg “Directory”, “Fader”). The only exception to this is the word “person” or “people” which can remain in lower case.
+Nouns use lower case unless they have a formal definition in this style guide (eg “Directory”, “Fader”) or refer to specific Jamulus features (eg "Audio Alerts", "Jitter Buffer"). The only exception to this is the word “person” or “people” which can remain in lower case.
 
-Refer to UI labels in inverted commas (eg 'click on the “mute” button')
+Refer to UI labels in inverted commas (eg 'click on the “Mute” button')
```
mcfnord:
```suggestion
Refer to UI labels in inverted commas (e.g. 'click on the “Mute” button').
```

PR #1088, opened 2025-04-20
Remove "(Universal)" from macOS download

ann0see:
Since Jamulus has been available for Arm and Intel for a while now on macOS, we don't need to specifically mention "Universal" anymore in my opinion. This doesn't mean we should drop intel support any time soon.

No

Ready for review

Nothing
YES

## Checklist

- [ ] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [ ] I'm sure that this Pull Request goes to the correct branch

### DIFF
--- a/wiki/en/Installation-for-Macintosh.md
+++ b/wiki/en/Installation-for-Macintosh.md
@@ -15,7 +15,7 @@ Make sure you've already read the [Getting Started](Getting-Started) page.
 
 Upgrading? You may want to [back up your configuration](Software-Manual#backing-up-jamulus) first.
 
-1. [Download Jamulus (Universal build)]({{ site.download_root_link }}{{ site.download_file_names.mac }}){: .button}\\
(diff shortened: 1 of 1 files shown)


pljones:
Having "(Universal)" was a hint there were not two (or more) separate downloads, I take it?  Given we've got the legacy download (intel only), will it still be clear that the unlabelled one is intended for current macOS versions?

(As I have absolutely no knowledge of the mac ecosystem, I could be talking rubbish...)

ann0see:
The "universal" note was just a hint that the app supports ARM and Intel natively.
For a short time we had one ARM and one Intel build. 

I no longer see the "Universal" term in the wild that often. 
Therefore I suggest to drop it.

pljones (approved):

PR #1133, opened 2026-04-26
Update server bandwith table with correct values for existing and new audio quality settings

dingodoppelt:
Updated audio bandwidth table to include 'max' quality and adjusted buffer duration samples.  

This is a follow up to [#3653](https://github.com/jamulussoftware/jamulus/pull/3653)  

Yes

## Checklist

- [X] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)


wiki/en/Server-Bandwidth.md
```diff
@@ -11,22 +11,24 @@ permalink: "/wiki/Server-Bandwidth"
 
 The audio settings have an impact on the required network bandwidth. The table below summarises network requirements with respect to the configuration of:
 * Channels : stereo/mono
-* Quality : high/medium/low
-* Audio buffer duration : 2.67 ms, 5.33 ms, 10.67 ms, 21.33 ms
+* Quality : high/medium/low/max
```
mcfnord:
```suggestion
* Quality : max/high/medium/low
```

wiki/en/Server-Bandwidth.md
mcfnord, line 37: Just noting that max is not included in this network bandwidth section. That diagram is already pretty confusing. But is it reasonable to infer here that Max's larger requirements present new risk? Does running with Max make a server more likely to fail (degrade) where High succeeds today? I tend to experience very good results on the smallest Linux cloud instance available. Still true with Max?

dingodoppelt: > Does running with Max make a server more likely to fail (degrade) where High succeeds today? I tend to experience very good results on the smallest Linux cloud instance available. Still true with Max?

Since there is no OPUS coding happening theoretically the opposite should apply, meaning there should be less stress on the server's CPU. I don't know of any case where an admin had to decrease the amount of server slots for it to work with raw audio.  
The only observed issue in terms of performance is the upload bandwidth available to the client, which is fully covered by the table. For today's smallest VPS machines you get 500Mbit to 1Gbit connections while Jamulus currently allows for 150 clients on a server so you can't break it bandwidth wise. Even the highest possible number of server slots is limited to 256 so almost no chance of exhausting bandwidth with Jamulus.

mcfnord: Good points. The client upload speed is the weakest link today. I bet running clients with max streams on a server hosted on a non-fiber residential service would be the biggest chance for major degradation. [Currently we say](https://jamulus.io/wiki/Running-a-Server): _"It’s usually fine to have less than 5 players on a slower-speed home connection (eg 10 Mbit/s down and 1 Mbit/s up)."_ Is that still true?

dingodoppelt: > [Currently we say](https://jamulus.io/wiki/Running-a-Server): "It’s usually fine to have less than 5 players on a slower-speed home connection (eg 10 Mbit/s down and 1 Mbit/s up)." Is that still true?

I don't know if it has ever been a valid point and this page hasn't seen updates in many years judging by the assumed bandwidth. Anyone trying these days to host a server on his 1Mbit home connection should rather be using a public server or not host a server at all. A user is better off doing a speed test for his home/server connection and comparing the results to the bandwidth table. If he isn't able to do the math then it's not our fault. He still will find his settings working or not once he just tries it out.  
The docs don't say anything about a default or recommended network rate so nowadays where bandwidth shouldn't be an issue we might as well get rid of that sentence because it just applies to cases where bandwidth _is_ an issue and therefore can only apply to the lowest possible settings, which are untouched by the raw audio code.
Most importantly: Any user will still benefit sound wise from raw audio servers regardless of his quality settings, as long as most of the clients are using raw audio since they don't send any noise to the server and so OPUS clients get just a fracture of noise, namely download jitter and OPUS noise on a one-time-encoded stereo sum of otherwise raw signals.

ann0see:
Due to the release, you should re create the PR (or rebase it).

dingodoppelt:
> Due to the release, you should re create the PR (or rebase it).

Done that :)

contribute/en/Release-Process.md
```diff
@@ -327,7 +327,7 @@ Current state: <!-- Planning|Translations (beta)|Code freeze (rc)|Released -->
     - [ ] Check the update notice is displayed correctly
     - [ ] Revert the update notification Server to the previous release
   - [ ] [Force tag that tag as `latest` and push.](https://jamulus.io/contribute/Release-Process#if-this-is-a-proper-release-move-the-latest-tag)
-  - [ ] [Upload the artifacts to SourceForge and set defaults](https://jamulus.io/contribute/Release-Process#upload-binaries-to-sourceforge).
+  - [ ] [Check that binaries have been auto-uploaded to SourceForge](https://jamulus.io/contribute/Release-Process#upload-binaries-to-sourceforge).
```
pljones: Nice but unrelated change.

ann0see: I think I did that on the release branch.

pljones: Still, it shouldn't show as a change in this branch.

PR #1156, opened 2026-07-21
Document the `-c` port-reachability limitation

### LINKED ISSUE #1122: -c can't reach all servers listed on a built-in directory
As @rdica  [mentions here](https://github.com/jamulussoftware/jamuluswebsite/pull/1116#discussion_r2865450164), the -c command line option can't reach every server listed on a built-in directory. In fact, most servers might not be reachable using -c, because most clients rely on a "hole punch" procedure to reach a server without the server explicitly opening the port.

This probably merits mentioning in [the command-line documentation for this feature](https://jamulus.io/wiki/Software-Manual#command-line-options).

mcfnord:

Expands the `-c` / `--connect` entry in `Include-Client-Commands.md` with concise guidance on why it can fail to reach a server that the GUI reaches fine.

`-c` connects directly to the given address. Unlike selecting a server from the directory list in the GUI, it does not have a directory open the server's port, so the port must already be reachable from the client — which may require opening it in the router, cloud firewall/security group, and/or OS firewall. This holds whether or not the server is listed in a directory. Today the docs don't mention this and the failure is silent and confusing.

Relates to jamulussoftware/jamuluswebsite#1122. This documents the current behaviour and its limitation; it intentionally does **not** close #1122, which stays open for the separate discussion of improving `-c` itself (directory-assisted hole-punch), which would land in the Jamulus repo.

Working implementation.

Nothing outstanding — documentation only, single-line change to one English file. Ready for review.

YES (new English string) — hence this PR targets the `next-release` branch.

## Checklist

### DIFF
--- a/wiki/en/Include-Client-Commands.md
+++ b/wiki/en/Include-Client-Commands.md
@@ -1,6 +1,6 @@
 - `-M` or `--mutestream`  Prevent others on a server from hearing what I play
 - `--mutemyown` Prevent me from hearing what I play in the server mix (headless only)
--  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`. `-c` connects directly, so the server's port must already be reachable from your computer — you may need to open it in the router, cloud firewall/security group, and/or OS firewall. (Selecting a server from the directory list in the GUI opens the port for you; `-c` does not, whether or not the server is listed.)
+-  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`. `-c` connects directly, so the server's port must already be reachable from your computer — the server's operator may need to open the port in the router, cloud firewall/security group, and/or OS firewall. (Selecting a server from the directory list in the GUI opens the port for you; `-c` does not, whether or not the server is listed.)
 -  `-j` or `--nojackconnect`  Disable auto JACK connections
 -  `--ctrlmidich`  MIDI channel to listen on, Jamulus control + MIDI control number and count of consecutive CC numbers (or Jamulus channels), pick-up mode, device selection option. Format: `channel[;fn[*n]][;pn[*n]][;sn[*n]][;mn[*n]][;on][;u][;dDeviceName]` See [Tips & Tricks](Tips-Tricks-More#using---ctrlmidich-for-midi-controllers).
 - `--clientname`  Window title and JACK client name
--- a/wiki/en/Include-Client-Commands.md
+++ b/wiki/en/Include-Client-Commands.md
@@ -1,6 +1,6 @@
 - `-M` or `--mutestream`  Prevent others on a server from hearing what I play
 - `--mutemyown` Prevent me from hearing what I play in the server mix (headless only)
--  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`
+-  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`. `-c` connects directly, so the server's port must already be reachable from your computer — you may need to open it in the router, cloud firewall/security group, and/or OS firewall. (Selecting a server from the directory list in the GUI opens the port for you; `-c` does not, whether or not the server is listed.)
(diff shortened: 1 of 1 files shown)


wiki/en/Include-Client-Commands.md
```diff
@@ -1,6 +1,6 @@
 - `-M` or `--mutestream`  Prevent others on a server from hearing what I play
 - `--mutemyown` Prevent me from hearing what I play in the server mix (headless only)
--  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`
+-  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`. `-c` connects directly, so the server's port must already be reachable from your computer — you may need to open it in the router, cloud firewall/security group, and/or OS firewall. (Selecting a server from the directory list in the GUI opens the port for you; `-c` does not, whether or not the server is listed.)
```
mcfnord:
```suggestion
-  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`. `-c` connects directly, so the server's port must already be reachable from your computer — the server's operator may need to open the port in the router, cloud firewall/security group, and/or OS firewall. (Selecting a server from the directory list in the GUI opens the port for you; `-c` does not, whether or not the server is listed.)
```

wiki/en/Include-Client-Commands.md
```diff
@@ -1,6 +1,6 @@
 - `-M` or `--mutestream`  Prevent others on a server from hearing what I play
 - `--mutemyown` Prevent me from hearing what I play in the server mix (headless only)
--  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`
+-  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`. `-c` connects directly, so the server's port must already be reachable from your computer — the server's operator may need to open the port in the router, cloud firewall/security group, and/or OS firewall. (Selecting a server from the directory list in the GUI opens the port for you; `-c` does not, whether or not the server is listed.)
```
pljones: As ever, try to convince the LLM not to generate em-dash characters but use a plain ASCII hyphen.

mcfnord: Fair to flag, but there's no machine dependency to worry about here: this repo's kramdown is configured with `input: GFM` and no `smart_quotes` override, so there's no auto-typographic pipeline turning `--` into en/em dashes or straight quotes into curly ones (unlike classic kramdown/SmartyPants). A literal em-dash renders exactly as typed, for humans, permanently - and CLI-flag hyphens elsewhere in the docs are a different character entirely, so there's no collision risk either.

So this reads to me as taste/provenance-signaling rather than correctness, and I'd rather not have an agent enforce a blanket rule against style on that basis. Happy to take it out PR-by-PR if it bugs you on a given page, though.

Related, since we're on the topic: I did find a real (not just aesthetic) risk worth ruling out for agent-authored content here - smart/curly quotes. Unlike em-dash, a curly `'`/`"` pasted into a shell single-quoted string, JSON, a regex, or a code span isn't the character the parser expects, and silently corrupts it. Opened jamulussoftware/jamuluswebsite#1157 (draft) adding an AGENTS.md with that rule if you want to weigh in.

pljones: Yep - in plain text, the `&mdash;` vs `-` is just taste and style.  I think for coherence with the existing manual, type-written text we should continue with ASCII, though.  Mixing them looks very messy to me ("bad style").

mcfnord:
```suggestion
-  `-c` or `--connect`  Connect to given server address on startup, format `address[:port]`. `-c` connects directly, so the server's port must already be reachable from your computer - the server's operator may need to open the port in the router, cloud firewall/security group, and/or OS firewall. (Selecting a server from the directory list in the GUI opens the port for you; `-c` does not, whether or not the server is listed.)
```

pljones (approved):

PR #1145, opened 2026-05-30
#1127 Clear up Server Command Line options

pljones:
**Short description of changes**
As per https://github.com/jamulussoftware/jamuluswebsite/issues/1127, the Release Process referred to a non-existent document that should be checked.

This change:
* Creates (restores?) that document
* Updates the Release Process also to mention the separate Shared command line options file
* Removes the unused `Command-Line-Options.md` file
* Turns all the options lists into markdown tables
* Adjusts Running-a-Server references for slightly improved consistency
  * Separately, we could do with cleaning this page up...
* Adjusts Server-Troubleshooting references for slightly improved consistency
* Adjusts Software-Manual a heading and mentions the server options are elsewhere

**Context: Fixes an issue? Related issues**

Fixes 1127

**Status of this Pull Request**

I think it's working - not sure about the linkcheck errors (not in English):
```
$ docker run -e LANG=en_GB.UTF-8 --network host --rm tennox/linkcheck http://localhost:9800/ > linkcheck.log
$ cat linkcheck.log
Crawling...

http://localhost:9800/wiki/Command-Line-Options
- (445:0) 'Consejos..' => http://localhost:9800/wiki/Tips-Tricks-More#utilizar---ctrlmidich-para-controladores-midi (HTTP 200 but missing anchor)
- (452:9) 'Ejecutar..' => http://localhost:9800/wiki/Running-a-Server#opciones-relacionadas-con-el-modo-servidor (HTTP 200 but missing anchor)


Stats:
    3354 links
      72 destination URLs
     787 URLs ignored
       2 warnings
       0 errors
```

**What is missing until this pull request can be merged?**

Style review, someone to double-check it makes sense still.

**Does this need translation?**

Yes.  **NOTE** Links have changed.

## Checklist

- [x] I've verified that this Pull Request follows the [general code principles](https://github.com/jamulussoftware/jamulus/blob/main/CONTRIBUTING.md#jamulus-projectsource-code-general-principles)
- [x] I waited some time after this Pull Request was opened and all GitHub checks completed without errors. 
- [x] I'm sure that this Pull Request goes to the correct branch

PR #1158, opened 2026-07-21
CI: pin jekyll/builder image and install gems from the lockfile

mcfnord:
## Problem

`jekyll_site_ci` is currently failing on every PR (e.g. #1145), in the **Build the site in the jekyll/builder container** step:

```
Bundler::Definition#materialize': Could not find jekyll-4.3.3, jekyll-polyglot-1.8.1,
jekyll-minifier-0.1.10, jekyll-paginate-v2-3.0.0, ... in locally installed gems
```

while the container is running Jekyll 4.4.1.

## Root cause

The build ran inside the floating `jekyll/builder:latest` image and relied entirely on the gems baked into that image (there was no `bundle install`). Docker Hub recently republished `latest` as **Jekyll 4.4.1**, whose bundled gem set no longer matches the committed `Gemfile.lock` (Jekyll 4.3.3, jekyll-polyglot 1.8.1, ...). Bundler can't reconcile the lockfile against the image's preinstalled gems, so it fails in `Definition#materialize` and `jekyll build` exits 1.

This is environmental, not a defect in any page content — it broke all open PRs at once when the upstream tag moved. (There is no `jekyll/builder:4.3.3` tag on Docker Hub, so the lockfile could no longer be satisfied by `latest`.)

## Fix

- Pin the image to a fixed tag (`jekyll/builder:4.4.1`) so it can no longer silently drift.
- Run `bundle install && bundle exec jekyll build` so the committed `Gemfile.lock` is authoritative and the image only needs to supply Ruby + Bundler.

This unblocks #1145 and the other open PRs. Supersedes the empty WIP #1155.

## Note

An alternative would be to regenerate `Gemfile.lock` against 4.4.1 and keep the fast no-install path, but that requires confirming the plugins (jekyll-polyglot / jekyll-minifier / jekyll-paginate-v2) are 4.4-compatible. Installing from the lockfile is the lower-risk unblock.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/jekyll.yml
+++ b/.github/workflows/jekyll.yml
@@ -62,7 +62,7 @@ jobs:
         run: |
           docker run \
           -v ${{ github.workspace }}:/srv/jekyll -v ${{ github.workspace }}/_site:/srv/jekyll/_site \
-          jekyll/builder:latest /bin/bash -c "chmod a+w /srv/jekyll/Gemfile.lock && chmod 777 /srv/jekyll && jekyll build --future"
+          jekyll/builder:4.4.1 /bin/bash -c "chmod a+w /srv/jekyll/Gemfile.lock && chmod 777 /srv/jekyll && bundle install && bundle exec jekyll build --future"
       - name: Zip Website
         if: ${{ github.event_name == 'pull_request' }}
         run: zip -r ${{ github.workspace }}/website.zip _site/*


pljones:
Just to be clear -- this will _fail_ here because of the problem it's trying to fix.  We need to force the merge?

pljones:
Um.  Maybe this needs to go to `release`?  @ann0see @gilgongo @ignotus666 ?

ignotus666:
Yes. Then that update will be automatically merged into next-release.

mcfnord:
Retargeted to `release` and trimmed the branch to just the single `jekyll.yml` commit (it was cut from `next-release`, hence the earlier 56 unrelated translation files — those are gone now).

As noted above, `jekyll_site_ci` will still fail on this PR since the check runs the old workflow from the base; it needs a force-merge, after which the fix flows into next-release automatically.

PR #1159, opened 2026-07-23
Use em-dash/en-dash instead of ASCII hyphen substitutes

mcfnord:

Converts hyphen-as-dash usage to the typographically correct character across the English docs:

- **Em-dash (—)** for parenthetical asides, appositives, and heading breaks (e.g. `Software-Manual.md`: "Even if that person leaves and rejoins - or you join a server..." → "...rejoins — or you join a server..."), including existing `--` double-hyphen stand-ins in `Release-Process.md` and `Tips-Tricks-More.md`.
- **En-dash (–)** for numeric ranges (`0-16` → `0–16`, `30-50 ms` → `30–50 ms`, etc.) and for the "Title – Subtitle" pattern already established by `Command-Line-Options.md`, `Privacy-Statement.md`, and `2020-03-28-Server-Rpi.md` — `404.html`'s title used a plain hyphen for the identical pattern; this makes it consistent with the other three.

Compound words, CLI flags, list-bullet hyphens, ISO dates, and anything inside code spans are untouched — those hyphens are already correct and this PR does not touch them.

19 files changed, 40 lines (one dash-class swap each).

**Context**

Follows the em-dash discussion on #1156, where the open question was really "what does correct typography require," separate from whether an individual line's style is a matter of taste. This applies that consistently rather than adding an isolated case.

YES (existing English strings changed at the character level) — targets `next-release`.

## Checklist


mcfnord:
The standard should be use of reliable typographical features for the intended audience. I think this implies that natural language content in the main repo should be ASCII, and content in this repo rendered on browsers for our largest audience should use the full spectrum of reliable English typography.

pljones:
I can't type those characters.  That should be the determining factor for the character set used in the web site source code.

I can type \&ndash; and \&mdash;, where it's important enough.

If we referred to money, I'd say use \&pound; not £, too.

mcfnord:
> I can't type those characters. That should be the determining factor for the character set used in the web site source code.

How did i make this PR in 2 minutes without typing any of those? Even updating any 1967-genre text you might type is now blindingly fast and easy.

